PR #27, opened 2015-09-16
Fix table on jessie migration staff docs

jvperrin:
Attempts to fix [this page](https://dev.ocf.berkeley.edu/docs/staff/backend/jessie/) (image below) into a table by using markdown formatting instead of straight HTML.

ocf-jessie


chriskuehl:
thanks!

kpengboy:
thanks a lot!

PR #34, opened 2015-09-26
Add front page banner display from YAML

matthew-mcallister:
Pretty straightforward, there is already a file in ~staff.

Also, there might be something better to do with the upcoming hours sidebar.


ocfweb/component/banner.py
```diff
@@ -0,0 +1,23 @@
...
+    'visible',
+    'force_lab_closed',
+    'html',
+])):
+    pass
+
+
+def get_banner_message():
+    """Get the front page banner message from the default location."""
+    with urlopen('file:///home/s/st/staff/banner.yaml') as f:
```
chriskuehl: it'd be nice if those could work without NFS, or at least not fail loudly. Often people are running these on staff VMs which won't have this file. Usually we symlink the file into `~staff/public_html` and write code like:

```python
def get_vhost_db():
    """Returns lines from the vhost database. Loaded from the filesystem (if
    available), or from the web if not."""
    try:
        with open('/home/s/st/staff/vhost/vhost.conf') as f:
            return list(map(str.strip, f))
    except IOError:
        # fall back to database loaded from web
        return requests.get(constants.VHOST_DB_URL).text.split('\n')
```

Just make sure to chown the symlink (you'll have to use `-h` to not dereference it) to `staff:staff`.

ocfweb/component/banner.py
```diff
@@ -0,0 +1,23 @@
...
+        tree = yaml.load(f)
```
chriskuehl: `yaml.safe_load` please

pyyaml is idiotic. `yaml.load` is stupidly dangerous and will execute Python code embedded in the YAML (it should be called `dangerous_load` instead). You almost always want `safe_load` instead.

ocfweb/component/banner.py
```diff
@@ -0,0 +1,23 @@
...
+
+import yaml
+
+
+class BannerMessage(namedtuple('BannerMessage', [
+    'visible',
+    'force_lab_closed',
+    'html',
+])):
+    pass
```
chriskuehl: This is fine, but you can also just write

```python
BannerMessage = namedtuple('BannerMessage', ['visible', 'force_lab_closed', 'html'])
```

up to you

ocfweb/context_processors.py
```diff
@@ -3,13 +3,18 @@
...
     now = datetime.now()
     today = date.today()
     hours = DayHours.from_date(today)
 
+    banner = get_banner_message()
+
     return {
         'lab_is_open': hours.is_open(now),
         'current_lab_hours': hours,
+        'force_lab_closed': banner.visible and banner.force_lab_closed,
```
chriskuehl: I think we shouldn't tie `force_lab_closed` and the banner message together.

For example we might want to show a banner "Printing is down" without closing the entire lab. They can probably just be independent.

ocfweb/main/templates/home.html
```diff
@@ -26,6 +26,9 @@
 {% block content %}
 <div class="row">
     <div class="col-sm-8 ocf-content-block">
+        {% if banner.visible %}
+            {{banner.html | safe}}
+        {% endif %}
```
chriskuehl: looks great

ocfweb/templatetags/banner.py
```diff
@@ -0,0 +1,10 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def lab_hours_day_and_holiday(hours):
+    if hours.holiday:
+        return '{hours.weekday} ({hours.holiday})'.format(hours=hours)
+    return hours.weekday
```
chriskuehl: was this file added by accident?

chriskuehl:
overall looks great, thanks!

ocfweb/context_processors.py
matthew-mcallister: Well you can still do that, but you're right they could be independent. I
was thinking about returning everything to normal just by changing
`visible` to false. If they were independent, where would we specify to
force close the lab?

ocfweb/templatetags/banner.py
matthew-mcallister: Eh, why did I even create that file?

ocfweb/context_processors.py
chriskuehl: I think all we'd need is
`force_lab_closed: true/false`

You can just set it to false when the lab shouldn't be forced closed.

You're right, you can have the banner but not close the lab. But I don't think you can close the lab without a banner? (Not that we probably need to, but I think it makes sense to be able to)

matthew-mcallister: OK, but they're still logically tied by being in the same file, which is
fine. You also have to update both fields when restoring to normal, which
is also fine. Maybe this will evolve into a more general thing.

ocfweb/component/banner.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def get_banner_message():
+    """Get the front page banner message from the default location."""
+    with urlopen('file:///home/s/st/staff/banner.yaml') as f:
+        tree = yaml.load(f)
+    return BannerMessage(
+        visible=tree['visible'],
+        force_lab_closed=tree['force_lab_closed'],
+        html=tree['html'],
```
nattofriends: Consider:
```
BannerMessage = namedtuple('BannerMessage', ['html', 'force_lab_closed'])

[...]

return BannerMessage(
    force_lab_closed=tree['force_lab_closed'],
    html=tree['html'] if tree['visible'] else '',
)
```

then obviating your `banner.visible` check.

chriskuehl: I think eventually we might like to put some kind of wrapper (alert message styling) around the html so checking if it's visible might be useful.

```html
{% if banner.visible %}
   <div class="alert">{{banner.html | safe }}</div>
{% endif %}
```

As-is is fine, otherwise maybe use `None` instead of empty string?

chriskuehl: (I know empty string is falsey, but that seems yucky)

nattofriends: I chose `''` on purpose because I'm not 100% sure what `safe` does to `None`, pretty sure it will do nothing to `''`.

chriskuehl: either is fine with me

nattofriends: Not too opinionated either, if you wish to keep `.visible` then do it (however... it can also be approximated by checking truthiness of `.html`...!)

matthew-mcallister: Getting rid of `visible` sounds fine. When taking down a banner can just comment it out or use a dummy key name so we don't have to retype the HTML tags next time.

chriskuehl: I think @nattofriends is suggesting leaving it in the YAML but removing it from the namedtuple since it can be derived by checking the truthiness of html. (Assuming you change the code like he suggested.) I do think we should probably leave it in the YAML.

matthew-mcallister:
Oh, sure, I lost track of what the original suggestion was somewhere in the long line of comments.

ocfweb/component/banner.py
matthew-mcallister: Oh, sure, I lost track of what the original suggestion was somewhere in the long line of comments.

nattofriends: I was suggesting removing it everywhere but I'm ok with keeping it in the yaml

matthew-mcallister:
I see adding a commit gets in the way of the conversation in the pull request.

chriskuehl:
GitHub is fucking terrible for reviewing code unfortunately. Can't even see diffs between revisions if the author rebases :\

matthew-mcallister:
At least it keeps the old diffs and you can still comment on the new one.

chriskuehl:
@matthew-mcallister only if you don't rebase, otherwise you lose the old diff :p

ocfweb/main/staff_hours.py
```diff
@@ -9,6 +13,8 @@ def staff_hours(request):
         {
             'title': 'Staff Hours',
             'staff_hours': get_staff_hours(),
+            'today': time.strftime('%A'),
```
chriskuehl: nit: maybe call this `weekday` for consistency?

matthew-mcallister: Will it be obvious what `staff_hour.day == weekday` means then?

ocfweb/main/templates/staff-hours.html
```diff
@@ -9,11 +9,11 @@
         <p>OCF staff members hold regular drop-in staff hours to provide assistance with account issues or with OCF services. We're always happy to help troubleshoot account or service issues!</p>
         <p>Keep in mind the OCF volunteers sometimes have last-minute conflicts, so it's a good idea to check this page before coming in for cancellations.</p>
         {% for staff_hour in staff_hours %}
-            <div class="hour {% if staff_hour.cancelled %} cancelled {% endif %}">
+            <div class="hour {% if staff_hour.cancelled or staff_hour.day == today and lab_status.force_lab_closed %} cancelled {% endif %}">
```
chriskuehl: nit: parens around `or staff_hour.day == today and lab_status.force_lab_closed`? order of operations is hard

ocfweb/main/templates/staff-hours.html
```diff
@@ -9,11 +9,11 @@
...
                 <div class="title">
                     <strong>{{staff_hour.day}}</strong>
                     {{staff_hour.time}}
-                    {% if staff_hour.cancelled %}
+                    {% if staff_hour.cancelled or staff_hour.day == today and lab_status.force_lab_closed %}
```
chriskuehl: nit: parens around `or staff_hour.day == today and lab_status.force_lab_closed`? order of operations is hard

ocfweb/main/templates/staff-hours.html
matthew-mcallister: That's a syntax error apparently.

ocfweb/main/staff_hours.py
chriskuehl: oh, this is for staff_hours, not lab_hours.

(lab_hours I recently renamed it to `weekday` which is why I was thinking it'd be consistent)

Seems fine as-is in that case.

ocfweb/main/templates/staff-hours.html
chriskuehl: ugh django, why? :\

matthew-mcallister:
Should we squash the commits?

chriskuehl:
your call, either is fine I think. The commits stand on their own.

ocfweb/component/lab_status.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def get_lab_status():
+    """Get the front page banner message from the default location."""
+    try:
+        with open('/home/s/st/staff/lab_status.yaml') as f:
+            tree = yaml.safe_load(f)
+    except IOError:
+        return requests.get(
+            'https://www.ocf.berkeley.edu/~staff/lab_status.yaml').text
```
chriskuehl: oh, this is probably the issue. you're just returning the text here, you probably want to replace this with `yaml.safe_load`ing the text.

sorry, didn't read the code properly the first time

you can probably replicate the test failure if you temporarily change the path on line 17 to something that doesn't exist

matthew-mcallister: You know, the website might actually go down if either `staff_hours` or `lab_status` are unavailable.

matthew-mcallister: Hypothesis tested, it will.

nattofriends: This sounds inadvisable

Issue #30, opened 2015-09-19
Figure out how to deploy in prod


chriskuehl:
My thoughts for deploying right now:

* Package as a Debian package (dh-virtualenv), similar to atool or create; install on death
* Run under gunicorn, supervise with systemd (just like atool)
* Proxy to gunicorn from Apache if the request doesn't match a userdir pattern.

Yet unanswered questions:

* **How do we serve static assets?** Currently we deploy them to https://www.ocf.berkeley.edu/~ocfweb/static/ but it's not a great solution.

    Right now I'm thinking package them and deploy them to some directory Apache serves at `static.ocf.berkeley.edu` or similar.

* **How do we gracefully fall back to the old content?** We can set up Apache to proxy if it doesn't match a userdir pattern, but this means old pages like `/OCF/whatever` will be going to ocfweb, and it needs some way to handle them if it doesn't know about the page.

    There aren't that many pages on our current site, maybe we can just hard-code them in the Apache config as we remove them one-by-one (so we never actually proxy to ocfweb for them). Here's a list of current pages: https://gist.github.com/chriskuehl/8be24760c01e76574dfd (some are not actual pages)

chriskuehl:
Unfortunately actually deploying on death is blocked by the jessie upgrade. We might consider hosting it on a separate server for the time being.

kpengboy:
If you're going to set up another server, it's your chance to use coma, for
Sleep is the brother of Death

chriskuehl:
This work is done.

Most of the packaging work took place in https://github.com/ocf/ocfweb/pull/37
Most of the deployment work took place in https://github.com/ocf/puppet/commit/90a791e42af9d17d9d3333ef225876130d49b0c4

https://dev.ocf.berkeley.edu/ is now served exactly as we intend it to be served when we launch (fallback to old website, userdirs, etc. should all be working).

chriskuehl:
This will get a lot simpler once we upgrade death to jessie (e.g. we can drop nginx entirely, and just serve static assets via Apache on death).

PR #37, opened 2015-09-29
Add initial Debian packaging

chriskuehl:
This gets us a lot closer to #30.

.The idea is to build a Debian package and serve it with gunicorn supervised by systemd (almost the same way we do atool), and stick the static files at `/usr/share/ocfweb/static` (which Apache will be configured to serve -- but not part of the packaging, would rather do it with Puppet).

Unfortunately this won't build on wheezy (well, maybe it will, but we don't have a wheezy Jenkins slave, so we can't yet install this on death -- will probably bring up a new server instead).

@daradib


chriskuehl:
merging this since it's blocking me and shouldn't break anything. still welcome to suggestions, though.

PR #50, opened 2015-10-02
Fix staff hours head spacing

matthew-mcallister:
We were double padding since Bootstrap pads list items by default.

https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_type.scss#L179

```css
// Inline turns list items into inline-block
.list-inline {
  @include list-unstyled;
  margin-left: -5px;

  > li {
    display: inline-block;
    padding-left: 5px;
    padding-right: 5px;
  }
}
```
screenshot - 10022015 - 02 03 09 pm

### DIFF
--- a/ocfweb/main/templates/home.html
+++ b/ocfweb/main/templates/home.html
@@ -95,11 +95,13 @@
                             <span class="cancelled-text">cancelled this week</span>
                         {% endif %}
                     </p>
-                    <ul class="ocf-staffhours-faces">
+                    <div class="ocf-staffhours-faces">
+                      <!--
                         {% for staffer in staff_hour.staff %}
-                            <li><img alt="{{staffer.user_name}}" src="{{staffer|gravatar:77}}" /></li>
+                            --><img alt="{{staffer.user_name}}" src="{{staffer|gravatar:77}}" /><!--
                         {% endfor %}
-                    </ul>
+                      -->
+                    </div>
                 </div>
             {% endfor %}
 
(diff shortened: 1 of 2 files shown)


chriskuehl:
LGTM, but how does it look if there are multiple rows?

e.g.

matthew-mcallister:
Good question. It looks as you'd expect.

We could add top/bottom padding; should we squeeze all dimensions on mobile?

screenshot - 10022015 - 02 07 59 pm

chriskuehl:
maybe just add top/bottom padding for now since that's easy, and we can rethink the mobile aspect in #3?

chriskuehl:
If you can come up with a good solution for mobile now feel free, but imo not a blocker.

matthew-mcallister:
This creates uniform spacing on the horizontal and vertical, and removes the spacing on small displays. I probably hate CSS now though.

screenshot - 10022015 - 04 50 39 pm

ocfweb/main/templates/home.html
```diff
@@ -95,11 +95,13 @@
...
                     </p>
-                    <ul class="ocf-staffhours-faces">
+                    <div class="ocf-staffhours-faces">
+                      <!--
                         {% for staffer in staff_hour.staff %}
-                            <li><img alt="{{staffer.user_name}}" src="{{staffer|gravatar:77}}" /></li>
+                            --><img alt="{{staffer.user_name}}" src="{{staffer|gravatar:77}}" /><!--
                         {% endfor %}
-                    </ul>
+                      -->
```
matthew-mcallister: Apparently, this is considered a legitimate way to remove the space between images while preserving.

chriskuehl:
looks a little hacky but css sucks
we can always revise later

matthew-mcallister:
If it turns out to be less-than-ideal, we can just leave the spaces and lower the horizontal width or something.

PR #51, opened 2015-10-04
Migrate web badges page

matthew-mcallister:
Copy and paste job.


chriskuehl:
Can you add a redirect from the old path to the new one?

chriskuehl:
ship it

PR #54, opened 2015-10-05
markdown: add simple code highlighting

chriskuehl:
The styles/colors are not too great, but we can make them better later.

Before:

After:

### DIFF
--- a/ocfweb/component/markdown.py
+++ b/ocfweb/component/markdown.py
@@ -3,6 +3,11 @@ import re
 import mistune
 from django.core.urlresolvers import reverse
 from django.utils.html import strip_tags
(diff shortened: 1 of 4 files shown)


matthew-mcallister:
Looks great, but seems inconsistent at identifying language. E.g.

screenshot - 10042015 - 06 40 24 pm

chriskuehl:
Looks like the first block is incorrect identified as "Velocity" while the second is correctly "Bash".

-shrug- I don't think we'll be able to improve on pygments' auto-detection without a lot of work. In the future we can just use the triple backtick syntax with the language, e.g.

    ```python
    # python code here
    ```

PR #57, opened 2015-10-06
Add links to home page

### LINKED ISSUE #7: Home page should have a link to password reset, check print quota, view print queue

chriskuehl:
This fixes #7 

We should move the Google Maps map to the wiki page (which should no longer be a wiki page anyway), but imo not blocking the site launch. I actually like this better since the map was slow to load and jumped on page refresh.

It doesn't look too great so suggestions welcome, but I think we can do that in a followup so that we can start launching the site.

### DIFF
--- a/ocfweb/context_processors.py
+++ b/ocfweb/context_processors.py
@@ -1,10 +1,18 @@
 import re
 from datetime import date
 from datetime import datetime
+from ipaddress import ip_address
+from ipaddress import ip_network
 
+from ipware.ip import get_real_ip
 from ocflib.lab.hours import DayHours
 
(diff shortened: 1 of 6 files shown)


chriskuehl:
Per IRC discussion (@hexecute and @Ranigad), we should add back a static version of the map which links to an interactive once. Should be easy but let's do it as a followup. I'll file an issue.

chriskuehl:
The main reason for removing the map was that it's too large, but we can easily add back a shorter version to the sidebar.

matthew-mcallister:
This change looks good. If you get the static map working, consider amending this pull request to kill two birds with one stone.

PR #59, opened 2015-10-06
Add Google Maps back to home page

chriskuehl:
The static map didn't work very well (hard to make it work since we support lots of widths now).

This is a smaller version of what was there before (and it now starts with the info window not open).


matthew-mcallister:
Tested and it looks good (we should add the full map to the lab hours page).

The sidebar is getting pretty long though

PR #60, opened 2015-10-07
Fix hosting urls, reduce duplication

chriskuehl:
This reduces some duplication and also fixes the URLs in the hosting images to point to the new locations (not the legacy redirects).


matthew-mcallister:
Great work. Only complaint is that "Badges" in the title is capitalized, which conflicts with e.g. "Virtual **h**osting".

chriskuehl:
@matthew-mcallister fixed, thanks!

PR #63, opened 2015-10-07
Use open style on dropdown nav item on hover

chriskuehl:
Follow-up to @matthew-mcallister's #62 

This gives us the same "open" styling as when clicking if you hover:


matthew-mcallister:
It's kind of weird with clicking, but I suppose it does look a bit better. I vote merge at your discretion.

PR #66, opened 2015-10-09
More responsive home page

chriskuehl:
This makes the home page (and some other areas of the site) look a lot better on mobile.

There are a bunch of minor fixes (most of which you can't see on the screenshots below).

before:

after:

There's still a big variance in mobile screen sizes (especially iPhone < 6). You can use Chrome's dev tools to simulate them easily.

Once this is merged we can launch!

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -24,3 +24,7 @@ linters:
 
     ColorKeyword:
         enabled: false
+
+    # don't want to force using `//` comments everywhere
+    Comment:
+        enabled: false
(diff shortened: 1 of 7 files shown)


matthew-mcallister:
These changes look pretty good on my 5.2'', maybe others can test some different screen sizes. Some issues still lingering:

Line break in lab hours bar now
hours line

Navbar still broken
navbar

chriskuehl:
The line break is intentional (basically, it keeps "9am-6pm on Thursday more >>" on one line; i.e. the line break is only there if it won't otherwise fit on one line), but the "Hours:" should be part of that line. Will fix.

Forgot about navbar on sm, will work on that.

Also the hours in the sidebar are broken on sm.

PR #67, opened 2015-10-09
Minor responsive fixes to size sm

chriskuehl:
* Reduce navbar spacing on sm so that the navbar items fit:

* Adjust hours columns on sidebar

* Wrap "Hours:" along with the actual hours

* Wrap "Sound like you?" with "Join the staff team!"


matthew-mcallister:
Nice, this looks great.

Issue #32, opened 2015-09-19
Add easy way to post urgent messages


chriskuehl:
On the current website we sometimes post urgent messages like this:

![](http://i.fluffy.cc/JLTRrPQ1XCW2Jfrj5XQggXfKLm3wgxbf.png)

It would be nice if we could do something similar on ocfweb, preferably reading from an external file so we don't have to push the site to update it.

It should also allow overriding the "is lab open" check (which currently is based on [ocflib hours](https://github.com/ocf/ocflib/blob/master/ocflib/lab/hours.py)).

matthew-mcallister:
So are we thinking something like the tried-and-true yaml file in ~staff?

chriskuehl:
I think so. A YAML file of this format maybe?

```yaml
# Whether to override the default "lab is open" logic. This can be
# used to force the lab to be closed in an emergency.
#
# For regular scheduled holidays, add them to ocflib instead.
# (See `pydoc3 ocflib.lab.hours` for more details.)
#
# If override_lab_hours is true, then `lab_open` controls whether
# the lab is considered open or not.
override_lab_hours: false
lab_open: false

# Whether to display an urgent status message on the home page.
#
# The urgent message can contain HTML. The recommended format is
# an <h2> title and <p> paragraphs with more info.
display_urgent_message: false
urgent_message: |
    <h2>The lab is currently closed due to xyz</h2>
    <p>For more info, visit <a href="http://status.ocf.berkeley.edu">our status blog</a>.</p>
```

matthew-mcallister:
You could merge `override_lab_hours` and `lab_open` into one field `lab_open` with values `true`, `false`, and `default`.

chriskuehl:
I think YAML will interpret that as sometimes a boolean and others a string. If we do that we should make sure to wrap in quotes (which normally isn't required for most YAML strings) since variables of unpredictable type suck.

Or we could just use JSON because YAML is horrible.

matthew-mcallister:
JSON sounds like a good idea.

nattofriends:
JSON cannot contain comments. If you want a ternary type, consider the
objective-c mainstays "yes" and "no".

chriskuehl:
the normal way to do json comments is...

```json
{
    "//": "this is a ghetto comment",
    "something": "foo"
}
```
we already use that in some places (e.g. Chrome's JSON preferences)

[...]

PR #70, opened 2015-10-12
Add hours to lab page, notice on home about new hours

chriskuehl:


kpengboy:
...well, I was about to do that...

chriskuehl:
@tmochida says shipit so shipping

comments still welcome, just want to get this out ASAP since the new hours start tomorrow

ocfweb/main/templates/home.html
```diff
@@ -26,7 +26,14 @@
...
 <div class="row">
     <div class="col-sm-8 ocf-content-block">
+
         {{lab_status.banner_html | safe}}
+
+        <a href="{% url 'doc' 'services/lab' %}" class="update-banner">
+            <h4>Lab hours have been extended!</h4>
+            <p>Click here to see the updated hours.</p>
+        </a>
+
```
matthew-mcallister: Why the banner code? What about https://github.com/ocf/ocfweb/issues/32?

chriskuehl: I was thinking to save that for urgent messages ("printing is on fire") rather than periodic updates but open to suggestions.

For example, we'll want to post a message long-term when a move date is set, this seemed a good use for that.

matthew-mcallister: I thought the point was to decouple actual development from temporary updates.

We could put multiple banners into lab_status.yaml if we need to have an urgent message and a long term message, for example.

PR #73, opened 2015-10-13
Offset anchors on docs

chriskuehl:
Previously anchors would take you to the header exactly, but the 60px fixed OCF navbar would cover the heading.

before:

after:

This hacks an empty block before it to offset it. It doesn't affect how the page looks at all, but makes the anchors behave properly.


matthew-mcallister:
This is sweet. That was annoying me but I figured probably no non-staff users would ever notice.

PR #75, opened 2015-11-01
Temporarily switch to pre-commit-by-types fork

chriskuehl:
This switches us to a pre-commit branch I'm developing that allows hooks to target files by "type" rather than merely extension. This means that it can, for example, target "all text files" or "all Python files, even if they only have a Python shebang but no extension".

The branch seems to be pretty stable but I'd like to get some more real-world testing, hence switching some OCF projects to use it.

The main changes you see here are switching the hooks repos to my own forks that declare hooks by type instead of extension, plus removing some things we no longer have to exclude (my branch fixes issues with submodules and symlinks being treated as regular files by some bad hooks).

OCF has Debian packages for the fork uploaded, and you can pip install it as `ckuehl-pre-commit-types` (you might need to target one of the dev targets specifically).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
--   repo: https://github.com/pre-commit/pre-commit-hooks.git
-    sha: 616c1ebd1898c91de9a0548866a59cbd9f4547f6
+-   repo: https://github.com/chriskuehl/pre-commit-hooks.git
+    sha: cd3f867cafee3488a93a13f97bf7fdab329c996e
     hooks:
     -   id: autopep8-wrapper
         language_version: python3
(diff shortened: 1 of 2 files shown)


matthew-mcallister:
Cloned, tested all the make targets, pre-commit runs successfully, its targets work on extension-less files with shebang. Seems like its ready to ship.

To anybody updating their repo, you should run `make clean` first or just delete and clone again.

nattofriends:
I don't see the point of committing the switch. You could switch, apply the pre-commit diff, commit that, and wait till pre-commit master updates. Then people who installed pre-commit in the intervening period wouldn't have to reinstall pre-commit.

chriskuehl:
@nattofriends the point of committing the switch is that I'd like to trial out the change for a few weeks before trying to merge it into pre-commit master to ensure it's solid. I'm using OCF as a testing ground, which is kind of useful seeing as this branch benefits us a lot (more in utils than here).

PR #84, opened 2015-11-22
Add daily graph to new stats section

chriskuehl:
This moves the daily graph generation from the labstats crummy CGI scripts, which are difficult to test (you either need to stop puppet and work as root on dementors, or go grab the secret MySQL password) and somewhat messy (that repo is responsible for cron jobs, cgi scripts, a private CA, some library-like functions, and a few shell scripts).

The graph looks just like what you're used to:

The code was originally written mostly by me and then heavily improved by @william-zt-ho. This is almost a straight copy.

The URL is like https://www.ocf.berkeley.edu/stats/daily-graph/graph (this won't work until this PR is merged, obviously).

You can specify `?date=2014-09-28` to see historical data, which you couldn't easily on the old labstats. It would be cool to eventually add a better UI that makes it easy to browse historical stats (plus more stuff like printing).


requirements.txt
```diff
@@ -1,15 +1,18 @@
...
 gunicorn==19.3.0
-ldap3==0.9.9.2
-libsass==0.9.1
-lxml==3.4.4
+ldap3==0.9.9.3
+libsass==0.9.2
+lxml==3.5.0
+matplotlib==1.5.0
 mistune==0.7.1
+numpy==1.10.1
```
chriskuehl: rip speedy build times

chriskuehl:
> that repo is responsible for cron jobs, cgi scripts, a private CA, some library-like functions, and a few shell scripts

and some PHP

PR #87, opened 2015-11-24
Fix stats plots in prod by providing matplotlibrc

chriskuehl:


chriskuehl:
The Debian package seems to work locally, going to push to prod.

PR #88, opened 2015-11-24
Add start of new stats page

chriskuehl:
This moves some more stats to ocfweb. It doesn't look very great but neither did the old page, and we can improve it later.

It's also pretty slow...


matthew-mcallister:
There are trailing ".0"s in the "Idle" column.

matthew-mcallister:
OK now it looks dandy. Push if it looks ready to you.

PR #91, opened 2015-12-05
Directly merge atool into ocfweb

matthew-mcallister:
All pages seem to work as well as authentication and the Debian package. Account creation will probably work on prod.


matthew-mcallister:
This is running on meltdown but you have to spoof the host HTTP header field and forward to port 8000 on meltdown via SSH.

chriskuehl:
Merged in #92

PR #95, opened 2015-12-06
Use shared Redis cache instead of per-process in-memory cache

chriskuehl:


chriskuehl:
works well, though we still don't truly share the cache due to functions not mapping to the same memory locations in each process. will fix that in a followup.

chriskuehl:
(i deployed the branch deb built by Jenkins into prod)

PR #96, opened 2015-12-06
Use a newer pip with dh-virtualenv to speed up builds

chriskuehl:


chriskuehl:
retest this please

chriskuehl:
This takes the build-deb time from ~11 minutes to 2.5 minutes :D

chriskuehl:
That's with a primed cache, but not much we can do to speed up the compilation of packages. This should make the Jenkins jobs fairly fast, though.

PR #98, opened 2015-12-11
stats: add basic printing numbers

chriskuehl:
Unfortunately text on Bootstrap progress bars is unreadable when the bars get too short, so I just left the labels blank. I think it still looks kinda reasonable though...

On small screens (phones) they're in the same column which still looks fine:


matthew-mcallister:
I don't know what you're complaining about; this looks great.

jvperrin:
Yeah, this looks nice to me, ship it! :shipit:

On a slightly different note, I'm working on moving over the printing and toner graphs we used to have as part of my highcharts branch so that the old stats page can be retired.

Issue #99, opened 2015-12-13
get_blog_posts sometimes returns None which isn't handled properly


chriskuehl:
didn't investigate at all yet

```
A problem was encountered and reported via ocflib:

An exception occured in ocfweb:

Traceback (most recent call last):
  File "/usr/share/python/ocfweb/lib/python3.4/site-packages/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/share/python/ocfweb/lib/python3.4/site-packages/ocfweb/main/home.py", line 24, in home
    blog_posts = get_blog_posts()[:2]
TypeError: 'NoneType' object is not subscriptable



Request:
  * Host: www.ocf.berkeley.edu
  * Path: /
  * Method: GET
  * Secure: True

Request Headers:
{'HTTP_ACCEPT_ENCODING': 'gzip, deflate',
 'HTTP_CONNECTION': 'close',
 'HTTP_HOST': 'www.ocf.berkeley.edu',
 'HTTP_USER_AGENT': 'Rackspace Monitoring/1.1 '
                    '(https://monitoring.api.rackspacecloud.com)',
 'HTTP_X_FORWARDED_FOR': '78.136.44.23, 169.229.226.23',
 'HTTP_X_FORWARDED_HOST': 'www.ocf.berkeley.edu',
 'HTTP_X_FORWARDED_PROTO': 'https',
 'HTTP_X_FORWARDED_SERVER': 'www.ocf.berkeley.edu',
 'PATH_INFO': '/',
 'QUERY_STRING': '',
 'RAW_URI': '/',
 'REMOTE_ADDR': '127.0.0.1',
 'REMOTE_PORT': '46784',
 'REQUEST_METHOD': 'GET',
 'SCRIPT_NAME': '',
 'SERVER_NAME': '127.0.0.1',
 'SERVER_PORT': '8000',
 'SERVER_PROTOCOL': 'HTTP/1.0',
 'SERVER_SOFTWARE': 'gunicorn/19.4.1',
 'gunicorn.socket': <socket.socket fd=16, family=AddressFamily.AF_INET, type=SocketType.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 8000), raddr=('127.0.0.1', 46784)>,
 'wsgi.errors': <gunicorn.http.wsgi.WSGIErrorsWrapper object at 0x7efe33be1908>,
[...]

PR #100, opened 2015-12-14
Fix rare race condition in caching code

chriskuehl:
If the key expires between when we check if it's in the cache and when we call `get`, we might return `None` in error.

### DIFF
--- a/ocfweb/caching.py
+++ b/ocfweb/caching.py
@@ -42,8 +42,16 @@ def cache(ttl=None):
                 tuple((k, v) for k, v in sorted(kwargs.keys())),
             )
 
-            if key in django_cache:
-                return django_cache.get(key)
+            # The "get" method returns `None` both for cached values of `None`,
+            # and keys which aren't in the cache.
+            #
+            # The recommended workaround is using a sentinel as a default
+            # return value for when a key is missing. This allows us to still
+            # cache functions which return None.
+            cache_miss_sentinel = {}
+            retval = django_cache.get(key, cache_miss_sentinel)
+            if retval is not cache_miss_sentinel:
+                return retval
 
             result = fn(*args, **kwargs)
             django_cache.set(key, result, ttl)


chriskuehl:
note: I have not tested this at all, need to try it in prod

nattofriends:
Your explanation in the PR and commit is inconsistent with the comment. Which one is it?

chriskuehl:
@nattofriends how is it inconsistent?

chriskuehl:
@nattofriends the problem with the code in master is "If the key expires between when we check if it's in the cache and when we call get, we might return None in error."

If we fix that by just calling `get` and checking if it's `None`, we won't be able to distinguish between a `None` in the cache, and a cache miss. Hence we use a sentinel (the bit explained in the code).

matthew-mcallister:
As @chriskuehl explained this fixes https://github.com/ocf/ocfweb/issues/99, where the issue was that a timeout lead to a cache miss lead to returning `None`

nattofriends:
OK, that makes sense. Why do you believe this will fix the issue?

chriskuehl:
> Why do you believe this will fix the issue?

The problem scenario with the old code is:

1. `if key in django_cache` is `True`, so we enter the if-block
2. key expires in the cache
3. `return django_cache.get(key)` now returns `None`

This can't happen anymore; `retval = django_cache.get(key, sentinel)` either returns the value or the sentinel. If the sentinel we know it's a cache miss, if the value we know it's a hit (and have the value which we can return without consulting the cache again).

jvperrin:
This looks good to me, it definitely explains how `None` could be returned from `get_blog_posts` and how rare of an issue it is. I thought your implementation would have an issue if we were caching an empty dictionary, but then I realized that you are actually comparing the objects, not the values, so that's fine. Ship it! :shipit:

chriskuehl:
seems to work

PR #101, opened 2015-12-16
Remake servers page: more maintainable, load some data from LDAP

chriskuehl:
@kpengboy @jvperrin @matthew-mcallister 

Eventually we want to load all of this data from LDAP, but this is a first step.

* Remove duplicated HTML, simplify CSS and markup
* Update servers list with more details
* Load desktops from LDAP
* Add links on the right (just munin currently)

Improvements still to be made:
* Collapsing of groups like in the old version.
* Grouping of hosts (besides via hypervisor). For example, grouping desktops together.
* Load all data from LDAP
* Allow more links that just munin
* Improve colors and look/feel


kpengboy:
I've been wanting to store the data in LDAP for a while, but since as of
now we can't remove attributes at will, I didn't want to rush to add extra
attributes we could regret later

chriskuehl:
We should try to avoid non-authoritative attributes as much as possible. In particular I don't like the idea of having a "hypervisor" or "parent" attribute, as it's going to become out-of-date really quickly. Instead maybe we can find a way to have that data shared with ocfweb (but I don't have any great ideas currently to figure out which hypervisor a VM runs on).

For now I think adding a description attribute would be a good start. There's not really any way to derive that from the configuration (besides parsing puppet classes, which isn't sufficient).

jvperrin:
We could periodically run `virsh list` on each of the physical servers in order to check which virtual machines run on each of them and then cache the returned result, but that might be too much work or moving parts for something like this. It wouldn't become outdated though, so that would be a plus. I like the added details for servers and links to munin though, those are really nice.

chriskuehl:
I added a bunch of descriptions to LDAP:
https://gist.github.com/anonymous/b875ec7ccf15e324787f

I'll try to clean up this branch a little more now and then we can merge it.

PR #105, opened 2015-12-19
Add start of periodic functions

chriskuehl:
Right now on ocfweb we have a lot of really slow calls:

* loading blog posts
* loading staff hours (mostly LDAP)
* loading lab_status (HTTP to death)
* loading stats

Some of these are behind a caching decorator, where the result is saved for some seconds (but one poor user still has to wait a long time to view the page, and because of low traffic on some of the pages, lots of users end up being poor users).

This introduces periodic functions. The docstring explains pretty well:

```python
def periodic(period, ttl=None):
    """Caching function decorator for functions which desire TTL-based caching.

    Using this decorator on a function registers it as a "periodic function",
    with the given period in seconds. The function will be run in the
    background at the requested frequency.

    When the function is called normally, a recent cached verison will thus
    almost certainly be available for immediate use, making the website fast
    for all users.

    The optional ttl (in seconds) specifies how long cache entries should live.
    By default, it is twice the period. If your function is called without
    being cached, it is synchronously executed (and the result stored for next
    time), much like the @cache decorator.

    Periodic functions can have no required arguments. While they can have
    keyword arguments, no caching is done if you call the function using them.

    Uses the Django cache (which uses Redis) to achieve a shared cache across
    worker processes.

    In DEBUG mode, no caching is done.

    Usage:

        @periodic(60)
        def get_blog_posts():
```

This will role out `periodic` without adding a cronjob yet to trigger the functions. This *does* add a script to trigger them, though, and we can use it to test in prod before adding the cronjob.

Without a cronjob, the periodic decorator acts just like the cache decorator, so the site will still be equally fast (or equally slow, depending on your mood).

I want to push this now, observe it in prod, and then push a cronjob if everything goes smoothly.

This also adds a test page where you can see some statistics:

This is purely informational and kind of useful. We can leave it on in prod since it's read-only and nothing sensitive is in it.

The decorator is very easy to use; you can throw it on any function you'd like (as long as it has no arguments), and it will automatically be added to the list of background jobs. You can call that function just like you normally would, and it will usually be instant.

This should make it easy for us to keep ocfweb fast without the same kind of cronjob hacks we had in e.g. labstats.

### DIFF
--- a/ocfweb/bin/run_periodic_functions.py
+++ b/ocfweb/bin/run_periodic_functions.py
@@ -1,0 +1,36 @@
+import logging
+import os
+from argparse import ArgumentParser
+
+from ocfweb.caching import periodic_functions
+
+
+_logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def run_periodic_functions(grace_period):
+    # First, import urls so that views are imported, decorators are run, and
+    # our periodic functions get registered.
+    import ocfweb.urls  # noqa
+
+    for pf in periodic_functions:
+        if pf.seconds_since_last_update() >= pf.period - grace_period:
+            pf.update()
+            _logger.info('Updating periodic function: {}'.format(pf))
+        else:
+            _logger.info('Not updating periodic function: {}'.format(pf))
+
+
+def main(argv=None):
+    os.environ['DJANGO_SETTINGS_MODULE'] = 'ocfweb.settings'
+
+    parser = ArgumentParser(description='Run ocfweb periodic functions')
+    parser.add_argument('-g', '--grace-period', type=int, default=5)
+    args = parser.parse_args(argv)
+
+    run_periodic_functions(args.grace_period)
+
+
+if __name__ == '__main__':
+    exit(main())
--- a/ocfweb/caching.py
+++ b/ocfweb/caching.py
@@ -1,10 +1,37 @@
 """Caching decorators for ocfweb."""
+import logging
+from collections import namedtuple
+from datetime import datetime
 from itertools import chain
 
+from cached_property import cached_property
 from django.conf import settings
 from django.core.cache import cache as django_cache
 
 from ocfweb.environment import ocfweb_version
+
+
+_logger = logging.getLogger(__name__)
+
+
+def cache_lookup(key):
+    """Look up a key in the cache, raising KeyError if it's a miss."""
+    # The "get" method returns `None` both for cached values of `None`,
+    # and keys which aren't in the cache.
+    #
+    # The recommended workaround is using a sentinel as a default
+    # return value for when a key is missing. This allows us to still
+    # cache functions which return None.
+    cache_miss_sentinel = {}
+    retval = django_cache.get(key, cache_miss_sentinel)
+    is_hit = retval is not cache_miss_sentinel
+
+    if not is_hit:
+        _logger.debug('Cache miss: {}'.format(key))
+        raise KeyError('Key "{}" is not in the cache.'.format(key))
+    else:
+        _logger.debug('Cache hit: {}'.format(key))
+        return retval
 
 
 def cache_lookup_with_fallback(key, fallback, ttl=None, force_miss=False):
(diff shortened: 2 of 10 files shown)


matthew-mcallister:
Killer. This would be the natural conclusion to the assorted caching
efforts and issues so far.

Should we ensure that calls to these periodic functions are staggered? Or
does it even matter since this isn't likely to slow down page serve times?

Right now on ocfweb we have a lot of really slow calls:

   - loading blog posts
   - loading staff hours (mostly LDAP)
   - loading lab_status (HTTP to death)
   - loading stats

Some of these are behind a caching decorator, where the result is saved for
some seconds (but one poor user still has to wait a long time to view the
page, and because of low traffic on some of the pages, lots of users end up
being poor users).

This introduces periodic functions. The docstring explains pretty well:

def periodic(period, ttl=None):
    """Caching function decorator for functions which desire TTL-based
caching.    Using this decorator on a function registers it as a
"periodic function",    with the given period in seconds. The function
will be run in the    background at the requested frequency.    When
the function is called normally, a recent cached verison will thus
almost certainly be available for immediate use, making the website
fast    for all users.    The optional ttl (in seconds) specifies how
long cache entries should live.    By default, it is twice the period.
If your function is called without    being cached, it is
synchronously executed (and the result stored for next    time), much
like the @cache decorator.    Periodic functions can have no required
arguments. While they can have    keyword arguments, no caching is
done if you call the function using them.    Uses the Django cache
(which uses Redis) to achieve a shared cache across    worker
processes.    In DEBUG mode, no caching is done.    Usage:
@periodic(60)        def get_blog_posts():

This will role out periodic without adding a cronjob yet to trigger the
functions. This *does* add a script to trigger them, though, and we can use
it to test in prod before adding the cronjob.

Without a cronjob, the periodic decorator acts just like the cache
decorator, so the site will still be equally fast (or equally slow,
depending on your mood).

I want to push this now, observe it in prod, and then push a cronjob if
everything goes smoothly.

This also adds a test page where you can see some statistics:

<https://camo.githubusercontent.com/7b468d88b38c003f168e007118cde6244338e077/687474703a2f2f692e666c756666792e63632f43486b3834483054647a516b70307a57564c35584e6c477a4a7878386e7252662e706e67>

This is purely informational and kind of useful. We can leave it on in prod
since it's read-only and nothing sensitive is in it.

The decorator is very easy to use; you can throw it on any function you'd
like (as long as it has no arguments), and it will automatically be added
to the list of background jobs. You can call that function just like you
normally would, and it will usually be instant.

This should make it easy for us to keep ocfweb fast without the same kind
of cronjob hacks we had in e.g. labstats.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/ocf/ocfweb/pull/105
Commit Summary

   - Add support for periodic functions
   - Add script for running periodic functions on-demand
   - component/blog: increase timeout 20x

File Changes

   - *A* ocfweb/bin/__init__.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-0> (0)
   - *A* ocfweb/bin/run_periodic_functions.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-1> (36)
   - *M* ocfweb/caching.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-2> (177)
   - *M* ocfweb/component/blog.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-3> (6)
   - *M* ocfweb/main/home.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-4> (13)
   - *M* ocfweb/settings.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-5> (1)
   - *A* ocfweb/test/periodic.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-6> (17)
   - *A* ocfweb/test/templates/periodic.html
   <https://github.com/ocf/ocfweb/pull/105/files#diff-7> (26)
   - *M* ocfweb/urls.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-8> (2)
   - *M* setup.py <https://github.com/ocf/ocfweb/pull/105/files#diff-9> (7)
   - *M* tests/caching_test.py
   <https://github.com/ocf/ocfweb/pull/105/files#diff-10> (48)

Patch Links:

   - https://github.com/ocf/ocfweb/pull/105.patch
   - https://github.com/ocf/ocfweb/pull/105.diff

—
Reply to this email directly or view it on GitHub
<https://github.com/ocf/ocfweb/pull/105>.

nattofriends:
I am not going to comment about the nature of what you are introducing here
and its relation to caching calls from the status blog.

2015-12-19 7:50 GMT-08:00 Matthew McAllister <notifications@github.com>:

jvperrin:
Looks good to me, I wasn't sure how this helped until I saw that the period was half the TTL (by default), so I think this will really help with these kinds of caching and loading time issues.

@matthew-mcallister What do you mean by staggered?

matthew-mcallister:
I just meant whether they would all update at once in a row, or whether it
would update staff hours, then 15 seconds later the status blog, etc. I
doubt it would be a problem, I don't expect Django will stall either way.

jvperrin:
Oh ok, that makes more sense, thanks.

chriskuehl:
I deployed this the test deb to prod and it seems to work great.

I'm going to merge and then probably work on improving the run script.

For now I'll add a cronjob that calls the script I added, but will improve this in the near future.

You can see status at:
https://www.ocf.berkeley.edu/test/periodic

chriskuehl:
reduced the total duration to about half (for a full http session), and reduced variation a lot as well

reduced page render time (as measured by ab on the same network) to 1/4

Issue #97, opened 2015-12-09
Mention user quota somewhere


chriskuehl:
I just raised quotas to 5 GB, but can't find any mention anywhere of the old quota to update. We should probably mention it somewhere...

chriskuehl:
assigning to @matthew-mcallister

matthew-mcallister:
Resolved in https://github.com/ocf/ocfweb/pull/110.

PR #110, opened 2016-01-17
Mention user disk quotas

matthew-mcallister:
Addresses https://github.com/ocf/ocfweb/issues/97, plus an unrelated commit. Not sure why I held onto this so long.


chriskuehl:
ship it

Issue #61, opened 2015-10-07
Docs: turn officers page into a template


matthew-mcallister:
Not as big a deal as the hosting badges page but it's large and messy right now, plus it gets updated regularly.

matthew-mcallister:
Resolved in https://github.com/ocf/ocfweb/pull/111.

PR #111, opened 2016-01-17
docs: Convert officers page into template

matthew-mcallister:
Addresses https://github.com/ocf/ocfweb/issues/61.


chriskuehl:
looks good

kpengboy:
Now it doesn't distinguish between when officers resigned midway through terms and when they jointly held the office.

And we still have to manually update this every semester. I'm not convinced yet that this is better...

chriskuehl:
Definitely seems better to me. Can you file tickets for (or just fix) the things you think are broken?

matthew-mcallister:
I was aware of that, and I don't think the distinction is necessary except
for historic interest—and that information can be found in the minutes.
Mid-term resignation has only happened a handful of times, and it's not at
all inaccurate to state that all persons involved were indeed site/general
managers during the semester, regardless of how the responsibility was
divided.

Also, short of scraping minutes, I don't know of any way to get around
manually entering the GM's and SM's names each semester.

matthew-mcallister:
Basically, the page is cleaner and easier to read without mixing in dates
with people's names.

PR #112, opened 2016-01-17
Make the login form look better

chriskuehl:

### DIFF
--- a/ocfweb/login/templates/ocf/login.html
+++ b/ocfweb/login/templates/ocf/login.html
@@ -1,26 +1,41 @@
 {% extends "base.html" %}
-{% block title %}OCF Login{% endblock %}
+{% load bootstrap %}
 
 {% block content %}
+    <div class="ocf-content-block">
+        <div class="row">
+            <div class="col-sm-8">
+                {% if error %}
+                    <div class="alert alert-danger" role="alert">
+                        {{ error }}
+                    </div>
+                {% endif %}
 
-{% if error %}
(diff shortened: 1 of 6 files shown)


ocfweb/login/templates/ocf/login.html
```diff
@@ -1,26 +1,41 @@
...
-    <input type="submit" value="Login" />
-</form>
-
+                <p>
+                    Sorry! Before you can proceed, we need to know who you are.
+                    Please enter your OCF username and password below to continue.
+                </p>
+                <form action="{% url 'login' %}" method="post">
+                    {% csrf_token %}
+                    <div style="max-width: 300px;">
```
jvperrin: Should this be moved into the stylesheets or just stay here?

jvperrin:
Looks great to me!

ocfweb/login/templates/ocf/login.html
chriskuehl: I'm not crazy about it either but I don't think moving it to a stylesheet helps much. It looks pretty silly if you let them be full width, and I don't think we can really have a standard "form width".

Granted, on all of these PRs, 300px has looked fine. But eventually we'll probably have others?

jvperrin: Ok, that makes sense, forms will likely be different sizes, so it shouldn't be limited. We can always change this if we encounter later on that we only have a couple common sizes or that kind of thing, but for now I would agree that it can just be left.

PR #113, opened 2016-01-17
Password form better

chriskuehl:
Based off my branch from #112.

PR #114, opened 2016-01-18
Vhost form better

chriskuehl:
based on #113 so it has some unrelated changes


jvperrin:
Looks great! One suggestion I have is to make it clear that *all* the checkboxes are required, because right now unless you submit the form it appears that they are part of the previous section and might be optional. I think just adding a header above the checkboxes like with the other fields and like the [old page](https://github.com/ocf/ocfweb/pull/114/files#diff-9e0f674a5d88014393a30911951bf44aL71) would help to clarify this. It's not much of a problem because if they try to submit the request they will quickly find out that the boxes are not optional, but I think it would help a bit to clear up any potential confusion.

chriskuehl:
@jvperrin okay, added a small label similar to the old form



The commit diff is weird because I can't change the base branch (thanks GitHub), look at 1c637723ba7aafa5bdcda7f5be93a72d5e4513a0 for my change

jvperrin:
@chriskuehl Ok, I think that looks a lot clearer what the checkboxes are for. Ship it!

chriskuehl:
oops merged into wrong branch

chriskuehl:
merged by hand in c24dbdfb82251aa6ec967541c79bea6499a5acee

PR #116, opened 2016-01-19
Improve registration form appearance

chriskuehl:


matthew-mcallister:
Testing went well judging by IRC, so ship it.

chriskuehl:
could still use some more design work, if somebody is interested please take it up

just want to get something shipped before tomorrow when we'll presumably have lots of people making accounts

PR #120, opened 2016-01-24
Put printer stats in a column

chriskuehl:
I think this looks a bit better and is a little easier to read.


jvperrin:
I was thinking of doing this too, looks good to me! :+1:

PR #123, opened 2016-01-24
Remove "move to MLK" page, redirect to lab page

chriskuehl:
Less code is better code?

Here's the slightly modified lab page:


jvperrin:
Looks great, I agree that we don't need the MLK page any more when the move is over and we have a page that shows all of the lab information that people will be looking for.

PR #131, opened 2016-02-01
Retry renaming csrf token

### LINKED ISSUE #130: CSRF token PR fails in prod due to referer checks
The cookies seem to be working properly, but referer checks are failing?

    Jan 31 18:46:57 coma gunicorn[23653]: Forbidden (Referer checking failed - https://www.ocf.berkeley.edu/account/register/ does not match any trusted origins.): /account/register/

Explained in the docs:
https://docs.djangoproject.com/en/dev/ref/csrf/#how-it-works

Maybe the referer header is not what we think it is due to the proxying?

I reverted my change in 89a8931ff0fe9e511905780a42be24a63b1d5c9a

jvperrin:
Fixes #130. Well, in theory anyway, it has to be tested first.

### DIFF
--- a/ocfweb/settings.py
+++ b/ocfweb/settings.py
@@ -164,6 +164,7 @@ if getuser() == 'ocfweb':
 
     CSRF_COOKIE_SECURE = True
     CSRF_COOKIE_DOMAIN = 'www.ocf.berkeley.edu'
+    CSRF_TRUSTED_ORIGINS = ['www.ocf.berkeley.edu']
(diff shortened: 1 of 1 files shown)


jvperrin:
Seems to work well, no errors in the log when I submit the form normally on production, and when I delete or modify the CSRF cookie, submission correctly gives a 403 error.

chriskuehl:
:+1: thanks for fixing this!

PR #138, opened 2016-02-02
Fix spurious CacheKeyWarnings. Close #135

### LINKED ISSUE #135: Spurious CacheKeyWarning messages in dev
```
/home/c/ck/ckuehl/ocf-proj/ocfweb/virtualenv_run/lib/python3.4/site-packages/django/core/cache/backends/base.py:244: CacheKeyWarning: Cache key contains characters that will cause errors if used with memcached: ":1:('2016.01.31.00.19-git182d1623', 'ocfweb.main.staff_hours#get_staff_hours', (), ())"
  CacheKeyWarning)
```

These are unimportant because we use Redis which allows basically anything as a key. These don't break anything, they just spam the console.

We should either rename our keys or silence the warning.

jvperrin:
This works to remove the `CacheKeyWarning` issue described in #135. However, it might not work if the key is the same for the first 240 characters, but differs later on in the key. If that's the case, then the cache will collide and return an incorrect result when fetched from the cache. I also tried limiting to 250 characters instead of 240, and it still said the cache key was over 250 characters, so I believe there are some extra characters being counted somewhere. If we wanted to use longer keys, or think that this might become a problem in the future, then we should just ignore the errors, as suggested as an option in #135.

### DIFF
--- a/ocfweb/caching.py
+++ b/ocfweb/caching.py
@@ -110,13 +110,15 @@ def _make_key(key):
     The returned key prepends a version tag so that we don't share the cache
     across ocfweb versions. This prevents strange behavior (e.g. if you change
     the return type of a function which was cached on the previous version).
+    The cache keys have spaces removed and are limited to under 250 characters
+    to preserve portability with memcached.
 
     :param key: some iterable key (e.g. a tuple or list)
     """
-    return tuple(chain(
+    return str(tuple(chain(
         [ocfweb_version()],
         key,
-    ))
+    ))).replace(' ', '')[:240]
 
 
 def _make_function_call_key(fn, args, kwargs):


chriskuehl:
the keys can be very long (they contain the entire markdown text and a bunch of stuff at the start) so collisions are somewhat possible

can we just disable the warning instead?

imo we shouldn't be bogged down by memcached limitations; we chose redis because we didn't want to use memcached, after all

jvperrin:
Yeah, I agree on that, collisions do seem to be a possible issue. Disabling the warning would work fine, and actually seems like a better option compared to trying to change the keys to work. It was good to try this approach, but it seems like too much pain for what it's worth to try and change the keys to be compatible with memcached. I'll disable the warning instead.

chriskuehl:
One option is to hash the keys before using them as cache keys, but it would result in unreadable keys during debugging which is probably not ideal.

Issue #136, opened 2016-02-01
Docs don't auto-reload in dev


chriskuehl:
Because we only read docs once, changes to the Markdown files require manually killing and restarting the server.

PR #141, opened 2016-02-04
Reload docs in development

jvperrin:
Addresses #136.

Unfortunately, reloading page _titles_ in development is not as easy, since they are stored in the documents themselves and require reading all of the markdown files every load to generate the proper trees. The current page title will update with these changes, but the same page's title shown on other pages will not change. Not sure if there is a better way to change that other than reading in all of the markdown files, but changing page titles should happen far less than changing their contents. New markdown files are also not recognized, so maybe reading in all the files would work well to fix these additional issues?

### DIFF
--- a/ocfweb/docs/markdown_based.py
+++ b/ocfweb/docs/markdown_based.py
@@ -1,6 +1,6 @@
 """Documents backed by Markdown.
 
-This is the most common type of doc. It reads a Markdown fil and renders it in
+This is the most common type of doc. It reads a Markdown file and renders it in
 a standard template.
 
 Markdown documents can specify the meta attributes:
@@ -16,6 +16,7 @@ import os
 from functools import partial
 from pathlib import Path
 
+from django.conf import settings
 from django.shortcuts import render
 
 from ocfweb.component.markdown import markdown_and_toc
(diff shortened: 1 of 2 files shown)


ocfweb/docs/markdown_based.py
```diff
@@ -56,5 +64,6 @@ def get_markdown_docs():
         yield Document(
             name='/' + name,
             title=meta['title'],
-            render=partial(render_markdown_doc, meta, text)
+            render=partial(render_markdown_doc, meta, text),
+            path=path,
```
chriskuehl: I think a better pattern is probably to add `path` as an argument to the `render_markdown_doc` function (e.g. as the first argument), then add it to the partial call. You don't need to modify `Document` at all this way.

chriskuehl:
I think this is a good approach. It's not easy to re-read the directories all the time because we purposefully read them only once (when the urls are imported). We'd have to use separate URL patterns in development which breaks the careful regex construction we do to prevent broken links.

ocfweb/docs/markdown_based.py
jvperrin: That's a good point, since `path` is only needed in the `render_markdown_doc` function. I'll change that.

chriskuehl: congrats on understanding the docs plumbing btw, it's quite convoluted :\

jvperrin: Yeah, I found that out pretty quickly, it's not as bad as it first appears, but there is a lot going on behind the scenes...

jvperrin:
@chriskuehl I changed what you suggested, it looks much cleaner now.

I think the ideal situation to solve this would be if we could change django's reloading feature in development to work for other files besides just python files. It uses `pyinotify` in linux, so it should be possible, but that's something that they don't currently support and probably won't. We could also have something that checks whenever a markdown file changes and then `touch`es a python file, but that seems like a bad idea too because it complicates things unnecessarily. I guess this will have to do for now, titles and new files excluded.

chriskuehl:
looks good, shipit



i agree about the django reloader; I had actually checked before to see if we could get it to watch *.md files as well and didn't find a good way either :\

PR #144, opened 2016-02-12
Add simple semester printing histogram

chriskuehl:
A first start at better printing stats in ocfweb

### DIFF
--- a/ocfweb/stats/printing.py
+++ b/ocfweb/stats/printing.py
@@ -1,0 +1,60 @@
+import pymysql
+from django.http import HttpResponse
+from django.shortcuts import render
+from matplotlib.figure import Figure
+
+from ocfweb.caching import periodic
+from ocfweb.stats.plot import plot_to_image_bytes
+
+
+# TODO: improve/move this after PyKota is gone
+CURRENT_SEMESTER_QUOTA = 100
+
+
+def pykota_connection(user='ocfpykota_user', password='not_a_secret'):
+    return pymysql.connect(
+        host='mysql.ocf.berkeley.edu',
+        user=user,
+        password=password,
+        db='ocfpykota',
+        cursorclass=pymysql.cursors.DictCursor,
+    )
+
+
+def stats_printing(request):
+    return render(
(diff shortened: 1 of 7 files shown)


ocfweb/stats/printing.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ocfweb.caching import periodic
+from ocfweb.stats.plot import plot_to_image_bytes
+
+
+# TODO: improve/move this after PyKota is gone
+CURRENT_SEMESTER_QUOTA = 100
+
+
+def pykota_connection(user='ocfpykota_user', password='not_a_secret'):
```
chriskuehl: as the password suggests, this is indeed not a secret password (it's read-only). it's already plaintext in puppet and any ocf user could read it straight out of pykota.conf.

ocfweb/stats/printing.py
```diff
@@ -0,0 +1,60 @@
...
+        content_type='image/svg+xml',
+    )
+
+
+@periodic(300)
+def _semester_histogram():
+    with pykota_connection() as c:
+        c.execute(
+            'SELECT GREATEST(0, `balance`) AS `balance` FROM `users` WHERE `balance` < %s',
+            (CURRENT_SEMESTER_QUOTA,),
```
chriskuehl: this is to exclude staff who give themselves a billion pages

jvperrin:
@chriskuehl This looks great to me! The navigation bar is actually really great for the future, because it allows for adding more stats sections later on. I was actually thinking of doing something similar in my project to use highcharts to show the lab usage and printing graphs.

Later on, maybe we can make a way to shift between days/weeks (arrows either side of the page perhaps?) to see past statistics for reference without having to open the images that make up the graphs in a separate page.

chriskuehl:
@jvperrin agreed, i've definitely wanted to do that for the other graphs

(for this one, we can't really since we don't have historical pykota data besides our own backups we make each semester. but hopefully our replacement for pykota should be able to handle that)

PR #145, opened 2016-02-12
Show minutes in lab hours when relevant

matthew-mcallister:
I think we didn't notice they were missing because we're used to the way they're presented now. It seemed like a good idea to hide the minute when we open/close unless either is not on the hour, i.e. 9:10am-6:00pm on Tuesday but 9am-9pm on regular days.


chriskuehl:
nice, thanks for fixing

noticed this the other day and meant to file an issue / fix it myself

PR #146, opened 2016-02-12
venv-update shinier

chriskuehl:
this is faster


chriskuehl:
the first build after you pull is going to take like 10 minutes because the cache location changed.

sorry. future builds will be faster than before, though, especially for commands like `make scss` or `make dev` which are usually no-op virtualenv-wise

PR #147, opened 2016-02-13
Add some simple account graphs

chriskuehl:
just a start


matthew-mcallister:
Really slow, but cool. The tabs are nice so far, except the label "OCF Stats" looks like it should be another tab.

chriskuehl:
yeah, especially in dev since it does the same ldap query for each graph

in prod it won't do that. ldap sux.

PR #152, opened 2016-03-22
Rename docs to help (rt#4635)

chriskuehl:


matthew-mcallister:
I see you've already done the work, but do you really think we should rename "docs" to "help" internally?

I think the only important change is replacing the copy that says "Documentation" with "Help", specifically the navbar menu item on the homepage. Changing all the URLs is not a big deal, but we will have to remember to change links [in other repos](https://github.com/ocf/utils/search?utf8=✓&q=docs) as well.

I just don't recommend changing `ocfweb.docs` to `ocfweb.help`. Now we have a mishmash of internal references where the code refers to them as "docs" while the directory is named "help". Plus, the symlink in the top directory is still called "docs". I would just leave it at renaming URLs and replacing copy while maintaining the package name and internal references to `docs`.

jvperrin:
I agree with @matthew-mcallister, although I do think renaming just the URLs but not the internal structure could be confusing for people trying to find things in the project structure. Maybe we could rename public documentation to help and leave our internal/staff documentation as docs to clarify the difference between the two?

chriskuehl:
I think let's just rename the text but keep the URLs for now.

I don't like the idea of having a package name not corresponding to the URL structure.

chriskuehl:
I reverted my commit (leaving it in history in case we change our minds later).

PR #155, opened 2016-04-04
Make commands look better

chriskuehl:
before:

after:

### DIFF
--- a/ocfweb/account/commands.py
+++ b/ocfweb/account/commands.py
@@ -5,6 +5,8 @@ from paramiko import AuthenticationException
 from paramiko import SSHClient
 from paramiko.hostkeys import HostKeyEntry
 
+from ocfweb.component.forms import Form
+
 
 def commands(request):
     command_to_run = ''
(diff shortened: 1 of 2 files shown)


jvperrin:
Looks good to me, the one thing I'm not sure about the note in the right sidebar. Do we want to imply that our website is less secure than connecting over SSH? It might be the case, but at the same time, I don't think we need to cast any of our services' security into doubt here. Thoughts?

chriskuehl:
Doesn't matter much to me, I'll change it.

Eventually I hope to destroy commands anyway.

jvperrin:
Sounds good. What would you replace commands with, buttons to create http and mysql spaces and statistics of pages and disk quota?

jvperrin:
(Like individual statistics shown in a sidebar or something when users sign in)

chriskuehl:
something like that, yeah. given that we have logging in already, it's weird to have another page which makes you enter username/password again

matthew-mcallister:
I like the idea of putting shell in a box in an iframe on https://ocf.io/ssh as well.

jvperrin:
Ok, both of those sound like great ideas to me.

PR #156, opened 2016-04-04
docs: create a live command reference page

matthew-mcallister:
@daradib suggested this several weeks ago. It seems like a modestly helpful improvement over logging directly into ssh.o.b.e from a browser.

screenshot - 04042016 - 03 48 54 pm

I duplicated the OCF command table to include some basic file manipulation help. I don't know if it's worth keeping, but, if we do so, the table could be made into a partial template and the reference section expanded.

Also, I couldn't get the iframe to properly scale while keeping aspect ratio, so this is my attempt at responsive.

### DIFF
--- a/ocfweb/docs/templates/commands.html
+++ b/ocfweb/docs/templates/commands.html
@@ -1,0 +1,93 @@
+{% extends 'base.html' %}
+
+{% block content %}
+  <div class="ocf-content-block">
+    <h2>Shell interface</h2>
+    <div class="shell-wrapper">
+      <iframe src="https://ssh.ocf.berkeley.edu" class="shell-frame">
+        <p>Sorry! Your browser can't display this window.</p>
+      </iframe>
+    </div>
+    <p>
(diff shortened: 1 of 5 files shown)


chriskuehl:
looks great except i think your html is broken in some parts:

ocfweb/docs/templates/commands.html
```diff
@@ -0,0 +1,93 @@
...
+                    <a href="{% url 'doc' command.doc %}">{{command.name}}</a>
+                  {% else %}
+                    {{command.name}}
+                  {% endif %}
+                </strong>
+                {{command.args}}
+              </code>
+            </td>
+            <td>
+              {{command.desc}}
```
matthew-mcallister: Culprit of the formatting bug; it should be `{{command.desc | safe}}` as it is above. This is why it should be a partial template if we keep it.

chriskuehl:
shipit assuming fixed :D

PR #157, opened 2016-04-08
cache home page staff hours harder

chriskuehl:


chriskuehl:
@matthew-mcallister says shipit

PR #168, opened 2016-08-18
Add vhost mail page

chriskuehl:
read-only. i'll do the rw stuff soon.

need a couple changes in ocflib:
* ~~crypt password generation~~
* ~~access to `last_updated` field~~

note that it's extremely unlikely a group has more than one vhost. currently none do except ggroup. but the UI will support it, even though it looks pretty derpy


matthew-mcallister:
I think this needs to support multiple "forward-to" addresses per receiving address.

chriskuehl:
I think this is ready to go @matthew-mcallister @jvperrin 

Some things to do in a follow-up:

* Support for multiple forward addresses
* Support for editing forwarding addresses (currently you must remove/add)
* Support for catch-all forwarding
* Add docs about how to use it
* Actually link to this from somewhere

The postfix config supports these already but not the UI.

To test, you can run the normal way. If you're using your staff VM, follow the instructions in this commit's message to get the relevant config: https://github.com/ocf/puppet/commit/eaa231b426923e2fadab6bac9193801f5b40d4aa

It uses a different database than prod so you can make any changes to it you like. The ggroup password was emailed out.

I've already put the config in prod.

Screenshots:

### empty state


### default state


### add modal


if you screw up:


### change password modal


### remove modal


### alert example

matthew-mcallister:
This is so cool. Just getting off the crummy .forward file system is cool, actually. I haven't tried every button yet, but it clearly works for the most part.

Some suggestions for the todo list:

- Add a menu item in the login drop down menu for group accounts (only those with mail vhost, preferably)
- In the password change modal, add text to indicated that an empty password field will unset the current password
- Replace the "X" icon for deleting addresses with something more obvious, e.g. a red button with a trash icon or the word "Delete"

I think this is good enough to ship as is.

PR #172, opened 2016-09-09
Add "edit forward to" modal (vhost mail)

chriskuehl:
Also moves some stuff around.

It looks pretty much the same as the add modal:


chriskuehl:
```
<ckuehl> with this i think the only remaining tasks are (1) editing the incoming address, easy; and (2) documentation
<ckuehl> and eventually figuring out how to do the migration i guess
```

PR #173, opened 2016-09-18
Add catch-all support for vhost mail

chriskuehl:
This is what it looks like if you haven't set a catch-all email:

Here's the modal where you add one (same as the normal add modal, but with the incoming address and password fields omitted):

Here's what it looks like when you have a catch-all email:

(The popovers obviously start closed, I just opened them for the sake of the screenshots.)


chriskuehl:
I'm gonna merge as-is to unblock myself, but comments welcome.

PR #174, opened 2016-09-28
Add vhost mail docs and link (rt#5219)

chriskuehl:
Comments welcome. Leaving this PR open until we're ready to launch, then I'll merge it right before we do the transition.

One note: we don't have a way to include images in Markdown that are hosted in the app itself currently, so the images are on fluffy.cc. We should fix that eventually but I don't think it's a blocker.

### DIFF
--- a/ocfweb/account/templates/account/vhost_mail/index.html
+++ b/ocfweb/account/templates/account/vhost_mail/index.html
@@ -122,7 +122,7 @@
                 Mail virtual hosting is a free service from the friendly
                 student volunteers of the Open Computing Facility. Interested
                 in signing up?
-                <a href="{% url 'doc' 'contact' %}">Get in touch with us!</a>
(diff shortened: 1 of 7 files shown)


chriskuehl:
retest this please

jvperrin:
Nice docs, the instructions look really clear to me! I don't think I have any suggestions really, it all looks great. The gmail instructions are the most important, since I imagine that's what the vast majority of students will be using.

Can the [vhost migration page](https://www.ocf.berkeley.edu/docs/staff/backend/vhost-mail/) be removed now that this new system is about to be put in place, or are there still relevant things on there?

chriskuehl:
oh yeah, good catch.

better than removing would probably be to update the page to match what we actually did (the plans on the page are slightly different than the actual setup) and then it turns into documentation :D

i'll do that

jvperrin:
Alright, sounds like a good plan!

PR #176, opened 2016-10-08
Minor fixes to some outdated docs

chriskuehl:


matthew-mcallister:
vhost page looks good.

I can't speak to changes in labstats, but but I've never had to generate a cert for a new desktop, so I presume it is no longer necessary.

PR #188, opened 2016-10-26
Initial from vhosting_check_script

hexecute:
Yes, the filenames are a mess. They can easily be changed.

You'll notice stuff like "binnov-157x46.gif", "binnov-157x46.png", and "binnov-157x46.gif.png" (lolwut). Unfortunately, these are actually different files (diff-wise) that are in use in multiple OCF websites. It seems like someone distributed these in the past, so we may have to keep them around unless there's a better solution.


chriskuehl:
I don't understand why we're adding these to ocfweb? We already have the ones we want on the website in `ocfweb/static/img/hosting-logos`.

hexecute:
For the vhosting script, I'm looking at webpages and comparing their banners to accepted banners (i.e. banners here). These are banners that were distributed by the OCF in ages past and are in use on older sites. If we don't compare these, then we have to either handle each old site manually forever or force old sites to update old banners to newer ones.

chriskuehl:
Since we're checking the image data (and not the pixels), can you just embed hashes in the script? That would solve the problem of needing the image files around.

hexecute:
Comparing images hashes is a good idea. I still think we need to have the images exist somewhere in OCF though for reference/documentation/archival purposes. I had to go and download these images from group websites since I couldn't find them anywhere else. Is there a better place for them?

matthew-mcallister:
> Comparing images hashes is a good idea. I still think we need to have the images exist somewhere in OCF though for reference/documentation/archival purposes. I had to go and download these images from group websites since I couldn't find them anywhere else. Is there a better place for them?
 
The staff directory, which will someday be checked into git.

chriskuehl:
yeah, ~staff sounds fine if you think they're worth saving (i don't actually care about historical junk, but other people probably will)

the main thing is not to make ocfweb a dumping ground

matthew-mcallister:
The staff directory, which will someday be checked into git.

PR #193, opened 2016-11-06
settings: add local host to ALLOWED_HOSTS in debug mode

kpengboy:
Django 1.10.3 introduced a [change in the behavior of ALLOWED_HOSTS][1]
that breaks our debug mode. This change allows `make dev` to continue
launching a usable website, by listing the local host as an allowed host
in debug mode.

[1]: https://docs.djangoproject.com/en/1.10/releases/1.10.3/#dns-rebinding-vulnerability-when-debug-true

### DIFF
--- a/ocfweb/settings.py
+++ b/ocfweb/settings.py
@@ -1,5 +1,6 @@
 import configparser
 import os
+import socket
 import warnings
 
 from django.core.cache import CacheKeyWarning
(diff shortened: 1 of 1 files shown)


chriskuehl:
nice fix

jvperrin:
Nice, looks good to me!

PR #194, opened 2016-11-06
[WIP] Add account policy and banning policy

kpengboy:
@chriskuehl @jvperrin @matthew-mcallister and everyone else who reads this

This commit adds these two policies BoD recently approved. I've made a special view for the account policy so we can keep on including it verbatim on the sign-up page. Please review for quality etc.

When this gets merged in, we may have to prominently notify members via homepage, lab startup page, tsunami motd, etc. (We may be required to do so for the same legal reasons other online services have to loudly inform you when they change their terms of use. But then again, IANAL...)


chriskuehl:
lgtm, idk anything about the policy but i assume it's the one BoD approved

ocfweb/docs/templates/docs/partials/account-policies.html
```diff
@@ -0,0 +1,31 @@
...
+        <li>Using electronic mail or communication to harass other users or send bulk unsolicited communications.</li>
+        <li>Posting materials on electronic bulletin boards that violate existing laws or the University's codes of conduct.</li>
+        <li>Attempting to monitor or tamper with another user's electronic communications, or reading, copying, changing, or deleting another user's files or software without the explicit agreement of the owner.</li>
+        <li>Using OCF or University equipment or resources for commercial purposes unless specifically authorized.</li>
+        <li>Failing to follow common rules of etiquette while in the lab or while using OCF services.</li>
+    </ol>
+</p>
+<p>Violations of this policy or any other OCF policy may result in account suspension. Additionally, misuse may result in suspension or termination of computing privileges. Furthermore, OCF staff may disable an account and/or ban a person from the OCF if there exist exigent circumstances which require such actions.</p>
+<p>The banning of any person from the OCF will be subject to the <a href="/docs/membership/banning/">User Banning Policy</a>.</p>
```
jvperrin: Maybe add line wrapping in this file? Some of these lines are really long.

matthew-mcallister:
We should probably link to the account policy page from the account page (/docs/services/account).

ocfweb/docs/templates/docs/partials/account-policies.html
kpengboy: We're pretty inconsistent with our line wrapping in our docs. Do you think we should always line wrap? Given the existing inconsistency, I'm fine either way...

jvperrin: I think we should if the lines are this long and probably should be doing it for other files too. It just makes editing and reading a lot easier, and if it doesn't affect the page's layout (and it shouldn't with HTML), then I think it's a good idea.

kpengboy:
For the future, we can add the sidebar to the account policies document, or try to rewrite the document in Markdown.

PR #205, opened 2016-11-15
docs: linewrap everything

matthew-mcallister:
It makes paragraphs with long URLs look kind of derpy, but this formatting is
the divine decree of vim's linewrap feature (`gq}`).

This touches a lot of files. It would be very useful for others to check for broken headers, indentation, etc.


matthew-mcallister:
Cool, I'm sure errors will be spotted faster this way.

PR #208, opened 2016-11-19
Only add a reverse trigger on ocflib if we're master

chriskuehl:
Currently if you push to ocflib, it rebuilds every single ocfweb branch, which is probably unnecessary.


chriskuehl:
@jvperrin

PR #209, opened 2016-11-19
Pass ocflib version as argument to docker build

### LINKED ISSUE #192: figure out how better to handle unpinned ocflib
Currently we use `--no-cache` when building all the Docker images, which is really really slow.

The reason is that otherwise the image gets cached up to (and including) this line: https://github.com/ocf/ocfweb/blob/c0bf5228accb7a7b5bbc963426213b944dd3fbb9/Dockerfile#L32-L35

We should probably do something like install everything besides ocflib in that layer, then install the latest ocflib in the next layer (which we always run via some Docker cache-busting trick).

The slow part is installing all the system packages at the start, and then all the Python packages later on. Installing ocflib and everything else is very fast. So if we can cache everything up until then, builds will normally take only seconds instead of the minutes they take now.

chriskuehl:
Fixes #192

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,9 @@ RUN virtualenv -ppython3 /opt/ocfweb/venv \
     && /opt/ocfweb/venv/bin/pip install \
         -r /opt/ocfweb/requirements.txt
 
+ARG ocflib_version=
+RUN /opt/ocfweb/venv/bin/pip install ocflib${ocflib_version}
+
 COPY bootstrap-scss /opt/ocfweb/bootstrap-scss/
 COPY manage.py /opt/ocfweb/
 COPY ocfweb /opt/ocfweb/ocfweb/
--- a/Makefile
+++ b/Makefile
@@ -22,10 +22,11 @@ Dockerfile.%: Dockerfile.%.in
 
 .PHONY: cook-image
 cook-image: Dockerfile.web Dockerfile.worker Dockerfile.static
-	docker build --no-cache -t $(DOCKER_TAG_BASE) .
-	docker build --no-cache -t $(DOCKER_TAG_WEB) -f Dockerfile.web .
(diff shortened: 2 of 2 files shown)


matthew-mcallister:
Nice solution.

I'm fine with parsing json from pypi. The only alternative I can think of would be for the target to install ocflib in a venv and run a script which just spits out the version number, but that seems very roundabout.

jvperrin:
This does have some problems, as in it fails when pypi returns something that `jq` can't parse (maybe a 503 or something), as in [these](https://jenkins.ocf.berkeley.edu/job/ocfweb/job/master/286) [builds](https://jenkins.ocf.berkeley.edu/job/ocfweb/job/master/285), and also [these](https://jenkins.ocf.berkeley.edu/job/ocfweb/job/master/283) [two](https://jenkins.ocf.berkeley.edu/job/ocfweb/job/master/281). pypi does seem to have quite a few errors for whatever reason:

pypi-flakiness
(from https://status.python.org/#week)

Maybe we could have a fallback to the most current version if we can't fetch the version number? It might then get cached, but that sounds better than failing the whole build to me, especially since we remove docker images so frequently and aggressively on reaper that it is unlikely to be cached for very long. We also now trigger ocfweb builds when ocflib is pushed, so I don't think the caching would be as big of a problem as before.

chriskuehl:
The only thing i worry about is that it might start failing consistently and we won't notice. That's usually the problem with adding fallbacks. They're fine for isolated incidents, but might mean things can break for a long time without anybody noticing.

Maybe we can read from the new PyPI instead? https://pypi.org/pypi/ocflib/json

Presumably that might be more reliable than the old one...

We could also set up our own PyPI (other benefit: way faster install times, since we could use wheels for packages with compiled bits), but that's probably more work than it's worth.

PR #210, opened 2016-11-22
add option #4 - Slack to IRC documentation

tahabi:
on dkessler's suggestion

### DIFF
--- a/ocfweb/docs/docs/contact/irc.md
+++ b/ocfweb/docs/docs/contact/irc.md
@@ -8,7 +8,7 @@ discussion-type questions.
 We normally chat in the `#rebuild` channel. For historical reasons, `#ocf` is
 mostly for non-OCF-related discussion.
 
-You have three simple options for chatting:
+You have four simple options for chatting:
 
 ### Option 1: Use our web client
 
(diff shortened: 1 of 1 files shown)


dkess:
Thanks for doing this-- only nitpick is that the line should be wrapped to 80 chars to match the rest of the file (and the rest of the docs).

chriskuehl:
Do we want to advertise this to non-staff?

tahabi:
I guess you did say that a benefit of IRC was the barrier to entry for non-techies, but so few outside of staff even come to IRC via thelounge that I don't think it will be a big deal. Perhaps move this to the staff documentation instead of the user docs?

dkess:
+1 to moving to staff documentation. Slack is really for people who want to stay active in IRC, not people dropping in for a one-time question.

chriskuehl:
> I guess you did say that a benefit of IRC was the barrier to entry for non-techies

heh, i was mostly being facetious

i probably just wouldn't advertise it to non-staff until it's reliably working (currently @jvperrin has to restart the bridge, and it sometimes drops messages)

Issue #212, opened 2016-11-23
Docs for mesos/marathon


chriskuehl:

chriskuehl:
We have a nice start for these docs: https://www.ocf.berkeley.edu/docs/staff/backend/mesos/

I'd still like to add subpages for:

* Marathon
* Mesos
* Zookeeper
* Mesos DNS (maybe not worth a whole page, idk)

Overall I think the intro page (linked above) explains pretty well how these fit together. The sub pages would just include specific troubleshooting information, links to the dashboards, etc. 

Eventually Marathon might be worth some extra documentation for working with services (or maybe that should go elsewhere, idk)

PR #213, opened 2016-11-24
Add a first draft of Mesos documentation

chriskuehl:
This adds a high-level overview of how Mesos/Marathon/Zookeeper fit together in general (and at the OCF).

The plan is to later add sub-pages for each of these which discuss how to troubleshoot each component, as well as giving links to web UIs and general "how-to" topics.

Marathon would probably need more information about actually writing apps to be deployed by it, and how to deploy them.

### DIFF
--- a/ocfweb/docs/docs/staff/backend/mesos.md
+++ b/ocfweb/docs/docs/staff/backend/mesos.md
@@ -1,0 +1,202 @@
+[[!meta title="Mesos"]]
+
+At the OCF, we use [Mesos][mesos] and [Marathon][marathon] in order to deploy a
+number of important services. This page will try to document how all of these
+pieces fit together. The sub-pages are primarily troubleshooting steps and
+instructions for working with each individual component.
+
+
+## Mesos
+
+**[Apache Mesos][mesos]** is software originally developed at Berkeley which
+helps to schedule applications running on clusters of computers.
+
+From reading the Mesos website, it's difficult to tell exactly what Mesos does
+and does not do. The best way to think of Mesos is as a *scheduler*, which
+determines how to run a set of applications across a cluster of computers.
+
+
+### A Mesos cluster
+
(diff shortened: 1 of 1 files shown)


chriskuehl:
This is a start for #212

ocfweb/docs/docs/staff/backend/mesos.md
```diff
@@ -0,0 +1,202 @@
...
+
+At the OCF, we have three Mesos masters:
+
+* `mesos0` (`whirlwind`, running on `jaws`)
+* `mesos1` (`pileup`, running on `pandemic`)
+* `mesos2` (`monsoon`, running on `hal`)
+
+At any given time, a Mesos cluster has only one *leader*, which is always one
+of the Mesos masters. At any given time, only one of the masters is the leader,
+and the other two masters are effectively doing no real work (but keeping track
```
Baisang: so the other two masters are just there to ensure higher availability? (i.e. one takes over when leader dies)

Baisang:
i assume eventually we would also have docs just in general about containers and containerization? because that's something I think a lot of staff have less experience with as well.

though, this doc does a good job explaining things at that higher level without having to talk explicitly about containers and things like that, I think people will just be curious when they click on e.g. the Marathon link and see that it is "A container orchestration platform..."

chriskuehl:
Probably we need some. But it's also hard to figure out where to draw the line at not reproducing documentation in our wiki that is well-covered elsewhere. This doc makes sense because it's about our infrastructure in particular, but we're not really using Docker any differently than other folks.

chriskuehl:
(granted, some of the stuff in this document is definitely covered elsewhere, but I find the Mesos introduction really hard to parse -- it's difficult to figure out exactly what it's doing (what does "Program against your datacenter like it’s a single pool of resources" even mean?))

PR #215, opened 2016-12-01
Add option to disable pages depending on create

chriskuehl:
This lets us turn off create based on a config option if create is broken for some reason.

Obviously we want to minimize this time, but it's useful when testing changes to create, since we can make sure users won't get a broken experience.

### DIFF
--- a/ocfweb/account/chpass.py
+++ b/ocfweb/account/chpass.py
@@ -11,6 +11,7 @@ from ocfweb.account.constants import TESTER_CALNET_UIDS
 from ocfweb.auth import calnet_required
 from ocfweb.component.celery import change_password as change_password_task
 from ocfweb.component.forms import Form
+from ocfweb.component.lab_status import create_required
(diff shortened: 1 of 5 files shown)


matthew-mcallister:
I agree this is a good idea. However, would you prefer to set this option somewhere besides `lab_status.yaml`? Maybe find a way to set this from the Marathon application definition?

I'm halfway through putting all our config files into a repo with tests, so I'm not worried about the option accidentally being left on, I just wonder if it shouldn't be set in a file everyone can change.

chriskuehl:
I think it belongs in a config file. Marathon is really just an implementation detail.

But yeah, I agree we should probably rename `lab_status.yaml` to something more general. It already contains stuff that's not just lab-specific.

We should probably chat about the configs repo. I've got a few thoughts about it but nothing really concrete -- it might help to try to figure out what our requirements are.

PR #217, opened 2016-12-04
Add logging to Docker services

chriskuehl:


chriskuehl:
hmm this works but the stream name isn't working (it arrives as empty string, which rsyslog turns into `_`), i think it has to do with logger changing on jessie?

chriskuehl:
lol amazing, the protocol totally changed in util-linux 2.26: http://man7.org/linux/man-pages/man1/logger.1.html

chriskuehl:
fixed in https://github.com/ocf/puppet/commit/a249a4ce1d88b761c8599527e48dc678327a72f4

PR #219, opened 2016-12-10
First Creation of RT Documentation

kevinmasd:
Assigned by RT 5636.

### DIFF
--- a/ocfweb/docs/docs/staff/backend/rt.md
+++ b/ocfweb/docs/docs/staff/backend/rt.md
@@ -1,12 +1,12 @@
 [[!meta title="Request Tracker"]]
 
-[**Request Tracker**](https://rt.ocf.berkeley.edu/) is the ticketing system 
-used by the OCF. It is the main way of keeping track of OCF-related activity. 
-Tickets are currently made either by staff to propose things to talk about 
+[**Request Tracker**](https://rt.ocf.berkeley.edu/) is the ticketing system
+used by the OCF. It is the main way of keeping track of OCF-related activity.
+Tickets are currently made either by staff to propose things to talk about
 or by emails from the Hostmaster or Help mailing lists.
 
 ## Queues
-Tickets are assigned to queues, or organized boards. Manually-created 
+Tickets are assigned to queues, or organized boards. Manually-created
 tickets are found under:
 - *bod* for meeting topics
 - *bureaucracy* for officer-related issues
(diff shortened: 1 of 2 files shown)


ocfweb/docs/docs/staff/backend/rt.md
```diff
@@ -0,0 +1,57 @@
+[[!meta title="Request Tracker"]]
+
+[**Request Tracker**](https://rt.ocf.berkeley.edu/) is the ticketing system
+used by the OCF. It is the main way of keeping track of OCF-related activity.
+Tickets are currently made either by staff to propose things to talk about
+or by emails from the Hostmaster or Help mailing lists.
```
chriskuehl: Or security, bod, etc.

I'd probably say "Tickets are automatically created when emails are received at the queue's name, e.g. help@ocf.berkeley.edu. Staff can also create tickets by logging in to the web UI".

ocfweb/docs/docs/staff/backend/rt.md
```diff
@@ -0,0 +1,57 @@
...
+- *operations* for Operations Strategist work (opstaff)
+- *projects* for long-term activities
+- *techtalks* for Tech Talk ideas and organization
+- *todo* for shorter-term items
+
+Other tickets are created automatically from emails:
+- *devnull* for devnull@ emails
+- *help* for questions from users (help@)
+- *hostmaster* for requests for virtual hosting ([form](https://www.ocf.berkeley.edu/account/vhost/))
+- *security* for vulnerability detections from ISP
```
chriskuehl: i don't think this distinction is really accurate, i seldom use the web UI and instead just send emails

as-written this might confuse people into thinking they're set up specially somehow

probably better to just say the thing in my comment above

ocfweb/docs/urls.py
```diff
@@ -57,6 +57,7 @@ def doc_name(doc_name):
     assert re.match(r'^/[a-zA-Z\-/]+$', doc_name), 'Bad document name: ' + doc_name
     return doc_name[1:].replace('-', '\-')
 
+
```
chriskuehl: thanks, this was necessary because a new rule was added to the latest flake8

ocfweb/docs/docs/staff/backend/rt.md
```diff
@@ -0,0 +1,57 @@
...
+to some mishaps. In the RT interface, *Reply* directly communicates with the poster, so look for the
+last communication with the ticket opener. *Comment* doesn't directly communicate and is generally for
+internal discussion. This can also be done through email, as RT defaults the reply-to field with the
+queue mailing lists. Be careful here though: to comment through email, send the email to {queue}-comment
+(i.e. help@ vs help-comment@). Also make sure that your reply does not include any of the comments, as in
+make sure the trimmed comment is all the information you want released.
+
+### Creation
+They can be manually created through the *New Ticket in* button on the top right of the page. If doing
+so to communicate to people outside of the OCF, add their email to the requestors field of the ticket.
```
chriskuehl: when creating tickets in order to communicate with outsiders, it's necessary to:

1. create a new ticket in the web UI, specifying them as the requestor. leave the body of the ticket entirely blank
2. reply to the newly-created ticket, including your initial email in the body

this is necessary because the first step doesn't generate an email to the requestor (but does email staff); the second step will email them

might be worth documenting this somewhere since it confuses people sometimes

ocfweb/docs/docs/staff/backend/rt.md
```diff
@@ -0,0 +1,57 @@
...
+Tickets may reference each other or there may be redundant tickets. If so, ticket relationships and merging
+can be done under *Links* for the ticket you want to keep/set relations for.
+
+Tickets are to be as *resolved* if done, *rejected* if not done for reasons, *new* on creation, and *open* otherwise.
+
+### Searching
+By default, the queues only show *open* or *new* tickets. To see other tickets, either search the ticket number in the
+top right or use *New Search* to do more advanced searching. If using the latter, don't forget to press either *Add these
+terms and Search* or *Update formate and Search*. Search arguments can also be saved for later use, as seen with the
+ocfstaff saved searches (bother a staff member to see these searches).
```
chriskuehl: might be worth mentioning somewhere the different statuses for tickets:

* `new`: a new ticket with no response from staff
* `open`: a ticket which staff have responded to
* `stalled`: a ticket which is marked by staff to resolve automatically in 5 (?) days if there is no response; any response by the requestor will move the ticket back into "open" state
* `resolved`: a ticket which was resolved by staff
* `rejected`: a ticket which staff rejected, often without comments
* `deleted`: a ticket which staff deleted; unlike rejected, these tickets don't appear in searches. only use delete on things which are obvious spam, and generally not on tickets which have any comments/replies

chriskuehl: mmm this regressed when we moved to marathon, i never duplicated that cronjob:

https://github.com/ocf/puppet/commit/30cbd462e5d21a0cfc547d576c030dbf5d4d7d6a#diff-59eb65743f2edbbb49d61bb1102b3c81L53

that's okay though, i always thought that was a dumb feature (i sent some emails a while back about how i'd use stalled)

PR #221, opened 2016-12-13
Add table of toner changes

chriskuehl:
Our data is bad on the new printers, so the charts aren't too great for them. deforestation is missing data since I only imported data for current printers into MySQL, I'll fix that soon.

### DIFF
--- a/ocfweb/stats/printing.py
+++ b/ocfweb/stats/printing.py
@@ -1,6 +1,7 @@
 from django.http import HttpResponse
 from django.shortcuts import render
 from matplotlib.figure import Figure
+from ocflib.lab import stats
 from ocflib.printing.quota import get_connection
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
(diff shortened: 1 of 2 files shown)


ocfweb/stats/printing.py
```diff
@@ -44,3 +46,53 @@ def _semester_histogram():
...
+                A.value as pages_before,
+                B.value as pages_after
+                FROM
+                    ordered1 as A,
+                    ordered2 as B
+                WHERE
+                    B.position = A.position + 1 AND
+                    B.value > A.value AND
+                    A.value > 0
+        ''')
```
jvperrin: This seems overly complicated, I think it would be better to just show a graph of the information and then let the person viewing it figure out when toner changes were made, since they sometimes aren't even full refills. It's also harder with a table to get a sense of time between rows, since they are just dates, which a graph would help with. Depends on what purpose the statistics are being shown for though I suppose.

chriskuehl:
yeah, still planning to add graphs -- but that's harder

jvperrin:
Alright, cool

PR #222, opened 2016-12-14
Improve requirements upgrading a bit

chriskuehl:
Also pin our dev dependencies, which is nice because venv-update can be
much faster (no-network installs, except for unpinned ocflib of course).


jvperrin:
Nice, looks good! Not that I know too much about Python's dependency management, but it looks better organized and simpler to work with, so that's always good.

chriskuehl:
There are some opinions. This is pretty much how I feel: https://github.com/Yelp/requirements-tools/tree/d3b55f7cdbd91ade14489b6a1e94f6f1d977ee36#our-stance-on-pinning-requirements

PR #223, opened 2016-12-14
Add pages printed per day

chriskuehl:
looks kinda bad, but it's a giant chart, so idk

someone can improve later maybe

i spot-checked that the numbers match stats.ocf.b.e

### DIFF
--- a/ocfweb/stats/printing.py
+++ b/ocfweb/stats/printing.py
@@ -1,7 +1,12 @@
+from collections import defaultdict
+from datetime import date
+from datetime import timedelta
+
 from django.http import HttpResponse
 from django.shortcuts import render
 from matplotlib.figure import Figure
 from ocflib.lab import stats
+from ocflib.printing.printers import PRINTERS
 from ocflib.printing.quota import get_connection
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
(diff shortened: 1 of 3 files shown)


chriskuehl:
shoot we need a total column

jvperrin:
Highcharts can make this nicer with a line chart with multiple lines on it, pretty much like the current oracle, but with the page information instead of toner. I think it would look pretty cool anyway, something like this but with printers instead: http://www.highcharts.com/stock/demo/compare

chriskuehl:
i would still want a chart, but a graph is a nice addition

jvperrin:
What's the difference?

jvperrin:
Image vs JS?

chriskuehl:
i don't mind image vs JS, i just want a table of numbers to go along with the fancy graph (that's what i mean by chart in case that was ambiguous)

e.g. on stats.ocf.b.e we almost always use the table rather than the graphs (partially because the graphs are slow and bad, but still, there are some use-cases where a table is better at giving data)

chriskuehl:
i'm all for highcharts or whatever if it means we can get rid of matplotlib and numpy :)

jvperrin:
Oh, sure, I agree on that, I thought you meant some other way of visualizing the data, not a table. I definitely would want to keep around the table, since it's easy to read and has the exact numbers.

I haven't worked on highcharts for a few months now, but it pretty much just needs some helper functions and it'll be good to go. It also needs to be expanded to cope with the large amounts of data it can get, but that shouldn't be terrible, it just involves a readjustment of the periodic functions to not spew out too much data. It would also be nice to be able to look at historical data, but that's a separate project really.

PR #224, opened 2016-12-14
Create pdf-open.md

tahabi:
Since the command seems undocumented, figured it would be a good idea to document it for future reference.


chriskuehl:
+1 nice

sorry about the whitespace bugs :\

PR #225, opened 2016-12-16
ignore this just testing jenkinsfile

chriskuehl:


jvperrin:
Branch name on point

chriskuehl:
ffs it has an environment variable for `SVN_REVISION: Subversion revision number that's currently checked out to the workspace, such as "12345"` but not one for git SHAs

jvperrin:
It should have the commit author, but we'd need to make an API call to check if the person submitting the pull request is a collaborator or not, and I'm not sure if that would be possible within a Jenkinsfile...

chriskuehl:
I'm just trying to figure out how to get the git sha so I can add it to the Marathon version to aid debugging.

I'm pretty sure it's not possible to fix the security problem at all using the Jenkinsfile, since a malicious user could just change the entire Jenkinsfile in their branch. (Unless the master Jenkinsfile controls which branches get built or something?)

jvperrin:
The master Jenkinsfile is used for untrusted pull requests, their changes are not used apparently. Haven't tested it, but that's how it's supposed to work according to what I have read.

chriskuehl:
I wonder how it even figures out who is trusted? I couldn't find any kind of list of trusted users.

jvperrin:
I think it just knows collaborators vs. other people? Not sure on that

PR #230, opened 2017-01-24
some style/grammar changes

tahabi:


ocfweb/about/templates/about/staff.html
```diff
@@ -93,16 +93,16 @@
     <div class="ocf-logos">
         <div class="container">
             <div class="row">
-                <h2>Our lab and servers run free and open-source software.</h2>
+                <h2>Our lab and servers run free and open-source software, like:</h2>
```
tahabi: I think "such as:" would also be a good alternative to the period alone as well

ocfweb/about/templates/about/staff.html
```diff
@@ -50,14 +50,14 @@
...
-                <h2>OCF staff help to provide cool services to the Berkeley community.</h2>
+                <h2>OCF staff help provide cool services to the Berkeley community.</h2>
             </div>
             <div class="row">
                 <div>
                     <div class="col-sm-4">
                         <p><img src="{% static 'img/about/web.png' %}" alt="web hosting" /></p>
                         <h3>Web &amp; Email Hosting</h3>
-                        <p>We provide free web and email hosting to our more than 40,000 members, both students and student groups.</p>
+                        <p>We provide free web and email hosting to more than 40,000 of our members, both students and student groups.</p>
```
matthew-mcallister: This changes the meaning. The fact is we _have_ more than 40 000 members, not that we provide web/email hosting to all of them.

tahabi: Makes sense, but the original phrasing still seems awkward. Is there a better way of expressing the original sentiment?

matthew-mcallister: Maybe "We provide free web and email hosting to more than 40,000 OCF members".

PR #231, opened 2017-01-24
add staff_in_lab_count

tahabi:
If I haven't made some serious error I think these three PRs take care of the issue @jvperrin mentioned.


tahabi:
I guess this won't pass Jenkins until the earlier PRs get acted on

PR #234, opened 2017-01-27
add anchors to server rows

tahabi:
I found myself often going to /docs/backend/servers/ to recall which VM corresponded to which task, this just adds a nice jump directly to the server name. For example, 

http://nyx.ocf.berkeley.edu:8934/docs/staff/backend/servers/#supernova

### DIFF
--- a/ocfweb/docs/templates/docs/partials/host.html
+++ b/ocfweb/docs/templates/docs/partials/host.html
@@ -3,10 +3,11 @@
 
 <div class="row host host-{{host.type}}">
     <div class="col-sm-3">
-        <a id="{{host.hostname}}" class="server-anchor"></a>
-        <p class="hostname">{{host.hostname}}</p>
-        <p class="ip">{{host.ipv4}}</p>
-        <p class="ip">{{host.ipv6}}</p>
+      <a id="{{host.hostname}}" class="server-anchor" href="#{{host.hostname}}">
+    <p class="hostname">{{host.hostname}}</p>
+      </a>
+      <p class="ip">{{host.ipv4}}</p>
+      <p class="ip">{{host.ipv6}}</p>
     </div>
     <div class="col-sm-2">
         <p class="type">{{host.english_type}}</p>
--- a/ocfweb/docs/templates/docs/partials/host.html
+++ b/ocfweb/docs/templates/docs/partials/host.html
@@ -3,7 +3,7 @@
 
 <div class="row host host-{{host.type}}">
     <div class="col-sm-3">
-        <a name="{{host.hostname}}" class="server-anchor"></a>
+        <a id="{{host.hostname}}" class="server-anchor"></a>
         <p class="hostname">{{host.hostname}}</p>
         <p class="ip">{{host.ipv4}}</p>
         <p class="ip">{{host.ipv6}}</p>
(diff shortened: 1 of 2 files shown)


dkess:
Probably insignificant, but according to [this](http://stackoverflow.com/questions/484719/html-anchors-with-name-or-id) you should use `id=` instead of `name=` for anchors.

ocfweb/static/scss/pages/docs/_servers.scss
```diff
@@ -69,6 +69,13 @@
                     }
                 }
             }
+
+            .server-anchor {
+                display: block;
+                position: relative;
+                top: -75px;
```
jvperrin: Where does this `-75px` come from? Does this work with the responsive design of the page? I'm assuming it does, but I'm just not sure where this number comes from.

ocfweb/docs/templates/docs/partials/host.html
```diff
@@ -3,6 +3,7 @@
 
 <div class="row host host-{{host.type}}">
     <div class="col-sm-3">
+        <a id="{{host.hostname}}" class="server-anchor"></a>
```
jvperrin: Could this link be on the hostname below so that when the name is clicked on it then changes the URL to go directly to that host? Seems like it would be a nice way to get a link to the host.

ocfweb/static/scss/pages/docs/_servers.scss
chriskuehl: not sure either, i imagine you just want the same thing regular docs does?

https://github.com/ocf/ocfweb/blob/74c5e192d126bb73ba79eef81f3012608328588e/ocfweb/static/scss/pages/_docs.scss#L20-L27

jvperrin: Ooh, I didn't know that was a thing, nice

tahabi: The 75px was just a number I picked while testing and it happened to work nicely on my screen but obviously doesn't work with the responsive layout. I've copied that code from _docs.scss into my .server-anchors class but not knowing sass too well I don't know if there's a more DRY way of doing it

tahabi:
not sure why the build is failing. something to do with caching?

chriskuehl:
The actual error isn't the caching, it's the traceback below it:

```
  File "/opt/jenkins/slave/workspace/ocfweb/PR-234/src/venv/lib/python3.4/site-packages/mistune.py", line 982, in parse
    out = self.output(preprocessing(text))
  File "/opt/jenkins/slave/workspace/ocfweb/PR-234/src/venv/lib/python3.4/site-packages/mistune.py", line 1031, in output
    out += self.tok()
  File "/opt/jenkins/slave/workspace/ocfweb/PR-234/src/venv/lib/python3.4/site-packages/mistune.py", line 1041, in tok
    return getattr(self, 'output_%s' % t)()
  File "/opt/jenkins/slave/workspace/ocfweb/PR-234/src/venv/lib/python3.4/site-packages/mistune.py", line 1102, in output_list
    while self.pop()['type'] != 'list_end':
TypeError: 'NoneType' object is not subscriptable
```

not sure what that means, i think i saw it before with invalid markdown or something

chriskuehl:
yeah it was broken on master, i fixed it: https://github.com/ocf/ocfweb/commit/ea3901e613b8029649483e467611d70e36d5f3e5

chriskuehl:
but it's still failing wtf

tahabi:
need to fix whitespace on all the scripts/*.md files as well, it seems. How did you figure out it was trailing whitespace?

chriskuehl:
it's not the whitespace actually, it's because the link is invalid

this is not valid syntax:

```markdown
[[web form|doc account/vhost/mail]]
```

i'm just going to revert that commit, the instructions are not correct anyway (we don't have a form for setting up mail vhosts)

PR #237, opened 2017-02-03
use variables in site.scss in other places

tahabi:
ignore the first and last commits in this branch


tahabi:
thanks, learned something new about git

PR #238, opened 2017-02-03
Hours Display for TV

tahabi:
Displays today's hours in large font on the TV. Works with days that have multiple blocks of hours.
screenshot - 02032017 - 05 52 38 am

Per @matthew-mcallister I structured this as a new app to Django as he suggested he had other ideas for ocfweb pages that could be created for the TV only, such as a lab map. 

Please advise if something I've added breaks a pattern or could be done better.

Improvements that can be made off the bat include centering the text and dynamically resizing it depending on how many hour-blocks there are, but that is currently beyond my limited CSS ability.


ocfweb/main/robots.py
```diff
@@ -16,6 +16,7 @@ def robots_dot_txt(request):
             User-Agent: *
             Disallow: /login/calnet/
             Disallow: /test/
+            Disallow: /tv/
```
matthew-mcallister: Nice.

ocfweb/tv/templates/tv/tv-base.html
```diff
@@ -0,0 +1,30 @@
...
+    <link rel="icon" href="{% static 'img/favicon/favicon-32.png' %}" sizes="32x32" />
+  </head>
+
+  <body>
+    <div class="page-tv">
+      {% block tv-body %}
+      {% endblock %}
+    </div>
+  </body>
+</html>
```
matthew-mcallister: Good idea copying out some stuff from the base template.

ocfweb/tv/templatetags/tv_formatting.py
```diff
@@ -0,0 +1,14 @@
...
+
+@register.filter
+def tv_lab_hours(hour):
+    if hour:
+        if hour.open.minute != 0 or hour.close.minute != 0:
+            fmt = '{:%-I:%M%P}-{:%-I%M%P}'
+        else:
+            fmt = '{:%-I%P}-{:%-I%P}'
+
+        return fmt.format(hour.open, hour.close)
```
matthew-mcallister: https://github.com/ocf/ocfweb/blob/0ae729c357a2b1d5002b393fafbc134e8a9b198b/ocfweb/templatetags/lab_hours.py#L14
Similar to this function; maybe try to adapt it? Not really necessary, your call.

ocfweb/tv/templates/tv/tv-base.html
```diff
@@ -0,0 +1,30 @@
...
+      {% if title %}
+      {{ title }} &ndash;
+      {% endif %}
+      Open Computing Facility
+    </title>
+
+    <link href="https://fonts.googleapis.com/css?family=Roboto:400,700,400italic,700italic,500,500italic" rel="stylesheet" />
+    <link href="{% static 'scss/site.scss.css' %}" rel="stylesheet" />
+    <link rel="icon" href="{% static 'img/favicon/favicon-16.png' %}" sizes="16x16" />
+    <link rel="icon" href="{% static 'img/favicon/favicon-32.png' %}" sizes="32x32" />
```
chriskuehl: is this entire head tag copied? can you move this to a partial and include it in both places rather than duplicating?

ocfweb/tv/templates/tv/hours.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'tv/tv-base.html' %}
+{% load tv_formatting %}
+
+{% block tv-body %}
+<div class="page-tv-hours">
```
chriskuehl: indent inside the block tag?

ocfweb/tv/templatetags/tv_formatting.py
```diff
@@ -0,0 +1,14 @@
...
+register = template.Library()
+
+
+@register.filter
+def tv_lab_hours(hour):
+    if hour:
+        if hour.open.minute != 0 or hour.close.minute != 0:
+            fmt = '{:%-I:%M%P}-{:%-I%M%P}'
+        else:
+            fmt = '{:%-I%P}-{:%-I%P}'
```
chriskuehl: probably want en-dashes like the other filter, instead of the hyphens you're using now?

8am-9pm
vs
8am–9pm

ocfweb/tv/templates/tv/tv-base.html
```diff
@@ -0,0 +1,30 @@
+{% load staticfiles %}
```
chriskuehl: i'd just call this file `tv/base.html`, it's not ambiguous because of the `tv/` prefix

ocfweb/tv/templatetags/tv_formatting.py
tahabi: I modeled this after that, that one joins everything into one string but that doesn't display very cleanly on the TV, as in it didn't break nicely, so I rewrote it to just format the time, while the loop in the template automatically breaks multiple hours so that everything is displayed cleanly

ocfweb/tv/templates/tv/tv-base.html
tahabi: Yeah I was trying to fix the the app prefix error and just moved the template in, I'll rename this

ocfweb/tv/templates/tv/tv-base.html
tahabi: I cut out a little bit but not a significant amount. I'll move it to a partial.

ocfweb/tv/templatetags/tv_formatting.py
tahabi: Yeah, good idea

tahabi: So after testing this, hyphens don't wrap and cause the font size to scale according to the viewport width. en-dashes and m-dashes will split before/after, making the text wrap. This reduces visibility because it will show half the hours on another line, and if there are more than one group of hours, i.e. 12-4, 5-7 the second block will be pushed below the fold.

chriskuehl: sounds like you want `text-wrap: none`?

ocfweb/tv/templates/tv/hours.html
tahabi: ?

ocfweb/tv/templatetags/tv_formatting.py
tahabi: Yeah I figured white-space after looking it up but for some reason text-wrap seems to work for me in Chrome (even though it's not an implemented property?) while white-space: nowrap just truncates the text. I'm going to explore this a bit more.

tahabi: nvm it was only working because I wasn't testing it properly.

ocfweb/static/scss/pages/_tv.scss
```diff
@@ -12,6 +12,7 @@ $base-font-size: 8vw;
         .hours-container {
             font-size: $base-font-size + 12;
             line-height: 1em;
+            text-wrap: none;
```
chriskuehl: see my edit, this is actually the wrong css property

tahabi: yeah, it's not even an implemented one that's supported on any browser. For some reason when I tried it it "worked" but that was because I was manually editing things in the Chrome Inspector. white-space: nowrap didn't work either because it just truncates any overflow. My current solution is hacky but it works - it just gives it a smaller font size overall if the hours would wrap. Apparently you can't do automatic resizing based on amount of text in CSS.

PR #243, opened 2017-02-11
Upgrade to celery 4, use stunnel instead of spiped

chriskuehl:
~~still using spiped for redis locking~~ that doesn't actually happen in ocfweb anyway


chriskuehl:
Looks like pysnmp 4.3.2 is broken? I'll try downgrading again to 4.3.1

PR #249, opened 2017-02-20
Added github button

doku:


matthew-mcallister:
Spaces, no tabs.

doku:
Documentation about installing pre-commit don't work. And I forgot to run make test. It would have caught the formatting issue. I'll try again.

ocfweb/templates/base.html
```diff
@@ -207,7 +209,12 @@
                         <p>The Open Computing Facility is run entirely by student volunteers.</p>
                         <p>Copyright &copy; 1989&ndash;{% now 'Y' %} Board of Directors of the Open Computing Facility.</p>
                         <p>The Open Computing Facility is a Chartered Program of the ASUC.</p>
-                    </div>
+                        <p>All our software including this website is open source.</p>
+                        <p>
+                            <a class="github-button" href="https://github.com/ntkme/github-buttons/fork" data-icon="octicon-repo-forked" aria-label="Fork ntkme/github-buttons on GitHub">Fork</a>
```
jvperrin: This link and the label text should be changed to be OCF-related.

ocfweb/templates/base.html
```diff
@@ -225,7 +232,7 @@
         <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
         <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/js/bootstrap.min.js"></script>
         <script src="{% static 'js/site.js' %}"></script>
-
+    <script async defer src="https://buttons.github.io/buttons.js"></script>
```
jvperrin: Indenting this like the other script would be good

ocfweb/templates/base.html
```diff
@@ -207,7 +209,12 @@
                         <p>The Open Computing Facility is run entirely by student volunteers.</p>
                         <p>Copyright &copy; 1989&ndash;{% now 'Y' %} Board of Directors of the Open Computing Facility.</p>
                         <p>The Open Computing Facility is a Chartered Program of the ASUC.</p>
-                    </div>
+                        <p>All our software including this website is open source.</p>
```
jvperrin: Nice, this is great to include.

jvperrin: Nice, this is great to include.

tahabi: small nitpick but I think grammatically it should be "All of our software, including this website, is open source"

chriskuehl:
Is there a way we can avoid including any more JS on our pages? Especially on all pages. I'm not crazy about adding more JS if we can avoid it, and these buttons are basically just image links.

I'm not a huge fan of their buttons either, and they don't match very well with our site -- what about just adding a text link to GitHub?

matthew-mcallister:
I see what you mean about JS, doubly so when it comes from a third-party
source.

As far as the visuals of the button, while I agree it sort of clashes, I
think there's some value in the iconography of the GitHub button. It's
highly recognizable and will attract the attention of those who might be
interested in the code.

kkuehler:
Been over a year so I'm gonna close this.

PR #250, opened 2017-02-21
Stop using my pre-commit fork

chriskuehl:
We don't need it for ocfweb (utils and puppet are the only repos that
need it).


chriskuehl:
Note that you might get errors like `pre-commit` not being installed in your virtualenv once this branch ships. I'm afraid I don't have a great solution besides `venv/bin/pip uninstall pre-commit && touch requirements.txt && make venv`.

Unfortunately this is a problem with running forks. `pip` can't understand that it needs to uninstall `ckuehl-pre-commit` before installing `pre-commit` (since they provide the same files), so the uninstall deletes the files provided by the mainline `pre-commit`.

PR #251, opened 2017-02-21
Add reverse trigger for dockers

chriskuehl:


jvperrin:
I'm thinking of making everything regular triggers, so like dockers would trigger a bunch of projects, but I do like that reverse build triggers are contained in the repo it builds, that seems nicer. Regular triggers are easier to add, and are a single line, which is pretty nice, so not sure which to go with...

chriskuehl:
I guess there's two conflicting views here. On the one hand, Jenkins apparently makes it slightly easier to configure non-reverse triggers. But on the other hand, a library (or in this case, `dockers`), shouldn't really have to care about things that depend on it -- it should be up to things depending on the library to declare that dependency).

Personally, I prefer the reverse trigger model. I'd rather that libraries (or upstream projects) don't have to care too much about their downstreams. But of course that isn't perfect either -- we still think about the projects using ocflib when we make changes to ocflib, for example.

In the real world, though, given that we don't actually pin ocflib (and other dependencies), it's not a big deal. I'm okay with either.

jvperrin:
Yup, that pretty much summarizes my views on it too, I just don't like the look of the reverse trigger model so much, but the rest makes a lot more sense, I do very much agree it would be best to have the build trigger in the repo that it actually builds, that makes the most sense to me.

PR #252, opened 2017-02-21
Make "make test" and "make cook-image" run in parallel

chriskuehl:
lol let's see how many tries it takes to get this right

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,35 +28,41 @@ node('slave') {
         }
     }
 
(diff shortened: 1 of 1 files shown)


chriskuehl:
3 tries, not bad. One downside is the UI doesn't do a great job of showing this (the console output does, though), but I think it's still worth it given that we can speed up testing and deploys quite a bit? What do others think?

(I'll squash before merging if people are OK with this.)

PR #254, opened 2017-02-26
README updates

tahabi:
note about bootstrap import, adding dev_config to Hiera.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,11 @@ To run tests locally, run `make test`. Please don't push to master with
 failing tests—Jenkins will refuse to deploy your code, and nobody will be able
 to deploy until fixing it.
 
+If you make a pull request to the OCF GitHub organization from your fork of
+ocfweb, Jenkins will attempt to build your branch, and if your build fails,
+you can log into [Jenkins](https://jenkins.ocf.berkeley.edu) with your OCF
+account credentials to see which tests you've failed and fix them. Almost
+invariably it's something `pre-commit` would have caught.
 
 ### Running pre-commit
 
@@ -61,5 +66,5 @@ Specifically, add:
 
     classes:
         - ocf_ocfweb::dev_config
-        
+
 It's probably easier to just run everything on `supernova`.
--- a/README.md
+++ b/README.md
@@ -40,10 +40,9 @@ failing tests—Jenkins will refuse to deploy your code, and nobody will be able
 to deploy until fixing it.
 
 If you make a pull request to the OCF GitHub organization from your fork of
-ocfweb, Jenkins will attempt to build your branch, and if your build fails,
-you can log into [Jenkins](https://jenkins.ocf.berkeley.edu) with your OCF
-account credentials to see which tests you've failed and fix them. Almost
-invariably it's something `pre-commit` would have caught.
+ocfweb, Jenkins will attempt to build and test your branch automatically.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,6 +39,11 @@ To run tests locally, run `make test`. Please don't push to master with
 failing tests—Jenkins will refuse to deploy your code, and nobody will be able
 to deploy until fixing it.
 
+If you make a pull request to the OCF GitHub organization from your fork of
+ocfweb, Jenkins will attempt to build your branch, and if your build fails,
+you can log into [Jenkins](https://jenkins.ocf.berkeley.edu) with your OCF
```
jvperrin: I don't think logging in here helps at all, the same results are shown, so just going to Jenkins should be sufficient.

README.md
```diff
@@ -6,20 +6,21 @@ ocfweb
...
 
+If you run into problems with scss where it can't import bootstrap, it's
+because you forgot to run the second command.
 
 ### Running in development mode
 
-Either on supernova, or on your own staff VM, run `make dev`. The first time
-will take a while, but future runs will be almost instant thanks to
-[pip-faster](https://github.com/Yelp/pip-faster).
+Either on supernova, or on your own staff VM<sup>[0]</sup>, run `make dev`.
```
jvperrin: I think putting a short reference to this inline instead of having a footnote would be better, just to keep related content closer together.

README.md
```diff
@@ -38,6 +39,11 @@ To run tests locally, run `make test`. Please don't push to master with
 failing tests—Jenkins will refuse to deploy your code, and nobody will be able
 to deploy until fixing it.
 
+If you make a pull request to the OCF GitHub organization from your fork of
+ocfweb, Jenkins will attempt to build your branch, and if your build fails,
+you can log into [Jenkins](https://jenkins.ocf.berkeley.edu) with your OCF
+account credentials to see which tests you've failed and fix them. Almost
+invariably it's something `pre-commit` would have caught.
```
jvperrin: There are quite a few tests on this repository, so I'd just remove this sentence altogether. I think it assumes the changes being made are small, which isn't necessarily the case, and there is a more detailed section directly below about `pre-commit` and the git hooks it installs.

README.md
tahabi: I thought about that but putting the actual code in-line doesn't look too great

README.md
tahabi: oh yeah lol I started always logging in by default to re-trigger jobs forgot you don't need to log in to see build output

README.md
tahabi: noted

README.md
jvperrin: Yeah, I wondered about that, maybe it could be in a block under the paragraph or something? I guess I'm just not a fan of footnotes, but if they seem like the best way then that's fine too.

README.md
jvperrin: Haha, yeah, I log in all the time now too, but I never used to. I actually didn't know I could log in to Jenkins for a long time, so figuring that out was nice.

tahabi:
ironic that my first commit in this PR failed because of something pre-commit would have caught

README.md
tahabi: I looked at GitHub's preview and thought it looked marginally better as a footnote, but I also lack any design sense so I'll defer to you if you want to move it up.

PR #255, opened 2017-02-26
Job frequency chart

shex1627:
I made the chart that shows the number of jobs printed based on the page number using the new public data
They look like this.
weekend
weekday


ocfweb/stats/job_frequency.py
```diff
@@ -0,0 +1,106 @@
...
+
+def pyday_to_sqlday(pyday):
+    """Converting weekday index from python to mysql."""
+    return (pyday + 1) % 7 + 1
+
+
+def get_jobs(sql='SELECT * FROM `public_jobs`'):
+    """Get the publicjobs database to python"""
+    conn = quota.get_connection(db='ocfprinting')
+    with conn.cursor() as curs:
```
chriskuehl: you can simply do
```python
with quota.get_connection() as cursor:
```

ocfweb/stats/job_frequency.py
```diff
@@ -0,0 +1,106 @@
...
+        curs.execute(sql)
+        return curs.fetchall()
+
+
+@periodic(60)
```
chriskuehl: 60 seconds seems unnecessarily frequent, maybe every 30 minutes or something?

ocfweb/stats/job_frequency.py
```diff
@@ -0,0 +1,106 @@
...
+    sql_today_freq = '''
+    SELECT `pages`,  SUM(`count`) AS `count`
+    FROM `public_jobs`
+    WHERE
+        (`pages` <= {0}) AND
+        (DAYOFWEEK(`day`) = {1}) AND
+        (`day` = "{2}" )
+    GROUP BY `pages`
+    ORDER BY `pages` ASC
+    '''.format(day_quota, day_of_week, day)
```
chriskuehl: you should never format into a query string unless you can avoid it, this leads to sql injection

instead use parameterized queries which pass the data along separately, e.g.

```python
cursor.execute(
    'SELECT foo FROM table WHERE column = %s AND column2 = %s',
    (3, thing),
)
```

(in python, use `%s` even for ints)

even if the code is correct now, it's easy for it to become wrong later when somebody makes a change. parameterizing your sql statements is a good insurance policy.

ocfweb/stats/templates/stats/summary.html
```diff
@@ -11,6 +11,7 @@
             <img class="img-responsive" src="{% url 'daily_graph_image' %}" />
             <img class="img-responsive" src="{% url 'session_count_image' %}" />
             <img class="img-responsive" src="{% url 'session_length_image' %}" />
+        <img class="img-responsive" src="{% url 'daily_job_image' %}" />
```
chriskuehl: mismatched indentation

ocfweb/stats/job_frequency.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # executing the sql query to get the data
+    today_freq_data = get_jobs(sql_today_freq)
+
+    # converting the data into np array using defaultdict
+    today_jobs_ddict = defaultdict(int)
+    today_jobs_tuple = [(row['pages'], int(row['count']))for row in today_freq_data]
+    for pages, count in today_jobs_tuple:
+        today_jobs_ddict[pages] = count
```
chriskuehl: you can avoid the extra list if you want

```python
today_jobs_dict = {row['pages']: row['count'] for row in today_freq_data}
today_jobs_count = [today_jobs_dict.get(i, 0) for i in range(1, day_quota + 1)]

File: ocfweb/stats/job_frequency.py [hunk 6]
```diff
@@ -0,0 +1,106 @@
...
+        return _jobs_graph_image()
+    else:
+        return _jobs_graph_image(day=day)
+
+
+def get_jobs_plot(day):
+    """Return matplotlib plot showing the number i-page-job to the day."""
+
+    day_of_week = pyday_to_sqlday(day.weekday())
+    day_quota = quota.daily_quota(datetime.combine(day, datetime.min.time()))
```
matthew-mcallister: You can just do `day_quota = quota.daily_quota(day)`.

shex1627: the parameter day is actually a "date" object, but quota.daily_quota takes in "datetime" object

ocfweb/stats/urls.py
```diff
@@ -27,4 +27,5 @@
     ),
     url(r'^printing/semester-histogram/graph$', semester_histogram, name='semester_histogram'),
     url(r'^printing/pages-printed$', pages_printed, name='pages_printed'),
+    url(r'^daily-job/graph$', daily_jobs_image, name='daily_job_image'),
```
shex1627: Never thought of that but it does sound more reasonable.
P.S: I am not sure how the whole url works so I just "copy"  the code from "daily_graph_image"

ocfweb/stats/job_frequency.py
matthew-mcallister: Are you sure about that? You should try it.

ocfweb/stats/urls.py
matthew-mcallister: Same here. :)

ocfweb/stats/job_frequency.py
shex1627: That is what I did at the beginning. And it took me a while to figure the bug out.

matthew-mcallister: Oh. That was a bug in ocflib which, by sheer coincidence, I fixed minutes before making this comment.

ocfweb/stats/job_frequency.py
shex1627: Even though I think jobs are issued much much more frequent, we don't really have to check it that often(or at all). I guess 30 minutes is alright.

ocfweb/stats/job_frequency.py
shex1627: what is the difference between these two way of making strings

chriskuehl: http://stackoverflow.com/a/8265319/612279

shex1627: All the comments are "fixed" now, check if it is good to move forward.

ocfweb/stats/urls.py
jvperrin: Also same lol

ocfweb/stats/templates/stats/summary.html
```diff
@@ -11,6 +11,7 @@
             <img class="img-responsive" src="{% url 'daily_graph_image' %}" />
             <img class="img-responsive" src="{% url 'session_count_image' %}" />
             <img class="img-responsive" src="{% url 'session_length_image' %}" />
+            <img class="img-responsive" src="{% url 'daily_job_image' %}" />
```
jvperrin: I think this should go under the printing tab (at [ocfweb/stats/templates/stats/printing.html](https://github.com/ocf/ocfweb/blob/master/ocfweb/stats/templates/stats/printing.html)) instead of the main stats page, since it has more to do with printing.

chriskuehl: lgtm also except this comment

shex1627: weblink adjusted please check. The website is hosting on supernova:8977

jvperrin: That's a different thing (but that's a better url anyway), I was talking about putting this image in the page at [/stats/printing/](http://supernova.ocf.berkeley.edu:8977/stats/printing/), not at the current page ([/stats/](http://supernova.ocf.berkeley.edu:8977/stats/)).

shex1627: you mean like this(haven't commit/push yet), 
supernova:8977/stats/printing/

jvperrin: Yup, that looks good!

jvperrin:
Looks good, squashing commits would be nice, but otherwise, this is good to merge! (unless anyone else has comments)

shex1627:
what does "squashing commits" mean?

jvperrin:
It just makes the git history better (no merges, small fixes, those kinds of things). It's a common thing to do for larger repositories, just because the history gets too messy and long otherwise. There's a pretty good guide on it here: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

shex1627:
Alright we should talk about this tmr night.(I don't want to mess with this myself for now)

jvperrin:
Alright, sounds fine, thanks for doing this, it looks nice!

ocfweb/stats/job_frequency.py
```diff
@@ -0,0 +1,93 @@
...
+    ax = fig.add_subplot(1, 1, 1)
+
+    tickLocations = np.arange(1, day_quota + 1)
+    width = 0.8
+    ax.bar(tickLocations, today_jobs_count, width, color='blue')
+
+    ax.set_xticks(ticks=tickLocations)
+    ax.yaxis.grid(True)
+    ax.yaxis.set_major_locator(MaxNLocator(integer=True))
+    ax.grid(True)
```
NickImpicciche: Do we want the grid lines in y-axis? IMO it doesn't look that great. I would do ax.xaxis.grid(True)

NickImpicciche: Or maybe just have the ticks on the side instead of going through the bars

shex1627: How does the first one look?
yaxisgrid


aaxisgrid

matthew-mcallister:
Looks great, merge at your discretion.

jvperrin:
Nice, sorry for not helping with git rebasing today, but looks like you got it!

PR #256, opened 2017-02-27
add lint option to makefile

tahabi:
seems useful

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,8 +52,8 @@ install-hooks: venv
 	$(BIN)/pre-commit install -f --install-hooks
 
 .PHONY: lint
-lint:
-	pre-commit run --all-files
+lint: venv
+	$(BIN)/pre-commit run --all-files
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ venv: requirements.txt requirements-dev.txt
 install-hooks: venv
 	$(BIN)/pre-commit install -f --install-hooks
 
+.PHONY: lint
+lint:
+	pre-commit run --all-files
+
 .PHONY: clean
 clean:
 	rm -rf *.egg-info venv


Makefile
```diff
@@ -51,6 +51,10 @@ venv: requirements.txt requirements-dev.txt
 install-hooks: venv
 	$(BIN)/pre-commit install -f --install-hooks
 
+.PHONY: lint
+lint:
+	pre-commit run --all-files
```
jvperrin: I think you want `$(BIN)/pre-commit` here, since it isn't necessarily installed system-wide but should be installed in the repo's `venv`.

Makefile
```diff
@@ -51,6 +51,10 @@ venv: requirements.txt requirements-dev.txt
 install-hooks: venv
 	$(BIN)/pre-commit install -f --install-hooks
 
+.PHONY: lint
+lint:
```
jvperrin: Putting `venv` as a dependency of this task would be good (like in `install-hooks` or in `test`), because then that makes sure `pre-commit` is installed so that it can be run.

tahabi: yeah I had did that at first but it re-installs the hooks every time which I thought was weird even though it doesn't make any significant difference.

Makefile
tahabi: nice catch didn't remember that

Makefile
jvperrin: Hmm, that's strange, it should be a dependency, but it shouldn't have to reinstall every time, it shouldn't even have to do anything if the venv is updated.

tahabi: yeah when I first tried, it reinstalled the whole venv, then re-installed the hooks. Second time and everytime thereafter it only re-installed the hooks on every invocation so I decided I'd leave the dependency out because I'm not really too familiar with the mechanics of how this works.

jvperrin: It works for me to just have it as a dependency without reinstalling, so I'm not sure what's happening for you on this.

tahabi: weird, I still get it calling the install command.
screen shot 2017-02-27 at 3 04 49 am
screen shot 2017-02-27 at 3 05 52 am

jvperrin: Wait, I mean to put `lint: venv`, not `lint: install-hooks` (I was using `install-hooks` as an example of another task that depends on `venv`).

Does it work with `venv` used instead?

tahabi: yep that works much better

Issue #56, opened 2015-10-05
Docs: explain off-site vs. OCF hosting


matthew-mcallister:
A common question, brought up in [rt#3973](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=3973). Should probably distinguish free vs. proprietary software, off-site vs. on-site, web software versus web hosting service (and how the latter two are often intentionally muddled up). Also give a somewhat more complete list of available software.

hexecute:
> free vs. proprietary software  

Do we really need to do that here? I think that might just make people angry without any benefits (e.g. "We can't use X because of OCF's political views?"). Instead, we could approach it as a requirement of the university for security purposes which is more accurate (I think?).

kpengboy:
Or cut to the chase and explain that it is not physically possible for us
to host Wix etc. sites?

matthew-mcallister:
I definitely don't recommend posting an entire Stallman article. It should be clear why we cannot host a Wix website. Example:

> Many "website builder" style services, such as Wix, Weebly, use a proprietary software backend owned by their companies. The OCF doesn't have the right or ability to transfer any website made with those services to our servers, so we cannot host them on-site. On the other hand, we can run anything using free and open-source software, like WordPress and Rails, so make your website with those!

nattofriends:
No. This is not about free software. Make it abundantly clear that what Wix
asks people to do is not OCF hosting a site, it is OCF pointing signs to
Wix hosting the site themselves. It is due to University rules that we
cannot do this.

Then proceed to describe the possible solutions, such as submitting an
offsite request, or using free software hosted on the OCF.

matthew-mcallister:
I agree, which is why I didn't say anything about the merits/demerits of free/proprietary software. However, a user might rightly wonder "Why isn't there a way to copy my Wix website over to the OCF servers?" After all, you can move from WordPress.com to WordPress on the OCF.

It sounds like you're referring to some specific user support page on Wix which tells people to ask for DNS to point to Wix. Yes, we should explicitly warn people that we cannot do that. If it needs to say more explicitly "No, we cannot give you offsite DNS *nor* can we copy your Wix website to our servers", then so be it. But it deserves at least a one-sentence explanation.

nattofriends:
Sure, that makes sense. I will add that Wordpress and Tumblr also have
CNAME handling support, so it isn't limited to just Wix.

2015-10-05 20:19 GMT-07:00 Matthew McAllister <notifications@github.com>:

PR #257, opened 2017-02-27
draft for fixing #56

abizer:
draft, obviously. Any thoughts?

### DIFF
--- a/ocfweb/docs/docs/services/web.md
+++ b/ocfweb/docs/docs/services/web.md
@@ -13,27 +13,46 @@ another domain (e.g., group.berkeley.edu).
 
 #### Off-site hosting
 
-One question that often comes up is if the OCF is able to  websites from
-services such as Squarespace, Wix, Weebly, Tumblr, etc. Unfortunately
-due to logistical considerations and University policy, the OCF is
-unable to do so.
+One question that often comes up is if the OCF is able to host websites
+from services such as Squarespace, Wix, Weebly, Tumblr, etc. Unfortunately
+due to logistical considerations and University policy, the OCF is unable
+to do so.
 
-First, University policy prohibits campus DNS from redirection to off-site
-hosts without explicit permission from the University. This means the OCF
-is not the appropriate point of contact in cases like this where providers
-allow you to use custom domains, or other such things - one would have to
+First, many website building services like Wix and Weebly make their money
+by using a proprietary service to allow users to build their sites, and
+subsequently charging users to host those sites. This means providing the
+servers, storage space, internet connection, and configurations to allow
+these sites to show up on the internet. Obviously, the OCF is unable to
+access these proprietary backends and is therefore unable to host sites
+developed using them. However, the OCF is capable of hosting anything made
+using the free and open-source backends we support, including Wordpress,
+Django, Ruby on Rails, Node.js, PHP, and others. You may wish to consider
+building your website with those if you desire to host with the OCF.
+
+Second, University policy prohibits campus DNS from redirection to off-site
+hosts without explicit permission from the University. This means that, if
+you want to host your website with Wix or some other provider, we cannot
+give you a URL like group.berkeley.edu that points to that website. At the
+very least, the OCF is not the appropriate point of contact in such cases -
+one would have to
 [contact the university](https://offsitehosting.berkeley.edu/) directly
 to request permission. This is primarily relevant to student groups.
+Similarly, the OCF does not generally allow hosting under custom domain
+names. This means we do not support things like "you_or_your_group.com"
+pointing to a website whose files are hosted on OCF servers except under
+special circumstances, usually with authorization from the University. If
+you have your own domain name and would like to host a website under it,
+we recommend you get an account with a shared hosting provider. Unfortunately
+it is unlikely that this will be free.
 
-Second, many website building services like Wix and Weebly make their money
-by using a proprietary product to allow users to build their sites, and
-subsequently charging users to host those sites. Obviously, the OCF is
-unable to access these proprietary backends and is therefore unable to host
-sites developed using them. However, the OCF is capable of hosting anything
-developed using the free and open-source software we support, including
-Wordpress, Django, Ruby on Rails, Node.js, PHP, and others. You may wish
-to consider building your website with those if you desire to host with the
-OCF.
+For reference, a typical web hosting scenario involves two things: the domain
+name, and the hosting server. The domain name is purchased from a registrar,
+and using it involves setting Domain Name System (DNS) records to point to
+the address of a hosting server. This server is purchased from any variety
+of sources, and must be configured to respond to web requests to the domain
+name in question. While the OCF provides both, only a limited subset of their
+functionality are available for general use by the campus community due to the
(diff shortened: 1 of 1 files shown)


tahabi:
good point

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,49 @@ where `user` is the account name.
...
+Second, University policy prohibits campus DNS from redirection to off-site
+hosts without explicit permission from the University. This means that, if
+you want to host your website with Wix or some other provider, we cannot
+give you a URL like group.berkeley.edu that points to that website. At the
+very least, the OCF is not the appropriate point of contact in such cases -
+one would have to
+[contact the university](https://offsitehosting.berkeley.edu/) directly
+to request permission. This is primarily relevant to student groups.
+Similarly, the OCF does not generally allow hosting under custom domain
+names. This means we do not support things like "you_or_your_group.com"
```
jvperrin: This is not entirely true, at least we host a few of these non-berkeley.edu domains and we have [a paragraph](https://www.ocf.berkeley.edu/docs/services/vhost/#h4_limitations-on-non-berkeleyedu) about hosting without a berkeley.edu subdomain, although given that regular subdomains require an application anyway, I think that could be revised to essentially say only the second point that says we don't pay for the domain, the group wanting to host with us does.

I think we need a clearer policy on what to do about these, and I don't see a problem with the OCF hosting them as long as they follow our requirements.

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,49 @@ where `user` is the account name.
 Group accounts may also request a [[virtual host|doc services/vhost]] for
 another domain (e.g., group.berkeley.edu).
 
+#### Off-site hosting
+
+One question that often comes up is if the OCF is able to host websites
+from services such as Squarespace, Wix, Weebly, Tumblr, etc. Unfortunately
```
jvperrin: I think it should be made clear here that we _can_ host websites _imported_ from these services, but that we don't have the same editing interfaces and stuff that they do. We also can't somehow host a website also hosted by them, that doesn't make any sense, which is more the point you made in the next paragraph.

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,49 @@ where `user` is the account name.
...
+due to logistical considerations and University policy, the OCF is unable
+to do so.
+
+First, many website building services like Wix and Weebly make their money
+by using a proprietary service to allow users to build their sites, and
+subsequently charging users to host those sites. This means providing the
+servers, storage space, internet connection, and configurations to allow
+these sites to show up on the internet. Obviously, the OCF is unable to
+access these proprietary backends and is therefore unable to host sites
```
jvperrin: Using the word `backend` will not make a lot of sense to most people wanting to host with us, maybe just use something like `access this proprietary software` instead?

tahabi:
relevant: [rt#6018](https://ocf.io/rt/6018)
I'll start working on this again

tahabi:
@jvperrin @kpengboy @matthew-mcallister any thoughts?

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,52 @@ where `user` is the account name.
...
+more information.
+
+For reference, a typical web hosting scenario involves two things: the domain
+name, and the hosting server. The domain name is purchased from a registrar,
+and using it involves setting Domain Name System (DNS) records to point to
+the address of a hosting server. This server is purchased from any variety
+of sources, and must be configured to respond to web requests to the domain
+name in question. While the OCF provides both, only a limited subset of their
+functionality are available for general use by the campus community due to the
+legal, technical, and policy considerations explained above.
```
jvperrin: I feel like this last part doesn't work so well, like the part about the server being "purchased from any variety of sources" and then explaining the limited functionality of our servers. To me, it seems like something we shouldn't be explaining (in vague terms anyway) and something that just complicates the topic.

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,52 @@ where `user` is the account name.
...
+[contact the university](https://offsitehosting.berkeley.edu/) directly
+to request permission. This is primarily relevant to student groups.
+
+Similarly, the OCF does not generally allow hosting under custom domain names
+for student groups. This means we do not usually support things like
+"you_or_your_group.com" pointing to a website whose files are hosted on OCF
+servers. However, we may agree to host websites under custom domains on a
+case-by-case basis, usually for University affiliates such as faculty members,
+ASUC/University programs, and others. Please [[contact us|doc contact]] for
+more information.
```
jvperrin: We'd want to link to the vhost policy we are writing for a lot of this part, so I think it might be good if this waits until that is written up, which should be in the next week.

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,52 @@ where `user` is the account name.
...
+Many website building services like Wix and Weebly make their money by
+using a proprietary service to allow users to design their websites, and
+subsequently charge users to host those sites. This means providing the
+servers, storage space, internet connection, and configurations to allow
+these sites to show up on the internet. Obviously, the OCF is unable to
+access this proprietary software and is therefore unable to host sites
+developed using them. However, the OCF is capable of hosting anything made
+using the free and open-source backends we support, including Wordpress,
+Django, Ruby on Rails, Node.js, PHP, and others. You may wish to consider
+building your website with those if you desire to host with the OCF.
```
jvperrin: I like this section, maybe mention the ability to export from those providers though in case they are already using them? I would make sure to discourage exporting from their service and importing to the OCF though because that complicates setting up their site and probably wouldn't be as easy to maintain due to whatever obfuscation/plugins they use or other stuff like that (unless it's simpler, like migrating from Wordpress.com to Wordpress at the OCF, which we already have a docs section for).

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,52 @@ where `user` is the account name.
...
+due to logistical considerations and University policy, the OCF is unable
+to do so. More specifically, the OCF cannot point  <group>.berkeley.edu
+domain names to off-campus hosting providers, but it can, however, host
+websites exported from such providers.
+
+Many website building services like Wix and Weebly make their money by
+using a proprietary service to allow users to design their websites, and
+subsequently charge users to host those sites. This means providing the
+servers, storage space, internet connection, and configurations to allow
+these sites to show up on the internet. Obviously, the OCF is unable to
```
jvperrin: Not so obviously for some ;)

ocfweb/docs/docs/services/web.md
```diff
@@ -11,6 +11,52 @@ where `user` is the account name.
 Group accounts may also request a [[virtual host|doc services/vhost]] for
 another domain (e.g., group.berkeley.edu).
 
+#### Off-site hosting
```
chriskuehl: Can we put most of this information somewhere else? A small section here that links to another page might work well. Right now this takes a page which contains useful technical information about updating your website and inserts a giant blob of political text which isn't relevant to most people.

PR #258, opened 2017-02-27
not too important but fixes #137

### LINKED ISSUE #137: Add open-graph tags to website (at least home page)
We should add opengraph image tags to prevent this. Probably just use the penguin (we might need a rasterized version).

this is actually a common issue...

tahabi:


ocfweb/templates/partials/head.html
```diff
@@ -12,6 +12,8 @@
         {% endif %}
         {% if og_image %}
             <meta property="og:image" content="{{og_image}}" />
+        {% else %}
+        <meta property="og:image" content="{% static 'img/penguin.png' %}"/>
```
chriskuehl: indentation

PR #259, opened 2017-02-28
Sign.json dict

asaiacai:
JSON dict for lab hours for LED signboard


tahabi:
You're going to have to register `sign` as an app with Django if you want to do it like this, see how I did it in #238 for reference. Add it to ocfweb/settings.py, among other things.

matthew-mcallister:
Do we want to make this a page in ocfweb or should we instead put a cgi script in e.g. the staff directory?

ocfweb/urls.py
```diff
@@ -28,6 +29,7 @@
 
     url(r'^$', home, name='home'),
     url(r'^robots\.txt$', robots_dot_txt, name='robots.txt'),
+    url(r'^sign\.json$', sign_text, name='sign.json'),
```
jvperrin: I think this should be under a `/api` namespace or something, just to make it clear it isn't for general use, and to make a place where other API-only URLs can be put.

Same with the location of the file above (in `ocfweb/api/sign/sign_text.py` or something)

chriskuehl: +1, /api/ sounds great. i was going to comment that we should move it out of the root, but couldn't think of a good name. /api/ is great.

matthew-mcallister:
Surely we can version control a CGI script? Nonetheless, it's fine in here as well.

chriskuehl:
Why do you want it as a CGI script?

matthew-mcallister:
I don't particularly. I just thought that this script was pretty much unrelated to the main website and didn't depend on its code, so it would be easy to factor out.

If we're going to go the route of adding an `/api/` tree, though, that seems like a perfect place to put it. I would like discuss using `/api/` as the canonical URL for `staff_hours.yaml`, etc., in another thread.

matthew-mcallister:
Protips:
1. Don't merge 'master' into your branch. Updating your branch is precisely the purpose of `git rebase`.
2. Don't push unrelated changes to your pull requests; instead, make as many simultaneous new PRs as you'd like.

ocfweb/api/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.conf.urls import url
+
+from ocfweb.api import api
+
+urlpatterns = [
+    url(r'sign\.json', api.sign_text, name='sign-json'),
```
chriskuehl: i'd probably remove the `.json`, that's pretty non-standard for APIs and we wouldn't include it if we e.g. expanded the API to include vhost mail things

chriskuehl: what about calling it `/api/hours`? there's nothing in here really specific to the sign, and we might find other uses for it

jvperrin:
(because approving a PR 2 times isn't enough, it has to be done a third time for good measure)

jvperrin:
Nice, that's a lot of check marks

chriskuehl:
@asaiacai I sent you an invite to the ocf GitHub org, you'll probably need to accept that before you can merge.

matthew-mcallister:
You may need to visit https://ocf.io/gh to accept the invitation.

PR #262, opened 2017-03-01
changed flask.md to use virtualenv

asaiacai:


tahabi:
ocf.io/rt/4149

ocfweb/docs/docs/services/web/flask.md
```diff
@@ -23,13 +23,21 @@ OCF servers requires only just a little extra configuration.
...
       from flup.server.fcgi import WSGIServer
       from flaskyapplication import app
 
       if __name__ == '__main__':
           WSGIServer(app).run()
 
+1.     Make a virtualenv.  This will allow you to *pip install* any python packages
+that you need that the OCF does not already have:
+```
+    $ virtualenv venv
```
chriskuehl: i'd suggest `virtualenv -p python3.5 venv`

hopefully most people are starting new things on python 3 (seems likely since that's what 61A teaches), and if not, this at least makes it obvious how to choose a different python version

ocfweb/docs/docs/services/web/flask.md
```diff
@@ -23,13 +23,21 @@ OCF servers requires only just a little extra configuration.
 
 * run.fcgi:
 
-      #!/usr/bin/env python
+      #!venv/bin/python
```
chriskuehl: this isn't a pattern i see very often, but it's fine if it works. did you confirm it does with fastcgi?

ocfweb/docs/docs/services/web/flask.md
```diff
@@ -23,13 +23,21 @@ OCF servers requires only just a little extra configuration.
...
 
       if __name__ == '__main__':
           WSGIServer(app).run()
 
+1.     Make a virtualenv.  This will allow you to *pip install* any python packages
+that you need that the OCF does not already have:
+```
+    $ virtualenv venv
+    $ . venv/bin/activate
+    $ pip install <YOUR PACKAGE>
```
chriskuehl: maybe also include flask here, just to make it obvious they will have to install that

ocfweb/docs/docs/services/web/flask.md
```diff
@@ -23,13 +23,21 @@ OCF servers requires only just a little extra configuration.
...
-      #!/usr/bin/env python
+      #!venv/bin/python
       from flup.server.fcgi import WSGIServer
       from flaskyapplication import app
 
       if __name__ == '__main__':
           WSGIServer(app).run()
 
+1.     Make a virtualenv.  This will allow you to *pip install* any python packages
+that you need that the OCF does not already have:
```
chriskuehl: maybe add a line like "The benefit is that you can choose the exact versions of dependencies you want, without being tied to the system versions that all OCF accounts use."

ocfweb/docs/docs/services/web/flask.md
asaiacai: Yeah.  This is what I used in https://www.ocf.berkeley.edu/~eau/flasky/

ocfweb/docs/docs/services/web/flask.md
asaiacai: if we use python3, we would have to do `pip install flup6` instead of `pip install flup`.  I can add that.

asaiacai:
I realize our current instructions for deploying with Django only works for Django 1.8 and not the latest Django 1.10
Do we want to change the whole django.md or just say to install Django 1.8 for now?

Baisang:
should encourage people to use 1.10 as 1.8 is old af (pretty sure upe website is 1.8)

matthew-mcallister:
@asaiacai Uh, you included some bogus commits there fella.

PR #267, opened 2017-03-04
Semester job

shex1627:
Adding semester-wise weekday/weekend job distribution graphs. You can also adjust adjust the start and end date of the url request to adjust to time interval that is being queried. 

Suggestions on graph locations/titles/labels and Code Reviews?
(also it is possible to use the code from semester_job.py to create daily job graph by making some changes not sure if you guys want it or not)
http://supernova:8977/stats/printing/

### DIFF
--- a/ocfweb/stats/semester_job.py
+++ b/ocfweb/stats/semester_job.py
@@ -1,0 +1,89 @@
+import ocflib.printing.quota as quota
+from django.http import HttpResponse
+from matplotlib.figure import Figure
+from matplotlib.ticker import MaxNLocator
+from ocflib.lab.stats import semester_dates
+from ocflib.printing.quota import WEEKDAY_QUOTA
+from ocflib.printing.quota import WEEKEND_QUOTA
+
+from ocfweb.component.graph import canonical_graph
+from ocfweb.component.graph import plot_to_image_bytes
+
+
+@canonical_graph(default_start_end=semester_dates)
+def weekday_jobs_image(request, start_day, end_day):
+    return HttpResponse(
+        plot_to_image_bytes(get_jobs_plot('weekday', start_day, end_day), format='svg'),
+        content_type='image/svg+xml',
+    )
+
+
+@canonical_graph(default_start_end=semester_dates)
+def weekend_jobs_image(request, start_day, end_day):
+    return HttpResponse(
+        plot_to_image_bytes(get_jobs_plot('weekend', start_day, end_day), format='svg'),
+        content_type='image/svg+xml',
+    )
+
+
+# dictionary that gives parameters/config of different job graphs
+graphs = {
+    'weekday':
+        {'query': '''
+                SELECT `pages`, SUM(`count`) AS `count`
+                FROM `public_jobs`
+                WHERE
+                    (`pages` <= %s) AND
+                    (DAYOFWEEK(`day`) MOD 7 > 1) AND
+                    (`day` BETWEEN CAST(%s AS Date) AND CAST(%s AS DATE))
+                GROUP BY `pages`''',
+         'quota': tuple([WEEKDAY_QUOTA]),
+         'title': 'Semester Weekday Jobs Distributiom'},
+    'weekend':
+        {'query': '''
+                SELECT `pages`, SUM(`count`) AS `count`
+                FROM `public_jobs`
+                WHERE
+                    (`pages` <= %s) AND
+                    (DAYOFWEEK(`day`) MOD 7 <= 1) AND
+                    (`day` BETWEEN CAST(%s AS Date) AND CAST(%s AS DATE))
+                GROUP BY `pages`''',
+         'quota': tuple([WEEKEND_QUOTA]),
+         'title': 'Semester Weekend Jobs Distribution'}
+}
+
+
+def freq_plot(data, title,
+              ylab='Number of Jobs Printed'):
+    """takes in data, title, and ylab and makes a histogram, with
+    the 1:len(data) as the xaxis
+    """
+    fig = Figure(figsize=(10, 4))
+    ax = fig.add_subplot(1, 1, 1)
+
+    tickLocations = range(1, len(data) + 1)
+    width = 0.8
+    ax.bar(tickLocations, data, width)
+
+    ax.set_xticks(ticks=tickLocations)
+    ax.yaxis.set_major_locator(MaxNLocator(integer=True))
+    ax.yaxis.grid(True)
+    ax.set_ylim(ymin=0)
+    ax.set_ylabel(ylab)
+    ax.set_title(title)
+
+    return fig
+
+
+def get_jobs_plot(graph, start_day, end_day):
+    """Return matplotlib plot of the number of jobs of different page-jobs"""
+    graph_config = graphs[graph]
+    with quota.get_connection() as cursor:
+        cursor.execute(graph_config['query'],
+                       graph_config['quota'] + (start_day, end_day))
+        data = cursor.fetchall()
+
+    jobs_dict = {row['pages']: row['count'] for row in data}
+    jobs_count = [jobs_dict.get(i, 0) for i in range(1, graph_config['quota'][0] + 1)]
+
+    return freq_plot(jobs_count, graph_config['title'])
--- a/ocfweb/stats/templates/stats/printing.html
+++ b/ocfweb/stats/templates/stats/printing.html
@@ -7,6 +7,8 @@
         {% stats_navbar %}
         <p>
             <img class="img-responsive" src="{% url 'daily_job_image' %}" />
+            <img class="img-responsive" src="{% url 'weekday_jobs_image' %}" />
+            <img class="img-responsive" src="{% url 'weekend_jobs_image' %}" />
             <img class="img-responsive" src="{% url 'semester_histogram' %}" />
         </p>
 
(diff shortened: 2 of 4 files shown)


ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+MONTH_INDEX = 1
+SEMESTER_CUTOFF_MONTH = 8
+
+year = datetime.today().year
+spring_start = date(year, 1, 1)
+fall_start = date(year, 8, 1)
+
+
+def get_start_end():
+    """reutnr a tuple of startdate and endate of the semester based on today"""
```
NickImpicciche: typo

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+from matplotlib.ticker import MaxNLocator
+from ocflib.printing.quota import WEEKDAY_QUOTA
+from ocflib.printing.quota import WEEKEND_QUOTA
+
+from ocfweb.component.graph import canonical_graph
+from ocfweb.component.graph import plot_to_image_bytes
+
+ONE_DAY = timedelta(days=1)
+MONTH_INDEX = 1
+SEMESTER_CUTOFF_MONTH = 8
```
chriskuehl: we have some logic for this elsewhere in ocflib, let's try not to duplicate how we determine what the current semester is even more than we already do

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+
+year = datetime.today().year
+spring_start = date(year, 1, 1)
+fall_start = date(year, 8, 1)
+
+
+def get_start_end():
+    """reutnr a tuple of startdate and endate of the semester based on today"""
+    day = datetime.today().timetuple()
+    if day[MONTH_INDEX] < SEMESTER_CUTOFF_MONTH:
```
chriskuehl: `MONTH_INDEX` makes no sense as a constant; i had no idea what it meant when i saw it at the top of the file

instead you could either just:
1. use `datetime.today().month`
2. use `datetime.today().timetuple().tm_mon`?

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+        return (spring_start, fall_start - ONE_DAY)
+    else:
+        return (fall_start, spring_start - ONE_DAY)
```
chriskuehl: can we make this a function in ocflib? sounds useful, something like

```python
def semester_dates(day=None):
    """Returns a tuple (start day, end day) for the semester.

    Defaults to today if none is provided.
    """
    if day is None:
        day = date.today()
    # implementation
```

shex1627: I find a function (current_semester_start) that gives me the start date for my current semester. I could use it in my semester job graph replacing (get_start_end). But you also mention that function like get_start_end may be useful for ocflib as well. So I am thinking to implement function (semester_dates) like you suggested in ocflib and use that function for my chart instead. How is that?  

I am assuming function (semester_dates) takes into a day(datetime object) and returns the start, end day tuple of that day's semester regardless whether that day is in the semester. Is that right?
If so, my implmentation is probably like 
```python
year = datetime.today().year
spring_start = date(year,1, 1)
fall_start = date(year, 8, 1)
if day.month < 8:
   return (spring_start, fall_start - ONE_DAY)
else:
   return (fall_start, spring_start - ONE_DAY)

Comment (jvperrin) [inline ocfweb/stats/semester_job.py]: This is already kind of a thing in https://github.com/ocf/ocflib/commit/d275dcf893015d21cddbb30577df8dda4ca69ab2, I'm not sure if that's exactly what you are looking for, since you need a range of dates, but it would be a good starting point instead of hardcoding the start dates. Maybe the start dates in the commit I mentioned above could be extracted out into `SPRING_SEMESTER_START` and `FALL_SEMESTER_START` constants so you can use those?

File: ocfweb/stats/semester_job.py [hunk 5]
```diff
@@ -0,0 +1,109 @@
...
+        content_type='image/svg+xml',
+    )
+
+
+@canonical_graph(default_start_end=get_start_end)
+def weekend_jobs_image(request, start_day, end_day):
+    return HttpResponse(
+        plot_to_image_bytes(get_jobs_plot('weekend', start_day, end_day), format='svg'),
+        content_type='image/svg+xml',
+    )
```
jvperrin: I think it would be good to have both of these in some shared function, since they are almost exactly the same except for the plot used, so it would be good to duplicate less here.

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+            SELECT `pages`, SUM(`count`) AS `count`
+            FROM `public_jobs`
+            WHERE
+                (`pages` <= %s) AND
+                (DAYOFWEEK(`day`) MOD 7 <= 1) AND
+                (`day` BETWEEN CAST(%s AS Date) AND CAST(%s AS DATE))
+            GROUP BY `pages`''',
+     'quota': tuple([WEEKEND_QUOTA]),
+     'title': 'Semester Weekend Jobs Distribution'}
+}
```
jvperrin: The indentation is confusing here, since it's a dictionary, but it isn't clear it's a dictionary because the key is on a separate line, so I think moving the key and indenting the query more would make this clearer.

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+from ocfweb.component.graph import canonical_graph
+from ocfweb.component.graph import plot_to_image_bytes
+
+ONE_DAY = timedelta(days=1)
+MONTH_INDEX = 1
+SEMESTER_CUTOFF_MONTH = 8
+
+year = datetime.today().year
+spring_start = date(year, 1, 1)
+fall_start = date(year, 8, 1)
```
jvperrin: These, if kept as constants, should be put into ocflib instead.

ocfweb/stats/semester_job.py
shex1627: I think making an additional function would be an overkill and may increase the trouble in understanding how the things work in the file.

ocfweb/stats/semester_job.py
shex1627: Could you show me a code block of how I should do it(I am having some trouble understanding the instruction here)

ocfweb/stats/semester_job.py
shex1627: The problem of putting start_date and end_date in constants is that we have to update the years. So I think the utility function that gives the interval is sufficient.

ocfweb/stats/semester_job.py
tahabi: add another layer of indentation in every line after `'weekday':` and `'weekend':` respectively, so it's clear that `'query':` is a key in the dictionaries that `'weekday':` and `'weekend':` point to.

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,109 @@
...
+    'weekday':
+    {'query': '''
+            SELECT `pages`, SUM(`count`) AS `count`
+            FROM `public_jobs`
+            WHERE
+                (`pages` <= %s) AND
+                (DAYOFWEEK(`day`) MOD 7 > 1) AND
+                (`day` BETWEEN CAST(%s AS Date) AND CAST(%s AS DATE))
+            GROUP BY `pages`''',
+     'quota': tuple([WEEKDAY_QUOTA]),
```
tahabi: why is this a tuple of a list of a single constant? It makes the logic in line 107 less clear.

shex1627:
Using the new ocflib function to edit the SemesterJob graphs, you guys can see the graphs here
http://supernova:8977/stats/printing/

ocfweb/stats/templates/stats/printing.html
```diff
@@ -7,6 +7,8 @@
         {% stats_navbar %}
         <p>
             <img class="img-responsive" src="{% url 'daily_job_image' %}" />
+        <img class="img-responsive" src="{% url 'weekday_jobs_image' %}" />
+        <img class="img-responsive" src="{% url 'weekend_jobs_image' %}" />
             <img class="img-responsive" src="{% url 'semester_histogram' %}" />
```
chriskuehl: indentation is off

ocfweb/stats/semester_job.py
```diff
@@ -0,0 +1,89 @@
...
+              ylab='Number of Jobs Printed'):
+    """takes in data, title, and ylab and makes a histogram, with
+    the 1:len(data) as the xaxis
+    """
+    fig = Figure(figsize=(10, 4))
+    ax = fig.add_subplot(1, 1, 1)
+
+    tickLocations = range(1, len(data) + 1)
+    width = 0.8
+    ax.bar(tickLocations, data, width, color='blue')
```
matthew-mcallister: I would leave `color` at default so it matches with the existing graphs on the page.

shex1627:
Applied the change need with the commit but having trouble squashing the old commits relating to this issue because I pull longs of other commits between my first commit(related to the graph) and the last commit. 
I tried using "git rebase SemesterJob --interactive" but failed to do the work, there is what I tried 
pick commit1
squash commit2
squash commit3
squash commit4
squash commit5

shex1627:
So guys are we ready to merge now?

chriskuehl:
@shex1627 fixnship means we trust you to fix whatever tiny issue we pointed out and then merge at will, without waiting for another review. So if you're happy with the branch, feel free to merge :)

shex1627:
@chriskuehl get it, thanks for the explanation.

PR #269, opened 2017-03-12
docs: add DGM and DSMs

kpengboy:
What do you think of the layout? Any other suggestions on how to display DGMs and DSMs?

### DIFF
--- a/ocfweb/docs/templates/docs/officers.html
+++ b/ocfweb/docs/templates/docs/officers.html
@@ -25,14 +25,25 @@
       <h3>
         General Managers
       </h3>
-        {% for gm in current_term.gms %}
-          {% if gm.end is None %}
-            <p>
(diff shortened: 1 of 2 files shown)


chriskuehl:
include screenshot?

if we're going to include non-elected officers (sounds great), how about including BoD somewhere as well? perhaps just including the elected officers on the main page, with a link to a page-per-semester listing appointed officers (DMs and such as well) and BoD? we could fill that out for at least the past few years and encourage alumni to contribute information.

NickImpicciche:
https://i.fluffy.cc/H1BgtBqcZ6FmHQMQLrG9ZLXbdXk7bhk4.html

Here are some screenshots

kpengboy:
Merging, we can add a separate appointed officers page later if we want to.

PR #271, opened 2017-03-17
add ocfweb dev_config info to staffvm tricks

tahabi:


ocfweb/docs/docs/staff/staffvm/ocfweb.md
```diff
@@ -0,0 +1,26 @@
...
+    classes:
+        - ocf_ocfweb::dev_config
+
+Take a look at the configs for [`fireball`][2] or [`raptors`][3] for
+examples.
+
+If you are in `ocfroot` you can push this directly to puppet and
+trigger a puppet run on your staffvm (`sudo puppet-trigger -fe <user>`)
+otherwise, push to your fork and submit a pull request and someone will
+merge it for you, after which you can trigger the puppet run on your VM.
```
jvperrin: I'd probably suggest just making a pull request for this anyway, but yeah, either way is fine, it's just a bit more complicated to explain this way.

ocfweb/docs/docs/staff/staffvm/ocfweb.md
```diff
@@ -0,0 +1,26 @@
+[[!meta title="Running ocfweb on your Staff VM"]]
+
+If you want to work on [`ocfweb`][0] on your Staff VM, it isn't as
+simple as cloning the repo and running `make dev`, unfortunately.
+In order to do so, you will need to add the `ocfweb` dev config to
+your Staff VM's Puppet configuration, so Puppet can install
+`ocfweb`'s dependencies. To do this, clone the [puppet repo][1] or
```
jvperrin: It's not really about dependencies, actually those still need to be installed manually (PRs on this are welcome), but instead it adds config files with secrets like the password to connect to redis on supernova, and the password to connect to the `ocfmail` database. [The manifest](https://github.com/ocf/puppet/blob/master/modules/ocf_ocfweb/manifests/dev_config.pp) really doesn't do much besides provide a few files.

tahabi: Like Celery getting uninstalled every now and then and needing to be manually reinstalled? I think redis also needs to be installed, or at least, I had to install it, though that may have been a mistake on my end. Would you suggest s/dependencies/configuration requirements/?

ocfweb/docs/docs/staff/staffvm/ocfweb.md
tahabi: I don't imagine people will ocfroot will really bother, tbh

tahabi:
I'm not really too familiar with the Docker and SSL stuff, but I'm happy to write the documentation for it here once I become more familiar with it. As you demonstrated, I wasn't particularly familiar with what this manifest was doing either, I just trusted that it worked, lol.

ocfweb/docs/docs/staff/staffvm/ocfweb.md
jvperrin: Not sure what's happening on those, I don't think celery being uninstalled would be related to this, the dependencies I was thinking of were more like development headers to that things like compile correctly (like I believe the `cracklib` package needs some extra packages to be installed to run). Also I don't think redis should be needed, maybe a redis client or something? I thought that was installed with the requirements though python instead.

Yeah, or maybe `config files`, just to be more accurate anyway.

jvperrin:
No, that's fair, it just seems like a good thing to add in the same vein as this change. Thanks for doing this!

ocfweb/docs/docs/staff/staffvm/ocfweb.md
chriskuehl: yeah redis and celery don't have system dependencies. this is strictly to provide the dev config.

there *are* some system dependencies (libcrack2-dev or whatever) but the dev config doesn't attempt to provide all of these (and probably shouldn't, since the list is hard to maintain)

chriskuehl: > Like Celery getting uninstalled every now and then and needing to be manually reinstalled?

this was because we switched to my fork which supports ssl for backends, and pip doesn't understand the order to uninstall and install packages which provide the same files

PR #272, opened 2017-03-17
remove unnecessary h_ tag in html ids/quicklinks

tahabi:
With the previous formatting, linking to sections in our docs gives things like this: "https://www.ocf.berkeley.edu/docs/services/vhost/#h4_university-policies". This would make it so links now look like this "https://www.ocf.berkeley.edu/docs/services/vhost/#university-policies" which IMHO is cleaner.

### DIFF
--- a/ocfweb/component/markdown.py
+++ b/ocfweb/component/markdown.py
@@ -108,7 +108,7 @@ class HeaderRendererMixin:
     """Mixin to render headers with auto-generated IDs (or provided IDs).
 
     If headers are written as usual, they'll be given automatically-generated
-    IDs based on their header level and text.
+    IDs based on their text.
 
     Headers can also be specified with an ID at the end wrapped in curly braces:
 
@@ -139,10 +139,7 @@ class HeaderRendererMixin:
             if id in self.toc_ids:
                 raise ValueError('Duplicate header ID in Markdown: "{}"'.format(id))
         else:
-            id = 'h{level}_{title}'.format(
-                level=level,
-                title=re.sub('[^a-z0-9\-_ ]', '', strip_tags(text).lower()).strip().replace(' ', '-'),
-            )
+            id = re.sub('[^a-z0-9\-_ ]', '', strip_tags(text).lower()).strip().replace(' ', '-')
 
             # dumb collision avoidance
             while id in self.toc_ids:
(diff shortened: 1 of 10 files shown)


tahabi:
Of course this breaks all the old links that have already been distributed, but the damage is minimal because the correct page will still load, it just won't jump to the section right away.

jvperrin:
Yeah, that's true, but I think that's fine too. It would be good to grep through all the repos we have to check that there aren't any broken links there though, if you haven't done that already. I did look through templates and short URLs, and I couldn't see any there, but maybe there still are?

tahabi:
Nice catch, there are two places in utils/banner_mass_email.py that use the old link styles. I only have forks of ocfweb, ocflib, utils, puppet, and dns though, but I can't find anything using GitHub code search either.

chriskuehl:
I don't think this is a good idea. It was done for a reason: otherwise it's not clear which headings are considered to have "permanent" IDs that can be reliably linked to, and which are not.

If you undo this, it will become easy for somebody to change a heading in a well-meaning way (e.g. to improve the wording) but to accidentally break a bunch of incoming links in the process. Currently we have no incoming links to the implicit IDs ourselves (and imo we should keep it that way).

We should instead assign actual IDs to headings that we intend to link to, e.g. https://ocf.io/vhost#disclaimer

There's sort of a compromise here: the implicit IDs are nice for referring to things in a non-permanent way (e.g. linking on IRC), but we should really be picking explicit IDs for things we want to link to otherwise.

tahabi:
I don't think it would be a problem to keep this and incrementally add the relevant explicit IDs as well. The informal shortlinks get nice URLs and important things that we link to often we give short, descriptive IDs to as well. fd2ffb0 for example.

chriskuehl:
I still don't think we should do this per my comment above. Why do you want to remove the heading level?

Baisang:
I think "https://www.ocf.berkeley.edu/docs/services/vhost/#h4_university-policies" is just as informative as without the `#h4`. If there aren't common situations where that it will get unreasonably ugly, then I think what @chriskuehl is saying makes more sense. We should just make it explicit when needed.

chriskuehl:
The whole problem is you *can't* tell what you can rely on and what you can't without the heading at the front. If you want to add another way to determine that, that's fine. But right now this PR removes that signal and makes the situation worse.

matthew-mcallister:
I don't really have anything to add here, but I agree with @chriskuehl. This seems like a change for change's sake.

PR #275, opened 2017-03-26
Dutifully destroy documents delivered by alumni

matthew-mcallister:
This isn't really to reassure alumni (most don't express concern for their privacy when they email in docs) but to make sure staff delete private data when it's no longer needed.

### DIFF
--- a/ocfweb/docs/docs/services/account.md
+++ b/ocfweb/docs/docs/services/account.md
@@ -101,6 +101,9 @@ For security purposes, please include the following with your request:
 If you no longer have your Cal ID, you may substitute it with another
 government-issued ID.
 
+Out of respect for your privacy, any copies of documents you send us will be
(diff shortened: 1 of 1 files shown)


chriskuehl:
I'd just say "staff will delete", saying we will destroy them seems unnecessarily precise (e.g. they probably continue to live on Google's servers for a long time)

matthew-mcallister:
Mm, well, maybe it would be more accurate to say "the OCF will destroy its copies of any documents received". The language needs to make sense for copies sent over either email or snail mail.

chriskuehl:
destroy is probably still inaccurate for snail mail given that the nsa probably keeps their copies for a while

matthew-mcallister:
No, it is true that the OCF will destroy _its_ copies of received attachments. No guarantees about any other copies though!

PR #277, opened 2017-04-05
update to django 1.11

tahabi:
seems like the right way to do this


tahabi:
Doesn't seem to be anything too major in the changelogs, AFAICT. Mostly minor version increases.

jvperrin:
This got rolled back since it broke the Django bootstrap form package, so we should try to figure out how to fix that.

PR #279, opened 2017-04-07
Move constants

tahabi:
continue updating to remove ocflib.constants

obviously this build will not work until ocflib is updated


jvperrin:
Looks good apart from `CREATE_PUBLIC_KEY`, the ocflib part of that obviously needs to be done first. The rest looks good though.

jvperrin:
Once ocflib tests pass, I'll re-trigger the test for this, but it should pass now! :D

jvperrin:
> Using cached ocflib-2017.4.11.17.47-py2.py3-none-any.whl

:(

jvperrin:
Anyway, I believe this should work when merged, but I don't know why ocflib was cached

PR #281, opened 2017-04-12
Allow faculty to request virtual hosts

matthew-mcallister:
Not the best code ever, but apparently there are already people lined up for this, and I think it would be helpful in the long run to make the form available for everyone rather than to have them email us and negotiate on a case-by-case basis.

### DIFF
--- a/ocfweb/account/templates/account/vhost/index.html
+++ b/ocfweb/account/templates/account/vhost/index.html
@@ -4,20 +4,20 @@
 {% block content %}
     <div class="ocf-content-block">
         <p>
-            One of the benefits of hosting your student group's site with the OCF is that you have the option of using an address like "mygroup.berkeley.edu" to access your site.
-        However, in order to obtain a berkeley.edu name for your site, you must meet both university policies regarding acceptable use of berkeley.edu names, as well as OCF policies for virtual hosting.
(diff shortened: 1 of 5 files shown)


matthew-mcallister:
@NickImpicciche quite rightly pointed out that several people have been waiting to hear about this, so it'd be a good idea to address issues and merge soon.

ocfweb/account/templates/account/vhost/index.html
```diff
@@ -4,33 +4,42 @@
...
+                        <a href="{{group_url}}">{{group_url}}</a>
+                    </li>
+                {% else %}
+                    <li><strong>Name:</strong> {{attrs.cn.0}}</li>
+                    <li><strong>OCF Account:</strong> {{user}}</li>
+                    <li>
+                        <strong>Current OCF-Hosted Website:</strong>
+                        <a href="{{group_url}}">{{group_url}}</a>
+                    </li>
+                {% endif %}
```
chriskuehl: lol do we really need this branch?

matthew-mcallister: I think we do so people requesting a vhost for their group don't put _their_ name at the top, scroll down, then realize it was supposed to be the _group_ name at the top.

Speaking of, I just realized the form now asks faculty for their name twice (oops). Thanks for inadvertently calling my attention to it.

PR #282, opened 2017-04-13
add blurb about relevant subdomain

tahabi:
[rt#5167](https://ocf.io/rt/5167)

### DIFF
--- a/ocfweb/docs/docs/services/vhost.md
+++ b/ocfweb/docs/docs/services/vhost.md
@@ -135,11 +135,11 @@ does not process nor advise off-site hosting requests, which can be
 [submitted directly to the university][offsite] (however, feel free to keep
 us informed if you have an existing OCF account).
 
-In other words, to comply with university policy, the OCF may only provide
-virtual hosting for websites that are hosted on OCF servers.
+   In other words, to comply with university policy, the OCF may only provide
+   virtual hosting for websites that are hosted on OCF servers.
 
-* **Relevant domain name**: Student Organizations may only use domains or
-request subdomains that are the officially recognized name of the group, or
+* **Relevant domain name**: Registered Student Organizations may only use domains
+or request subdomains that are the officially recognized name of the group, or
 an officially recognized short name of the group. Exceptions to this rule must
 be approved by the LEAD Center prior to requesting the domain from the OCF.
 In general, the subdomain must reasonably and uniquely identify the student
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/services/vhost.md
```diff
@@ -138,6 +138,14 @@ us informed if you have an existing OCF account).
 In other words, to comply with university policy, the OCF may only provide
 virtual hosting for websites that are hosted on OCF servers.
 
+* **Relevant domain name**: Student Organizations may only use domains or
```
chriskuehl: uncapitalize organizations? (or use the full name, Registered Student Organizations)

ocfweb/docs/docs/services/vhost.md
```diff
@@ -135,9 +135,17 @@ does not process nor advise off-site hosting requests, which can be
 [submitted directly to the university][offsite] (however, feel free to keep
 us informed if you have an existing OCF account).
 
-In other words, to comply with university policy, the OCF may only provide
+   In other words, to comply with university policy, the OCF may only provide
```
jvperrin: Why the extra space?

tahabi: Logically this paragraph belongs to the above bullet point, which is indented. Without the space, this paragraph is not indented. With only one bullet point, it was fine, but with two, it makes sense to promote it.

jvperrin: Hmm, I suppose so, but I think the bullet point thing is indented only because there is a bullet point, especially since the rest of the lines in the paragraph are not indented, so I see no reason for this paragraph to start with indentation either.

tahabi: It's not indented in the markdown, but when rendered, the bullet list gets indented because of the list. Adding the space forces the paragraph to be indented when rendered.

screenshot_2017-04-13_23-02-52

vs

screenshot_2017-04-13_23-07-26

without the indentation, that paragraph renders unindented and it breaks the continuity of the bullet list.

jvperrin: Maybe have the whole paragraph indented then instead? I feel like that would be better than having just the indentation at the beginning, since that looks sorta like it shouldn't be there.

tahabi: makes sense

PR #285, opened 2017-04-19
Add anchor ids to docs

tahabi:
draft, needs work

### DIFF
--- a/ocfweb/docs/docs/faq.md
+++ b/ocfweb/docs/docs/faq.md
@@ -1,13 +1,13 @@
 [[!meta title="Frequently asked questions"]]
 
 
-#### Why might I want an OCF account? What do I get? How do I use OCF services?
+#### Why might I want an OCF account? What do I get? How do I use OCF services?    {why}
 
 Access to computing by the students for the students!
 
 More specifically, we have a [[list of services|doc services]].
 
-#### Who are you?
+#### Who are you?    {who}
 
 The largest(\*) student organization on campus (\*by membership count). OCF is
 completely administered and operated by student volunteers ("staff"). [[About
@@ -16,15 +16,15 @@ the OCF|doc about]].
 Interested in [[joining staff|about-staff]]? Feel free to attend our meetings
 or [[contact us|doc contact]] by email.
 
-#### Where is the lab? What are the hours?
+#### Where is the lab? What are the hours?    {where}
 
 See [[lab|doc services/lab]].
 
-#### How do I join the OCF and get an OCF account?
+#### How do I join the OCF and get an OCF account?    {how}
 
 See [[Join the OCF|doc membership]].
 
-#### How do I join [[OCF staff|doc staff]]?
+#### How do I join [[OCF staff|doc staff]]?    {how-staff}
 
 Check out [[our staff recruitment website|about-staff]]!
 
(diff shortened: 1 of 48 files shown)


tahabi:
brb fixing merge conflicts

jvperrin:
This is related to #272 as well, just to connect the two.

tahabi:
I eyeballed it to avoid the repeated id thing, I think I got most of it right. I grepped the diff and it looks ok.

Also I checked the last few months of git history and it seems like we rarely change the header names and therefore ids: at most we add new sections or rewrite content but rarely changed the header titles and even in the cases we do, we almost never actually link to them, so we should be safe on that end.

chriskuehl:
The markdown parser raises an exception if you have duplicate heading IDs: https://github.com/ocf/ocfweb/blob/d6fcc8dd0acf2fafb0b193d9aeb989a410adfd99/ocfweb/component/markdown.py#L139-L140

This was one of the things I really wanted to do with ocfweb (make it harder to create broken links, broken anchors, etc. by failing loud during the tests).

kpengboy:
Lgtm, although I'm not sure every one of those anchors is necessary

ocfweb/docs/docs/faq.md
```diff
@@ -1,13 +1,13 @@
 [[!meta title="Frequently asked questions"]]
 
 
-#### Why might I want an OCF account? What do I get? How do I use OCF services?
+#### Why might I want an OCF account? What do I get? How do I use OCF services?    {why}
```
kpengboy: I'd use `{why-join}` in case we add another why question in the future

tahabi:
We link to certain places fairly often, and they should have regular-looking or permanent ids: e.g. the university disclaimer, the hosting banner, links to wordpress, etc. We also tend to update these pages somewhat regularly. However, we rarely change the titles of the sections, we almost always only change the content or add new sections. Historical evidence (i.e. `git log -p -S'###' -- ocfweb/docs/`) suggests that renaming headings (especially in the most-visited docs) is not a common occurrence and unlikely to be a problem. It also fails safe: the correct page will always load, and if the id isn't found, it will just drop the user at the top of the page, which is what would happen anyways if the user found the page on their own.

Other place we've never linked to at all, which is the vast majority of the pages modified/links added here, like in the staff documentation. Even if we change the ids here, it's not likely to be a problem because it's the staff documentation and staff are likely to be linked to them through transient mediums like IRC.

Furthermore, most of the methods by which we distribute these links are transient. I doubt users are digging through their emails to find the email they received from OCF staff and clicking through those links to find where we directed them in our docs, and even if we do, it's highly unlikely that the links will be broken, because we rarely change the headings, and if we do, it fails safe anyways.

Ultimately, it's an update with a non-breaking and bounded downside and an aesthetically pleasing upside.

chriskuehl:
> We link to certain places fairly often, and they should have regular-looking or permanent ids: e.g. the university disclaimer, the hosting banner, links to wordpress, etc

I totally agree with you that we can and should add permanent IDs to important things that we often link to. Many of them already have them and we can add more where they're missing.

> Other place we've never linked to at all, which is the vast majority of the pages modified/links added here, like in the staff documentation. Even if we change the ids here, it's not likely to be a problem because it's the staff documentation and staff are likely to be linked to them through transient mediums like IRC.
> Furthermore, most of the methods by which we distribute these links are transient. I doubt users are digging through their emails to find the email they received from OCF staff and clicking through those links to find where we directed them in our docs, and even if we do, it's highly unlikely that the links will be broken, because we rarely change the headings, and if we do, it fails safe anyways.

For me this begs the question: why change them at all, then, if they're only linked from one-off emails, IRC, etc? The current anchors work fine for that.

I feel like you are trying to solve a problem that I'm not seeing. Can we step back and look more at what problem you're trying to solve? I'm not sure I understand what you're trying to fix right now.

I'm pretty confident we can find a solution to this problem that involves less maintenance cost. I think so far we've focused too much on proposed solutions rather than the problem itself.

PR #288, opened 2017-04-26
Explain how to install packages in README

matthew-mcallister:


requirements-minimal.txt
```diff
@@ -6,7 +6,7 @@ django-bootstrap-form
 django-ipware
 django-mathfilters
 django-redis
-django>=1.10,<1.10.999
+django~=1.10
```
matthew-mcallister: Any reason not to use this?

chriskuehl: It is just one more thing to hate about the awful PEP440. You can read some comments in https://github.com/pypa/pip/issues/1551. I feel it's a bad idea (like almost everything else in that terrible PEP) because it's not obvious at all what it means to anyone looking at it.

However given that it has been officially adopted, I see no reason to block you.

matthew-mcallister: OK, I only made this change to ask why this syntax wasn't used. It seems less flexible than the explicit syntax anyways.

It looks a little silly to use an arbitrary high version number like `.999` though. What about `django>=1.10<1.11.0`?

chriskuehl: The problem is that 1.11.0a1 (an alpha release) satisfies that version spec :(

I think it is okay as-is. But personally I won't be using this new syntax, at least for a while.

matthew-mcallister: > The problem is that 1.11.0a1 (an alpha release) satisfies that version spec :(

Wow, that just makes me cringe. I won't actually merge this particular change since it doesn't accomplish anything; basically I was curious.

matthew-mcallister:
Also sorry for pinging you @jvperrin, I mistook you for writing that one line.

PR #289, opened 2017-04-27
Update requirements

chriskuehl:
Upgrading Django for https://www.debian.org/security/2017/dsa-3835

(didn't actually check if we're affected, but this probably doesn't hurt)


chriskuehl:
what could go wrong?

matthew-mcallister:
For the record, 1.10.5 is affected by 7233 and 7234, but ocfweb does not use the insecure code.

PR #291, opened 2017-04-30
Don't load officers' LDAP information on import

chriskuehl:
This makes importing ocfweb almost 13 seconds faster.

One problem is now this page will take forever to load the first time. Maybe we should make it possible to have periodic functions with an infinite TTL so that the worker will load it instead?


jvperrin:
Nice, this will help especially when `ocfweb/web` is deploying, since this happens multiple times in a row close together, generating more load on firestorm.

Why did the tests fail with serialization errors?

chriskuehl:
@jvperrin because pickle is a giant terrible hack :(

The "name" of the namedtuple was `ocfweb.docs.views.officers.Term` even though that is actually a _function_ that creates the namedtuple, rather than the namedtuple class itself. If you change the namedtuple to not lie about its name, then it works.

Basically: pickle tries to import the class so that it can re-create the instance based on the pickled data, but it imports the function rather than the class because the name was a lie.

chriskuehl:
I also changed this into a periodic function with "infinite" period. That should mean that the worker will load it once and then never refresh it (and also it will never get evicted from the cache), which seems ideal. This way there's no slow first request (unless the user gets really unlucky and hits it before the worker runs that function).

PR #293, opened 2017-05-20
add mirrors stats to stats page

tahabi:
[rt#6177](https://ocf.io/rt/6177)

mirrors_stats

Maybe some pretty line graphs in the future

### DIFF
--- a/ocfweb/stats/mirrors.py
+++ b/ocfweb/stats/mirrors.py
@@ -1,7 +1,6 @@
 from datetime import date
 
 from django.shortcuts import render
-from humanize import naturalsize
 from ocflib.lab.stats import current_semester_start
 from ocflib.lab.stats import get_connection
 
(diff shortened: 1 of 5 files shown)


tahabi:
I have a cronjob on fallingrocks that should update ocfstats.mirrors with the latest data from the logs, modeled on @jvperrin's script. I'll puppet it eventually.

ocfweb/stats/mirrors.py
```diff
@@ -0,0 +1,42 @@
...
+            'start_date': current_semester_start,
+        }
+    )
+
+
+def _bandwidth_by_dist(start):
+    with get_connection() as c:
+        c.execute(
+            'SELECT `dist`, SUM(`up` + `down`) as `bandwidth` FROM `mirrors_public` WHERE `date` > %s'
+            'GROUP BY `dist` ORDER BY `bandwidth` DESC', (start))
```
chriskuehl: start isn't a tuple here, either add a comma (like `(start,)`) or drop the unnecessary parens

ocfweb/stats/mirrors.py
```diff
@@ -0,0 +1,42 @@
...
+
+    return [(i['dist'], naturalsize(i['bandwidth'])) for i in c]
```
chriskuehl: can we avoid adding another dependency for this and just write the function ourselves? here's the implementation i usually use:

```
def human_size(bytes):
    for label, n in (
        ('TB', 2**40),
        ('GB', 2**30),
        ('MB', 2**20),
        ('KB', 2**10),
    ):
        if bytes > n:
            return '{:.1f} {}'.format(bytes / n, label)
    else:
        return '{} B'.format(bytes)
```

ocfweb/stats/templates/stats/mirrors.html
```diff
@@ -0,0 +1,42 @@
+{% extends 'base.html' %}
+{% load common %}
+{% load stats %}
+
+{% block content %}
+<div class="ocf-content-block">
```
chriskuehl: indent the stuff inside `{% block %}`?

PR #295, opened 2017-07-09
Update Django docs to work with new Django

chriskuehl:
Removed all the docs about actually making a new Django app, they were outdated and not exhaustive and probably most people already have an app by the time they're looking for hosting. Replaced that with a link to the official Django docs.

I tested this with my own account. It did work eventually, but initially I made a typo in `run.fcgi` which caused the first version of my app to fail, and then I was never able to get it to work until after I restarted apache on death :\

Something else to look into...

(Note that if you're going to attempt these steps yourself, you'll want to do it from tsunami and not supernova, since supernova is running stretch whereas the webserver is still jessie.)

### DIFF
--- a/ocfweb/docs/docs/services/web/django.md
+++ b/ocfweb/docs/docs/services/web/django.md
@@ -1,91 +1,99 @@
 [[!meta title="Django"]]
 
-Django is a popular web framework for Python applications.
+Django is a popular web framework for Python applications. At the OCF, you can
+host it using our regular web hosting, though for groups with virtual hosting,
+we strongly recommend instead using our [[application hosting service|doc
+services/webapps]] instead, as it provides a first-class solution for webapps.
 
+## Running a Django app
 
-## Starting a Django project
+The steps below assume you have already created a Django application. If you
+haven't, take a look at [the Django documentation][django-docs] for help with
+that.
 
-The necessary scripts for you to create Django projects are installed on our
-login server, _ssh.ocf.berkeley.edu_.
+1. Copy your Django application to your OCF account. For example, you can use
+   `scp` to upload it, or if you're using source control, clone it using `git`
+   (or another tool) into your home directory.
 
-1.   First, you will want to create a directory to store all of your Django
-     projects. This directory should not be in your *public_html* directory.
+   In the steps below, we'll assume you put it at `~/app` in your home
+   directory.
 
-         tsunami$ mkdir ~/projects
+2. Inside your project directory, create a virtualenv and install your
+   dependencies. For example, to create a Python 3 virtualenv, try:
 
-1.   Go into the directory you just created:
+   ```bash
+   user@tsunami:~/app$ virtualenv -ppython3 venv
+   user@tsunami:~/app$ venv/bin/pip install django
+   ```
 
-         tsunami$ cd ~/projects
+   You'll want to install at least `django` to start with. If you're using
+   other packages, install those too. You can also install from a
+   `requirements.txt` file, if you have one.
 
-1.   Create a new Django project. For this example, the project name will be
-     "hatsune". You will want to stick with Python-friendly package and module
-     names, as detailed in [PEP
-     8](https://www.python.org/dev/peps/pep-0008/#package-and-module-names).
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/services/web/django.md
```diff
@@ -1,91 +1,97 @@
...
+   ```python
+   #!/home/u/us/user/app/venv/bin/python
+   import os
+   import sys
 
-         tsunami$ django-admin startapp miku
-         tsunami$ vim miku/views.py
-         ...
-         tsunami$ vim urls.py
+   sys.path.insert(0, os.path.expanduser('~/app'))
```
matthew-mcallister: The benefit of calling expanduser seems pretty minor if you already have the full path in the shebang...

ocfweb/docs/docs/services/web/django.md
```diff
@@ -1,91 +1,97 @@
...
-        tsunami$ mkdir ~/public_html/hatsune
+   if __name__ == '__main__':
+       WSGIServer(wsgi.application).run()
+   ```
 
-1.    Go into the directory you just created, and create two files: *.htaccess*
-      (note the dot), and *run.fcgi*.
+   Make sure to replace the first line of `run.fcgi` file with the actual path
+   to your project's virtualenv Python. You can find it by running the command
+   `readlink -f ~/app/venv/bin/python`.
```
matthew-mcallister: It doesn't hurt to use the exact path but you could just do `echo` instead and use the symlink.

ocfweb/docs/docs/services/web/django.md
```diff
@@ -1,91 +1,97 @@
...
-      projects_location = "/home/k/ke/kedo/django"
-      project_name = "hatsune"
+Your app should now be accessible!
 
-      sys.path.insert(0, projects_location)
-      sys.path.insert(1, os.path.join(projects_location, project_name))
-      os.chdir(projects_location)
+Once your app has started running, changes you make to the Python code or
+templates won't take effect for a few hours. To apply changes immediately, you
+can touch the `run.fcgi` file with the command:
```
matthew-mcallister: Maybe say "make the `run.fcgi` file more recent" or "tell the webserver to reload your app" instead of "touch" since it's slightly esoteric terminology.

ocfweb/docs/docs/services/web/django.md
chriskuehl: what symlink are you talking about?

ocfweb/docs/docs/services/web/django.md
chriskuehl: this is just copy/paste from before

ocfweb/docs/docs/services/web/django.md
chriskuehl: one replacement vs two

ocfweb/docs/docs/services/web/django.md
matthew-mcallister: `venv/bin/python` should be a link to `venv/bin/python3`; I thought you specified to use `readlink` because you wanted the path to the binary?

chriskuehl: i included that because you can't put `~/apps/...` in a shebang (no squiggly expansion)

i figured it was easier to say "run `readlink -f ~/apps/...`" than "use the full path, such as /home/c/ck/ckuehl" and explain to use the first letter, then the first two letters, than the full name

matthew-mcallister: I mean, you could replace `readlink -f` with `echo` for that purpose.

chriskuehl: you could yes

chriskuehl:
@matthew-mcallister updated the paragraph about touching the fcgi file

ocfweb/docs/docs/services/web/django.md
matthew-mcallister: Ah, that's a pretty good point, hadn't thought of what happens if you use a relative path.

matthew-mcallister:
Awesome, thanks for taking the suggestion.

PR #301, opened 2017-08-18
Document Mosh support

matthew-mcallister:


matthew-mcallister:
Great! Thanks for your suggestion.

PR #304, opened 2017-08-23
Unfork celery, replace pycrypto with pycryptodome

### LINKED ISSUE #299: Upgrade Celery to 4.1.0
Celery 4.1.0 was just released, so we can unfork since @chriskuehl's changes will be included!

jvperrin:
Also unpin `pyasn1` since we can use a newer version of `pysnmp` now. Fixes #299


requirements-minimal.txt
```diff
@@ -1,7 +1,5 @@
 cached-property
-# TODO: stop using our fork
-# https://github.com/celery/celery/pull/3831
-ckuehl-celery[redis]==4.0.2.post1
+celery[redis]==4.1.0
```
chriskuehl: do we need to pin here?

also need to update requirements.txt (or run make update-requirements) for this to actually do anything

jvperrin: Ah, you're right, we don't actually need to pin there lol, thanks for catching that

PR #306, opened 2017-08-24
changed hours spreadsheet

asaiacai:
The hours that are currently being used by opstaff is linked below.
https://docs.google.com/spreadsheets/d/1FHgW0wCnAh49bZIw54eoBT9xEoJ3AGG7Y2LhQdi0Lvs/edit#gid=0
we might need to consider changing the table entries to opstaff usernames depending our needs

### DIFF
--- a/ocfweb/api/hours.py
+++ b/ocfweb/api/hours.py
@@ -9,8 +9,8 @@ from django.http import JsonResponse
 
 from ocfweb.caching import periodic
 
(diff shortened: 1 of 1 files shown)


jvperrin:
Yeah, I do think we might want more granularity on the shifts to have half-hour shifts in the future, but using the spreadsheet they are actually updating is a good idea.

tahabi:
Why not just have opstaff edit the original spreadsheet? It has range protection and data validation to make sure errant entries didn't get added.

tahabi:
considering asai made the original spreadsheet as early as June, why was a second spreadsheet distributed instead of the original one? I mean, either way it just means more work for us.

PR #311, opened 2017-08-30
remove old references to pmc

Baisang:
mostly just used "front desk staff" when possible b/c people won't know what OCF Operations Staff are, and am worried that people will confuse that more with OCF (volunteer) staff.

### DIFF
--- a/ocfweb/docs/docs/services/lab/printing.md
+++ b/ocfweb/docs/docs/services/lab/printing.md
@@ -27,7 +27,7 @@ Check your print balance regularly!
 ## Printing refunds
 
 Pages are **not** normally refunded, except in cases of technical issues. In
-this case, you should talk with the Publications Center staff at the front desk
+this case, you should talk with the OCF Operations Staff at the front desk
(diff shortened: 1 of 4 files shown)


ocfweb/docs/docs/services/lab/printing.md
```diff
@@ -27,7 +27,7 @@ Check your print balance regularly!
 ## Printing refunds
 
 Pages are **not** normally refunded, except in cases of technical issues. In
-this case, you should talk with the Publications Center staff at the front desk
+this case, you should talk with the OCF Operations Staff at the front desk
```
jvperrin: This isn't actually correct though, right? Like they can't give refunds currently

Baisang: technically not lmao but at least they can direct you to get refunds or they can print the thing for you

jvperrin: Yeah, less inaccurate than it was before at least

PR #313, opened 2017-09-02
hours: fix for display

tahabi:
I don't think this is a good long-term solution but it should work for now. Once the other parts of big brother are in place wrt tracking if opstaff actually came to their shifts I think this should be rewritten.

### DIFF
--- a/ocfweb/api/hours.py
+++ b/ocfweb/api/hours.py
@@ -51,6 +51,14 @@ def _get_hours():
     return all_shifts
 
 
+def _merge_shifts(first, second, staffer=True):
+    return Hour(
+        open=min(first.open, second.open),
+        close=max(first.close, second.close),
+        staffer=first.staffer if staffer else None,
+    )
+
+
 def _combine_shifts(shifts):
     """combine a list of shifts into a list of Hour()s.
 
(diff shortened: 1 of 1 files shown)


ocfweb/api/hours.py
```diff
@@ -51,6 +51,14 @@ def _get_hours():
     return all_shifts
 
 
+def _merge_shifts(first, second, staffer=True):
+    return Hour(
+        open=min(first.open, second.open),
+        close=max(first.close, second.close),
+        staffer=first.staffer if staffer else None,
```
jvperrin: I don't really like having this pattern of having a staffer associated with each `Hour` object, even if it doesn't actually mean anything. It's not a big deal for this pull request at least, but it seems like a pattern that should be fixed in the future sometime.

ocfweb/api/hours.py
```diff
@@ -71,41 +79,71 @@ def _combine_shifts(shifts):
...
 
-@periodic(5 * 60)
 def _generate_regular_hours():
     raw_hours = _get_hours()
     return {day: _combine_shifts(shifts) for day, shifts in raw_hours.items()}
 
 
+@periodic(5 * 60)
+def display_hours():
+    """merge shifts even further for display.
```
jvperrin: Capitalize?

ocfweb/api/hours.py
tahabi: The idea is that in the near future we can use this to run reports of sorts on whether or not staffers actually showed up to their shifts. If we integrate labstats into ocfweb, we can then have everything run nicely and concisely.

ocfweb/api/hours.py
tahabi: good catch x 2

ocfweb/api/hours.py
jvperrin: That's a good point with labstats being merged, but maybe it would be possible to just not even need to put a staffer here and have it be optional or something?

ocfweb/api/hours.py
jvperrin: Meh, pretty small catch honestly :P

jvperrin:
Ok, I think this is ready unless @chriskuehl (or anyone else) has any comments?

Baisang:
let's try to merge this asap, hours look really bad rn

tahabi:
did the new hours change the URL, and if so, if we're going to use that as
the final spreadsheet, let's do the data validation stuff in that sheet and
then I'll update the link and we'll ship it.

jvperrin:
No, the same spreadsheet is used for the new shifts, but now things seem to be even more broken, like the hours are actually wrong sometimes...

tahabi:
I'm investigating this, no logic changed changed and it seems reasonable so
I'll test some things and see if I can figure out what's going on.

PR #314, opened 2017-09-02
django 1.11: third time's the charm

tahabi:
hopefully, at least. 

django-bootstrap-form 3.3 has been released, as noted earlier, with fixes for Django 1.11.


tahabi:
good catch

tahabi:
users are the best QA

PR #315, opened 2017-09-05
fix hour display bug

Baisang:
We were dropping a lot of shifts before resulting in perforated hours on the site.
After much pdb-ing, issue was found to be wrong variable name used.


tahabi:
i am an idiot

PR #316, opened 2017-09-14
Added table to lifetime printing stats that estimates number of trees…

redplanks:
… killed by the printers.

### DIFF
--- a/ocfweb/stats/templates/stats/printing/pages-printed.html
+++ b/ocfweb/stats/templates/stats/printing/pages-printed.html
@@ -11,7 +11,7 @@
                 <table class="table table-hover table-responsive">
                     <tr>
                         <th>Printer</th>
-                        <th>Trees killed over lifetime</tr>
+                        <th>Trees killed over lifetime*</th>
                     </tr>
                     {% for dict in data %}
                         <tr>
@@ -21,6 +21,7 @@
                         </tr>
                     {% endfor %}
                 </table>
+                <small>*Assuming each tree yields ~8333 pages.</small>
             </div>
         </div>
     </div>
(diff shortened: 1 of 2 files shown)


ocfweb/stats/templates/stats/printing/pages-printed.html
```diff
@@ -6,6 +6,23 @@
     <div class="ocf-content-block">
         {% stats_navbar %}
         <div id="chart"></div>
+        <div class="row">
+            <div class="col-md-12">
+                <table class="table table-hover table-responsive">
+                    <tr>
+                        <th>Printer</th>
+                        <th>Trees used over lifetime</tr>
```
chriskuehl: s/used/killed/ maybe?

ocfweb/stats/templates/stats/printing/pages-printed.html
```diff
@@ -6,6 +6,23 @@
     <div class="ocf-content-block">
         {% stats_navbar %}
         <div id="chart"></div>
+        <div class="row">
+            <div class="col-md-12">
+                <table class="table table-hover table-responsive">
+                    <tr>
+                        <th>Printer</th>
+                        <th>Trees killed over lifetime</tr>
```
tahabi: s/</tr>/</th>/

also I'd give these a more explicit title with an explanation of how we got the numbers, e.g. "Lifetime trees killed per printer assuming 8333 pages per tree" or something similarly descriptive but less verbose

redplanks: Where would be a good place to add the explanation? Adding it within the table heading feels like it would make it too clunky to read.

matthew-mcallister: Try putting `<small>` text after/beneath the table.

jvperrin:
@tahabi @redplanks That is what happened (`git log` shows `Author: Benjamin Zhang <bzh@ocf.berkeley.edu>`), but the email hasn't been associated with his GitHub account, so it's not being attributed correctly here.

redplanks:
image

jvperrin:
Nice, thanks for the screenshot

PR #318, opened 2017-09-15
account.vhost_mail: Add CSV import/export feature

matthew-mcallister:
I narrowed the text width on the mail vhost page to make it easier to read.

Also, I had to write 4 lines of JS. It was harrowing.

EDIT: Screenshots

screenshot_2017-09-15_15-41-40
* * *
screenshot_2017-09-15_15-41-18
* * *
screenshot_2017-09-15_15-42-29


harrisonkuo:
Nice js

On Sep 14, 2017 5:38 PM, Matthew McAllister <notifications@github.com> wrote:

I narrowed the text width on the mail vhost page to make it easier to read.

Also, I had to write 4 lines of JS. It was harrowing.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/ocf/ocfweb/pull/318

Commit Summary

  *   account.vhost_mail: Add CSV import/export feature

File Changes

  *   M ocfweb/account/templates/account/vhost_mail/index.html<https://github.com/ocf/ocfweb/pull/318/files#diff-0> (80)
  *   M ocfweb/account/urls.py<https://github.com/ocf/ocfweb/pull/318/files#diff-1> (4)
  *   M ocfweb/account/vhost_mail.py<https://github.com/ocf/ocfweb/pull/318/files#diff-2> (106)
  *   M ocfweb/static/js/pages/account/vhost_mail.js<https://github.com/ocf/ocfweb/pull/318/files#diff-3> (4)

Patch Links:

  *   https://github.com/ocf/ocfweb/pull/318.patch
  *   https://github.com/ocf/ocfweb/pull/318.diff

ocfweb/account/templates/account/vhost_mail/index.html
```diff
@@ -113,12 +117,66 @@
                         <span class="glyphicon glyphicon-plus" aria-hidden="true"></span>
                         Add address
                     </button>
+                    <h4>CSV import/export</h4>
+                    <form class="form-inline js-import-csv-form" enctype="multipart/form-data" action="{% url 'vhost_mail_csv_import' vhost.domain %}" method="post">
+                        <a class="btn btn-default" href="{% url 'vhost_mail_csv_export' vhost.domain %}">
+
```
jvperrin: I'd remove this blank line, it doesn't add much

ocfweb/account/templates/account/vhost_mail/index.html
```diff
@@ -113,12 +117,66 @@
...
+                        You can use CSV spreadsheets to manage your forwarding
+                        addresses. Clicking "Download CSV" will allow you to
+                        download all your forwarding addresses for a particular
+                        domain and the emails they forward to. Clicking "Upload
+                        CSV" will let you pick a CSV file to add new forwarding
+                        rules or update existing ones; no addresses will be
+                        deleted.
+                    </p>
+                    <p>
+                        The CSV format for downloading and uploading is UTF-8
```
jvperrin: I feel like nobody will get the encoding part, I'd just mention the columns and what they contain

ocfweb/account/templates/account/vhost_mail/index.html
```diff
@@ -113,12 +117,66 @@
...
+                        quoted when line breaks or commas are used.
+                    </p>
+                    <p>
+                        Here's what an example of what a valid uploaded CSV
+                        file might look like in a text editor:
+<pre>president,john.doe@berkeley.edu
+officers,john.doe@berkeley.edu jane.doe@berkeley.edu
+committee,"john.doe@berkeley.edu, jane.doe@berkeley.edu"
+members,"john.doe@berkeley.edu,
+jane.doe@berkeley.edu,"</pre>
```
jvperrin: Could we have this indented somehow? Would be nicer than this anyway

ocfweb/static/js/pages/account/vhost_mail.js
```diff
@@ -175,5 +175,9 @@
                 showAddAddressModal($(this).data('domain'), true);
             }.bind(this));
         });
+
```
jvperrin: Wow, a whole newline, that's a hard line of JS to write...

ocfweb/account/templates/account/vhost_mail/index.html
matthew-mcallister: That must have been a mistake, thanks.

ocfweb/static/js/pages/account/vhost_mail.js
matthew-mcallister: OK, maybe it was only 3 lines of JS.

ocfweb/account/templates/account/vhost_mail/index.html
matthew-mcallister: Shall I pass it through a filter?

matthew-mcallister:
Yeah you're right, I'll move the code out of the views and add tests.

matthew-mcallister:
OK, we're at 100% coverage again!

matthew-mcallister:
Is it reasonable to say that `vhost_mail.py` could use some refactoring to further separate the view code from the validation code?

- Having to pass the request along to the validation functions makes error handling in helper functions more awkward. In [`_parse_csv`](https://github.com/matthew-mcallister/ocfweb/blob/88b0f9dd8b55710347514332e5cea0106426f0c7/ocfweb/account/vhost_mail.py#L221-L249), I switched to exception-based handling mainly so `_parse_csv_forward_addrs` could return multiple error messages without having to pass in the request.
- Having to pass the request also makes testing more verbose, since you have to create fake request objects a lot, hence why I id the above workaround in the first place.
- If we eventually try to implement a REST API for mail vhosts, then we could move the validation code to that module and have it shared with the vhosts page. It would also allow sharing the tests.

I wouldn't get started on it right away, but I would do this refactoring myself at least to make it somewhat easier for others to do the API feature if they're interested.

chriskuehl:
Yes, that is definitely fair to say. The existing code was convenient for the original use-case, but it doesn't separate concerns super well and doesn't expand well to include things like an API.

ocfweb/account/templates/account/vhost_mail/index.html
matthew-mcallister: This is better now FYI.

ocfweb/account/vhost_mail.py
```diff
@@ -117,6 +135,137 @@ def vhost_mail_update(request):
...
+
+def _parse_csv(request, domain):
+    """Parse, validate, and return addresses from the file uploaded
+    with the CSV upload button/form."""
+    csv_file = request.FILES.get('csv_file')
+    if not csv_file:
+        _error(request, 'Missing CSV file!')
+
+    addresses = {}
+    with io.TextIOWrapper(csv_file, encoding='utf-8') as f:
```
kpengboy: Let's do something more graceful when someone gives us non-UTF-8 text than crash:


For that matter, since we don't accept special characters anyway, we could as well use ascii for our codec.

ocfweb/account/vhost_mail.py
```diff
@@ -117,6 +135,137 @@ def vhost_mail_update(request):
...
+        reader = csv.reader(f)
+        for i, row in enumerate(reader):
```
kpengboy: I don't know if we should be concerned about somebody trying to upload an inanely large CSV...

ocfweb/account/templates/account/vhost_mail/index.html
```diff
@@ -113,12 +117,61 @@
...
+                    <p>
+                        The CSV format for downloading and uploading has two
+                        columns. The first contains the name part of the
+                        forwarding address (i.e. the part before
+                        <code>@mygroup.berkeley.edu</code>). The second
+                        contains all the receiving emails in a list. When
+                        uploading, the list may be separated by any combination
+                        of commas or whitespace, as long as the column is
+                        quoted when line breaks or commas are used.
+                    </p>
```
kpengboy: I'm concerned that this paragraph might fly over the heads of a large chunk of our target audience.
Perhaps explain what the CSV should look like in a spreadsheet program? (Two columns, left column is ..., right column is ..., addresses in right column can be separated by comma or whitespace, yadda yadda)

matthew-mcallister: That's a good suggestion. I think this image will do the trick:

example_csv

ocfweb/account/vhost_mail.py
matthew-mcallister: That's a good idea, it should display an error message. That's probably more helpful than simply stating "UTF-8 encoded" on the page anyways.

I don't think there's any need to restrict it to ASCII, though. We don't support unicode now, but I don't want to add barriers for anyone who wants to add support later.

For the record, I believe postfix supports UTF-8, so the validation regex and perhaps the database might be the only hindrances to someone wanting to test that out.

ocfweb/account/vhost_mail.py
matthew-mcallister: I originally thought about this briefly and found out that our nginx is configured to put a 10M limit on uploaded files, so no one can upload a truly enormous file. Given that there are much worse ways you could DoS the OCF, I don't think anyone is likely to abuse this anyways. Still, 10M is enough to upload as much as a million lines of CSV.

As far as the website goes, this could stall a web worker for a few minutes as it parses the file and makes a zillion SQL queries. If you did it 100 times, you might be able to DoS the website. More dangerously, if you uploaded enough unique addresses, you could flood the mail vhost table which might break postfix, which is bad.

Do you suggest we do anything about it? If I were serious about it, I would set a line limit on uploaded CSVs and a hard limit in ocflib on the number of addresses for a single vhost, which would prevent flooding the table.

matthew-mcallister: OK, I've addressed the other concerns, and I don't think this needs to stall the PR. If we want to harden our vhost mail database, we can make a ticket for it or an issue in ocflib. This feature is probably fine to release as-is.

ocfweb/account/templates/account/vhost_mail/index.html
kpengboy: I feel like we can entirely omit the technical specification of the CSV format and simply show the image of what the CSV file looks like in a spreadsheet program.

matthew-mcallister: I can make it somewhat shorter, but I don't think the text description should be entirely omitted. What if someone using a screen reader visits the page? Plus, it's important for people uploading CSVs, so the information bears repeating.

matthew-mcallister:
I'll merge this evening if there are no further comments.

PR #319, opened 2017-09-15
account.chpass: Show error when no accounts found

matthew-mcallister:
EDIT: Screenshot (updated):

screenshot_2017-09-15_17-41-20

### DIFF
--- a/ocfweb/account/templates/account/partials/chpass-no-accounts.html
+++ b/ocfweb/account/templates/account/partials/chpass-no-accounts.html
@@ -1,0 +1,12 @@
+<p>Didn't find any accounts associated with your CalNet login.</p>
+<p>
+    If you have an individual account and it isn't showing up here, please
+    <a href="{% url 'doc' 'contact' %}">contact us</a> and mention that your
+    CalNet UID is <tt>{{calnet_uid}}</tt>.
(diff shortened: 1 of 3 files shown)


tahabi:
out of curiosity, how did you test this, by removing the calnetuid from your own ldap entry?

matthew-mcallister:
I added the line `accounts = []` to the view. If the situation were more complex, I think I would try patching the ocflib functions with mock.

ocfweb/account/templates/account/partials/chpass-no-accounts.html
```diff
@@ -0,0 +1,11 @@
+<p>Didn't find any accounts for CalNet UID {{calnet_uid}}.</p>
+<p>
+    If you have an individual account and it isn't showing up here,
+    please <a href="{% url 'doc' 'contact' %}">contact us</a>.
```
chriskuehl: maybe ask them to include the calnet UID in the contact email? we ask that on other pages and it's usually pretty helpful to link accounts

matthew-mcallister: Oh, thanks, you're right, it's square on the same page.

PR #322, opened 2017-09-30
tentatively fix ocfweb->ocflib->ocfweb dependency problem

tahabi:
cut out calls to REGULAR_HOURS, I think this should work for now to stop the ocfweb deploy problems


tahabi:
basically the same format. tests pass and visual inspection of the hours page and graph looks ok

PR #324, opened 2017-09-30
docs: add hypervisor reboot guidelines

kpengboy:


ocfweb/docs/docs/staff/procedures/restarting-services.md
```diff
@@ -19,3 +21,27 @@ Planned restarts of hypervisors should also be announced on this blog, since
 restarting hypervisors can often take several minutes or more.
 
 [status]: http://status.ocf.berkeley.edu
+
+## Rebooting hypervisors
+
+Rebooting hypervisors is a slightly risky business. Hypervisors aren't
+guaranteed to always reboot without problems. Therefore, you shouldn't reboot
```
jvperrin: Honestly neither are VMs, but at least it's easier to debug the VMs remotely.

ocfweb/docs/docs/staff/procedures/restarting-services.md
```diff
@@ -19,3 +21,27 @@ Planned restarts of hypervisors should also be announced on this blog, since
...
+them unless you can physically access the lab in case problems arise.
+Additionally, this risk is the reason (D)SM permission is normally required to
+reboot hypervisors.
+
+So you've gotten the necessary permission and made a post on the status blog
```
jvperrin: Link to the status blog?

ocfweb/docs/docs/staff/procedures/restarting-services.md
```diff
@@ -19,3 +21,27 @@ Planned restarts of hypervisors should also be announced on this blog, since
...
+(if it's a scheduled restart). What now?
+
+First, gracefully shut down all of the VMs. For planned shutdowns of login
+servers (i.e. tsunami, werewolves and corruption), run the `shutdown` command
+well in advance of the shutdown, so users have impending warning. For other VMs,
```
jvperrin: I'd add an example command here, like `sudo shutdown +15 "Rebooting for kernel upgrades"` or something.

ocfweb/docs/docs/staff/procedures/restarting-services.md
```diff
@@ -19,3 +21,27 @@ Planned restarts of hypervisors should also be announced on this blog, since
...
+you can shut them each down via `sudo virsh shutdown`.
+
+Be careful to **always shut down firestorm last**. This is because once
+firestorm is shut down, LDAP logins go offline, and the hypervisors can
+thereafter only be logged into via the root account. It's strongly recommended
+to `sudo -i` before shutting down firestorm.
```
jvperrin: Thanks for mentioning this, it's definitely a good point to make

ocfweb/docs/docs/staff/procedures/restarting-services.md
```diff
@@ -19,3 +21,27 @@ Planned restarts of hypervisors should also be announced on this blog, since
...
+
+Once all of the VMs have been shut down, you can then power off the hypervisors
+via `poweroff`.
```
jvperrin: Or `shutdown` like mentioned above? Why mix the two?

ocfweb/docs/docs/staff/procedures/restarting-services.md
kpengboy: See the text above?

jvperrin: Sure, I do think making this a link wouldn't hurt though, you can just use the same reference as above, and it helps if someone doesn't read the rest of the sections. I do agree they should read those too, but in my opinion having an extra link is pretty harmless.

ocfweb/docs/docs/staff/procedures/restarting-services.md
chriskuehl: Even if logins (passwd) are cached, you won't be able to sudo if kerberos is down (auth won't work).

ocfweb/docs/docs/staff/procedures/restarting-services.md
kpengboy: Eh...well I like `poweroff` because `len("poweroff") < len("shutdown -h now")`

PR #326, opened 2017-10-04
Horizontal line of total desktops for stats/daily-graph

wilswu:
Added a dashed line representing total desktops in the daily graph

### DIFF
--- a/ocfweb/stats/daily_graph.py
+++ b/ocfweb/stats/daily_graph.py
@@ -1,4 +1,3 @@
-
 import urllib.parse
 from datetime import date
 from datetime import datetime
@@ -133,7 +132,7 @@ def get_daily_plot(day):
     ax.set_xlim(start, end)
 
     ax.set_ylim(0, desks_count + 5)
-    ax.set_ylabel('Computers in Use ')
+    ax.set_ylabel('Computers in Use')
 
     ax.set_title('Lab Utilization {:%a %b %d}'.format(day))
     return fig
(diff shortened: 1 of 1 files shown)


chriskuehl:
Code looks great! Can you include a screenshot as well?

wilswu:
stats

matthew-mcallister:
Hm, there should only be 28 computers, so it looks like we need to fix `list_desktops` to hide eruption and the front desk computer and remove one retired desktop.

ocfweb/stats/daily_graph.py
```diff
@@ -85,6 +85,7 @@ def get_daily_plot(day):
     """Return matplotlib plot representing a day's plot."""
     start, end = get_open_close(day)
     profiles = UtilizationProfile.from_hostnames(list_desktops(public_only=True), start, end).values()
+    desks_count = len(list_desktops(public_only=True))
```
jvperrin: Maybe only call `list_desktops` once here and put it into a variable? Every time it is called it makes an external call to LDAP (see [ocflib/lab/stats.py#list_desktops](https://github.com/ocf/ocflib/blob/327dff4a18ba9805aabae8f4c3b4b47f16c42ec5/ocflib/lab/stats.py#L141-L149)), so it seems like a good idea to only make that call once instead of twice.

jvperrin:
@wilswu Also add your OCF email (`wilswu@ocf.berkeley.edu`) to your GitHub account so you get credit for your commits!

PR #328, opened 2017-10-08
Bump pre-commit hooks

jvperrin:
A lot of small reorganization changes as well from the pre-commit hooks


ocfweb/account/vhost.py
```diff
@@ -203,12 +203,14 @@ class VirtualHostForm(Form):
     # website requirements
     website_complete = forms.BooleanField(
         label='The website is already complete and uploaded to the OCF \
```
kpengboy: Wait so what is our string concatenation policy? If we're going to refactor code we might as well decide whether we want to use \ or implicit string concatenation.

jvperrin: I feel like it doesn't matter too much either way, they are both useful in different situations.

PR #330, opened 2017-10-09
Enable SSL verification for celery

jvperrin:
Same as https://github.com/ocf/create/pull/12, https://github.com/ocf/approve/pull/4, and https://github.com/ocf/ircbot/commit/ca9483b04d4b7.

### DIFF
--- a/ocfweb/component/celery.py
+++ b/ocfweb/component/celery.py
@@ -14,14 +14,13 @@ celery_app = Celery(
     broker=settings.CELERY_BROKER,
     backend=settings.CELERY_BACKEND,
 )
-# TODO: use ssl verification
 celery_app.conf.broker_use_ssl = {
-    'ssl_cert_reqs': ssl.CERT_NONE,
+    'ssl_ca_certs': '/etc/ssl/certs/ca-certificates.crt',
+    'ssl_cert_reqs': ssl.CERT_REQUIRED,
 }
-# `redis_backend_use_ssl` is an OCF patch which was proposed upstream:
-# https://github.com/celery/celery/pull/3831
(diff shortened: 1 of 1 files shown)


ocfweb/component/celery.py
```diff
@@ -14,14 +14,13 @@
...
-    'ssl_cert_reqs': ssl.CERT_NONE,
+    'ssl_ca_certs': '/etc/ssl/certs/ca-certificates.crt',
+    'ssl_cert_reqs': ssl.CERT_REQUIRED,
 }
-# `redis_backend_use_ssl` is an OCF patch which was proposed upstream:
-# https://github.com/celery/celery/pull/3831
 celery_app.conf.redis_backend_use_ssl = {
-    'ssl_cert_reqs': ssl.CERT_NONE,
+    'ssl_ca_certs': '/etc/ssl/certs/ca-certificates.crt',
+    'ssl_cert_reqs': ssl.CERT_REQUIRED,
```
chriskuehl: could reduce duplication a bit since these are the same /shrug

```python
celery_app.conf.broker_use_ssl = celery_app.conf.redis_backend_use_ssl = {
    'ssl_ca_certs': '/etc/ssl/certs/ca-certificates.crt',
    'ssl_cert_reqs': ssl.CERT_REQUIRED,
}
```

jvperrin: Good point, that is definitely cleaner

jvperrin: I'll leave this for now, but I do think this would be better, maybe combined with something in ocflib so that this section doesn't have to repeated across 4 repos and changed in each of them.

PR #336, opened 2017-10-31
mention scanners

asaiacai:
A few people e-mail help asking if we have scanners.  Hopefully this will clear that up a bit


matthew-mcallister:
How about making this another paragraph after the second? There, we can optionally list other hardware we have, such as 2 standing desks, a BR reader/writer, and any future acquisitions.

PR #337, opened 2017-11-04
docs: add docs about mailing lists

kpengboy:


ocfweb/docs/docs/staff/mailing-lists.md
```diff
@@ -0,0 +1,53 @@
...
+There are several mailing lists used internally by staff. Append
+`@ocf.berkeley.edu` to the end of each mailing list.
+
+All staffers are automatically added to the following two mailing lists:
+
+ * `staff`: For staff announcements, such as meeting times and events.
+ * `wheel`: For technical discussion among all staff
+
+Staffers can also choose to be further added to the following mailing lists.
+Technical Managers are especially expected to join them:
```
abizer: I'd make it a policy consideration that /root or /admin requires that you be CC'ed on rt@ and root@

Baisang:
IIRC wheel and staff are not on the google admin group page thing, so we should also include that sort of information since that has been a point of confusion in the past

kpengboy:
@Baisang agreed, although documenting how our mailing lists are configured should go on the [infrastructure page about mail](https://www.ocf.berkeley.edu/docs/staff/backend/mail) instead. For that matter, that page could use a lot of love.

ocfweb/docs/docs/staff/mailing-lists.md
```diff
@@ -0,0 +1,53 @@
...
+    * ocflib sends emails whenever an uncaught exception is thrown in ocfweb,
+      create, enforcer, and several other background tasks
+    * Miscellaneous other emails are sent here
+   This mailing list gets a median of ~10 messages every day, although on some
+   days it can get a lot more.
+ * `puppet`: Error messages from puppet runs go here. This list tends to be
+   very noisy.
+ * `mon`: We've set up Rackspace Cloud Monitoring to email us alerts when our
+   important services are inaccessible from outside the OCF network. Alerts
+   from Rackspace get sent to this list.
```
chriskuehl: munin should also be changed to send here imo

jvperrin: Yeah, that would make more sense than mailing root really

ocfweb/docs/docs/staff/mailing-lists.md
```diff
@@ -0,0 +1,53 @@
...
+    * Jenkins sends emails whenever a Jenkins build fails
+    * ocflib sends emails whenever an uncaught exception is thrown in ocfweb,
+      create, enforcer, and several other background tasks
+    * Miscellaneous other emails are sent here
+   This mailing list gets a median of ~10 messages every day, although on some
+   days it can get a lot more.
+ * `puppet`: Error messages from puppet runs go here. This list tends to be
+   very noisy.
+ * `mon`: Monitoring alerts are sent here:
+    * Rackspace Cloud Monitoring email us alerts when our important services are
```
jvperrin: Maybe link the to [monitoring login page](https://intelligence.rackspace.com/login) or something? I don't think there's a link anywhere else that I know of.

jvperrin: Also I think this should be `emails` instead of `email`?

ocfweb/docs/docs/staff/mailing-lists.md
```diff
@@ -0,0 +1,53 @@
...
+      inaccessible from outside the OCF network.
+    * Munin sends mail whenever some munin measurement (e.g. disk usage, RAM
```
jvperrin: Link to Munin from here too? I think it helps make it a little clearer what Munin is if there's a link to the actual site or something like that.

ocfweb/docs/docs/staff/mailing-lists.md
```diff
@@ -0,0 +1,53 @@
...
+      usage, etc.) is outside the normal range
+
+On the administrative side, the `officers` mailing list receives emails related
+to OCF administrivia. Cabinet members are expected to be on this mailing list,
+and any other staffer can audit it as well.
+
+Operations Staff are added to the `opstaff` mailing list.
+
+<!-- TODO: uncomment when this list becomes official
```
jvperrin: Is this not official right now? I mean it doesn't get any emails, but I think it's still the official mailing list.

kpengboy: I don't want to expose this so openly to the public, at least not yet.

ocfweb/docs/docs/staff/mailing-lists.md
kpengboy: Sure, we could add a link, although if I were trying to find the Rackspace login link, I wouldn't think of looking on this page first...

jvperrin: Yeah, that's fair, I suppose we should put a link somewhere else then, I just figured if it was being mentioned that it would be useful to have a link to it as well for more information if wanted.

PR #338, opened 2017-11-05
wordpress web interface setup

asaiacai:
New wordpress web interface download so you don't have to ssh


matthew-mcallister:
Sorry my friend, but we already decided not to do this! See the discussion at https://github.com/ocf/utils/pull/62. Some blockers are the fact that this resets the database password (even it fails) and, worse, that it can fail for reasons outside of our control if `wp-cli` is bugged, which it was for weeks this semester.

PR #344, opened 2017-11-15
Specify the exception type gethostbyaddr() throws.

redplanks:
Flake8 was complaining during testing, so I added the specific [exception](https://docs.python.org/3/library/socket.html#socket.herror).

It also complained about a bare except in vhost_mail.py, but I looked in PyMySQL's source code and the connection object throws several exception types, so a bare except is probably OK there.


chriskuehl:
> so a bare except is probably OK there.

it should be `except Exception:` rather than `except:` so that it doesn't catch `BaseException`s (think `KeyboardInterrupt`, `SystemExit`, `GeneratorExit`, etc.)

PR #345, opened 2017-11-16
Added username recommendations and quicker verification

tliu22:
Also added recommender.py in ocflib

### DIFF
--- a/ocfweb/account/recommender.py
+++ b/ocfweb/account/recommender.py
@@ -6,39 +6,29 @@ from ocflib.account.creation import ValidationWarning
 
 
 def recommend(real_name, n):
-    name_fields = real_name.split()
-    for i in range(len(name_fields)):
-        name_fields[i] = name_fields[i].lower()
+    name_fields = [name.lower() for name in real_name.split()]
 
-    recs = []
-
+    # Can reimplement name_field_abbrevs to only remove vowels or consonants
+    name_field_abbrevs = [[] for i in range(len(name_fields))]
     for i in range(len(name_fields)):
-        rec = ''
-        for j in range(len(name_fields)):
-            if i == j:
-                rec += name_fields[j][0]
-            else:
-                rec += name_fields[j]
+        name_field = name_fields[i]
+        for j in range(1, len(name_field) + 1):
+            name_field_abbrevs[i].append(name_field[:j])
+
+    unvalidated_recs = name_field_abbrevs[0]
+    for i in range(1, len(name_fields)):
+        new_unvalidated_recs = []
+        for name_field_abbrev in name_field_abbrevs[i]:
+            for rec in unvalidated_recs:
+                new_unvalidated_recs.append(rec + name_field_abbrev)
+        unvalidated_recs = new_unvalidated_recs
+
+    validated_recs = []
+    while len(validated_recs) < n and len(unvalidated_recs) > 0:
+        rec = unvalidated_recs.pop(randint(0, len(unvalidated_recs) - 1))
         try:
             validate_username(rec, real_name)
-            if len(recs) >= n:
-                break
-            recs.append(rec)
+            validated_recs.append(rec)
         except (ValidationError, ValidationWarning):
             pass  # Account name wasn't valid, skip this recommendation
-
-    attempts = 0
-    while len(recs) < n and attempts < 20:
-        rec = ''
-        for name_field in name_fields:
-            rand_index = randint(1, len(name_field))
-            rec += name_field[:rand_index]
-        if rec not in recs:
-            try:
-                validate_username(rec, real_name)
-                recs.append(rec)
-            except (ValidationError, ValidationWarning):
-                pass  # Account name wasn't valid, skip this recommendation
-        attempts += 1
-
-    return recs
+    return validated_recs
--- a/ocfweb/account/recommender.py
+++ b/ocfweb/account/recommender.py
@@ -1,0 +1,41 @@
+from random import randint
+
+from ocflib.account.creation import validate_username
+
+
+def recommend(real_name, n):
+    name_fields = real_name.split()
+    for i in range(len(name_fields)):
+        name_fields[i] = name_fields[i].lower()
+
+    recs = []
+
+    for i in range(len(name_fields)):
+        rec = ''
+        for j in range(len(name_fields)):
+            if i == j:
+                rec += name_fields[j][0]
+            else:
+                rec += name_fields[j]
+        try:
+            validate_username(rec, real_name)
+            if len(recs) < n:
+                recs.append(rec)
+        except Exception:
+            pass
+
+    attempts = 0
+    while len(recs) < n and attempts < 20:
+        rec = ''
+        for name_field in name_fields:
+            rand_index = randint(1, len(name_field))
+            rec += name_field[:rand_index]
+        if rec not in recs:
+            try:
+                validate_username(rec, real_name)
+                recs.append(rec)
+            except Exception:
+                pass
+        attempts += 1
+
+    return recs
--- a/ocfweb/account/register.py
+++ b/ocfweb/account/register.py
@@ -7,13 +7,18 @@ from django import forms
 from django.core.urlresolvers import reverse
 from django.forms.forms import NON_FIELD_ERRORS
 from django.http import HttpResponseRedirect
+from django.http import JsonResponse
 from django.shortcuts import render
 from ocflib.account.creation import CREATE_PUBLIC_KEY
 from ocflib.account.creation import encrypt_password
 from ocflib.account.creation import NewAccountRequest
+from ocflib.account.creation import validate_username
+from ocflib.account.creation import ValidationError
+from ocflib.account.creation import ValidationWarning
 from ocflib.account.search import user_attrs_ucb
 from ocflib.account.submission import NewAccountResponse
 
+import ocfweb.account.recommender as recommender
 from ocfweb.account.constants import TESTER_CALNET_UIDS
 from ocfweb.auth import calnet_required
 from ocfweb.component.celery import celery_app
@@ -106,6 +111,38 @@ def request_account(request):
             'title': 'Request an OCF account',
         },
     )
+
+
+def recommend(request):
+    real_name = request.GET.get('real_name', '')
+    if real_name == '':
+        return JsonResponse({'recommendations': []})  # Return empty if no real name
+
+    recommendations = recommender.recommend(real_name, 3)
+    return JsonResponse(
+        {
+            'recommendations': recommendations,
+        },
+    )
+
+
+def validate(request):
+    real_name = request.GET.get('real_name', '')
+    username = request.GET.get('username', '')
+    try:
+        validate_username(username, real_name)
+        return JsonResponse({
+            'is_valid': True,
+            'msg': 'Username is available.',
+        })
+    except (ValidationError, ValidationWarning) as e:
+        msg = str(e)
+        if msg[-1] != '.':
+            msg += '.'
+        return JsonResponse({
+            'is_valid': False,
+            'msg': msg,
+        })
 
 
 def wait_for_account(request):
--- a/ocfweb/account/register.py
+++ b/ocfweb/account/register.py
@@ -119,7 +119,7 @@ def recommend(request):
     if real_name == '':
         return HttpResponseBadRequest()
 
-    recommendations = recommender.recommend(real_name, 3)
+    recommendations = recommender.recommend(real_name, 10)
     return JsonResponse(
         {
             'recommendations': recommendations,
--- a/ocfweb/account/register.py
+++ b/ocfweb/account/register.py
@@ -115,9 +115,9 @@ def request_account(request):
 
 
 def recommend(request):
-    real_name = request.GET.get('real_name', '')
-    if real_name == '':
-        return HttpResponseBadRequest()
+    real_name = request.GET.get('real_name', None)
+    if real_name is None:
+        return HttpResponseBadRequest('No real_name in recommend request')
 
     recommendations = recommender.recommend(real_name, 10)
     return JsonResponse(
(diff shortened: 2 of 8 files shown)


ocfweb/account/recommender.py
```diff
@@ -0,0 +1,24 @@
+from ocflib.account.creation import validate_username
```
matthew-mcallister: This whole file is unused and duplicates https://github.com/ocf/ocflib/pull/102, though there I suggested that the code be permanently moved here. (Oops, I should have left my comments here instead.)

ocfweb/account/register.py
```diff
@@ -21,7 +24,6 @@
 from ocfweb.component.forms import Form
 from ocfweb.component.forms import wrap_validator
 
-
```
matthew-mcallister: This newline is supposed to be there.

ocfweb/account/templates/account/register/index.html
```diff
@@ -118,9 +123,76 @@
...
+                <ul class="list-group list-group-flush" id="recommendations" style="margin: 7px"></ul>
+                <button type="button" class="btn btn-secondary" onclick="recommend()">More</button>
+            </div>
+            <hr class="visible-sm" />
 
             {% include "account/partials/password-requirements-sidebar.html" %}
+
         </div>
     </div>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
```
matthew-mcallister: We already link to jquery in [`base.html`](https://github.com/ocf/ocfweb/blob/master/ocfweb/templates/base.html#L227) ([and we also redundantly vendor it in for some reason](https://github.com/ocf/ocfweb/blob/master/ocfweb/static/js/jquery-1.11.3.min.js)). If you need a different version, we should try and see if we can upgrade it site-wide instead of linking it twice.

ocfweb/account/templates/account/register/index.html
```diff
@@ -118,9 +123,76 @@
...
+                             '<li class="list-group-item list-group-item-success">'+ data[i] +'</li>'
+                         );
+                     }
+                 }
+             });
+        }
+        $(document).ready(function() {
+            recommend();
+        });
+    </script>
```
matthew-mcallister: Can we try to move most/all of this under [`ocfweb/static/js/pages`](https://github.com/ocf/ocfweb/tree/master/ocfweb/static/js/pages/) instead of making it an inline script? Perhaps we could inline only a stub script that sets the `real_name` variable.

ocfweb/account/recommender.py
```diff
@@ -0,0 +1,24 @@
...
+from random import randint
+
+def recommend(first_name, last_name, n):
+    first_name = first_name.lower()
+    last_name = last_name.lower()
+
+    recommendations = []
+    attempts = 0
+    while len(recommendations) < n and attempts < 20:
+        first_rand = randint(1, len(first_name))
```
abizer: same comments as Matt from the ocflib PR

ocfweb/account/register.py
```diff
@@ -107,6 +109,23 @@ def request_account(request):
         },
     )
 
+def recommend(request):
+    real_name = request.GET.get('real_name', '')
+    first_name, last_name = real_name.split()
```
matthew-mcallister: LDAP entries have middle names, so this is just going to crash.

ocfweb/account/register.py
```diff
@@ -107,6 +109,23 @@ def request_account(request):
...
+    rec_lst = recommender.recommend(first_name, last_name, 3)
+    rec_dict = {}
+    for i in range(len(rec_lst)):
+        rec_dict[str(i)] = rec_lst[i] # convert i to string because JS dicts need string keys
```
matthew-mcallister: Does this even need to be a map? You can return just a list in a JSON response.

ocfweb/account/register.py
```diff
@@ -107,6 +109,23 @@ def request_account(request):
...
+    return JsonResponse(rec_dict)
+
+def validate(request):
+    real_name = request.GET.get('real_name', '')
+    username = request.GET.get('username', '')
+    try:
+        validate_username(username, real_name)
+        return JsonResponse({'is_valid': True})
+    except:
+        return JsonResponse({'is_valid': False})
```
matthew-mcallister: Instead of returning a boolean, can we return a string, i.e. one of `"available"`, `"taken"`, or `"invalid"`? That way we can give more specific feedback on the form.

ocfweb/account/register.py
abizer: I don't think you need to do this conversion manually, the JsonResponse function should automatically convert this dict to valid JSON. Also, these variables should probably be named properly

ocfweb/account/templates/account/register/index.html
```diff
@@ -118,9 +123,76 @@
...
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
+    <script>
+        $(document).ready(function() {
+            var finTypingCountdown = 250; // 250 milliseconds
+            var typingTimer;
+            var $input = $('#id_ocf_login_name');
+
+            // On keyup, start countdown
+            $input.keyup(function() {
+                //alert('key UP');
```
abizer: remove these debug alerts

abizer:
You probably also want to delete those extraneous vim crud files

ocfweb/account/register.py
tliu22: screenshot-docs djangoproject com-2017-11-17-09-10-42-640

I wasn't all too sure whether it was okay to make this assumption.

tliu22: As in the keys of the dictionary? I was intending this to essentially mirror the surface functionality of an array or a list. But if the potential security concern I listed above is to be disregarded, then I'll just switch to a list which'll be much cleaner.

ocfweb/account/recommender.py
tliu22: screenshot-supernova ocf berkeley edu-8684-2017-11-17-09-39-37-165
screenshot-supernova ocf berkeley edu-8684-2017-11-17-09-39-12-874
This verify text and green/red highlighting occurs after a specified timeout after user has stopped timing (currently 250ms). 
I originally had the recommendations immediately below the username field (as opposed to the sidebar) but if they were listed vertically, they took up too much space, and I didn't like the way they looked when listed horizontally. But this cosmetic judgment is up to a majority consensus.

ocfweb/account/register.py
matthew-mcallister: It might be better to return a map with a single entry, i.e. `{"recommendations": [...]}`. This is more flexible in the future event that you want to add additional fields to the output.

tliu22: Yea that sounds like a better idea

ocfweb/account/register.py
```diff
@@ -108,6 +113,38 @@ def request_account(request):
...
+
+def validate(request):
+    real_name = request.GET.get('real_name', '')
+    username = request.GET.get('username', '')
+    try:
+        validate_username(username, real_name)
+        return JsonResponse({
+            'is_valid': True,
+            'msg': 'Username is available.',
+        })
```
matthew-mcallister: Nice, I like this.

ocfweb/account/register.py
```diff
@@ -108,6 +113,38 @@ def request_account(request):
...
+    except (ValidationError, ValidationWarning) as e:
+        msg = str(e)
+        if msg[-1] != '.':
+            msg += '.'
```
matthew-mcallister: Nice.

ocfweb/account/recommender.py
```diff
@@ -0,0 +1,41 @@
...
+
+    recs = []
+
+    for i in range(len(name_fields)):
+        rec = ''
+        for j in range(len(name_fields)):
+            if i == j:
+                rec += name_fields[j][0]
+            else:
+                rec += name_fields[j]
```
matthew-mcallister: I feel like we can do better than this by systematically generating names in a clever order. I admittedly haven't thought of a way to iterate through _all_ possible names, but we can get a large number of reasonable names using, say, [`itertools.combinations`](https://docs.python.org/3/library/itertools.html#itertools.combinations).

For example, we could try looping over all combinations of names and taking the first `n` letters of those chosen, where `n = 0, 1, 2, ...`:
```python
# Stop at some low n so we don't loop for a million years
for n in range(3):
    for r in range(1, len(name_fields) + 1)
        for c in itertools.combinations(range(len(name_fields)), r):
            try_names = list(name_fields)
            for i in c:
                try_names[i] = try_names[i][:n]
            <test validity of ''.join(try_names)>
```
We'd probably try the full name first of all, though.

ocfweb/account/recommender.py
```diff
@@ -0,0 +1,41 @@
...
+        for name_field in name_fields:
+            rand_index = randint(1, len(name_field))
+            rec += name_field[:rand_index]
+        if rec not in recs:
+            try:
+                validate_username(rec, real_name)
+                recs.append(rec)
+            except Exception:
+                pass
+        attempts += 1
```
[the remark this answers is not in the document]
tliu22: I don't know if the procedural generator alone will suffice, because in the case of my name, none of the procedural suggestions work ("tonyl", "tliu", and "tonyliu" are all already taken)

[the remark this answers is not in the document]
matthew-mcallister: Well, the problem there is that the first stage doesn't generate enough names as-is, not that we need two methods of generating names.

ocfweb/static/js/pages/account/quick_username_validation.js
```diff
@@ -0,0 +1,41 @@
...
+
+    function validate_username() {
+        $.ajax({
+             type: 'GET',
+             url: 'validate',
+             data: {'username': $input.val(),
+                    'real_name': $('#real_name').html()},
+             success: function(data) {
+                 if(data.is_valid) {
+                     $input.css('border', '5px solid #3c763d');
```
chriskuehl: i think a slightly nicer approach would be to add css classes like `.valid` and `.invalid` and then use `$input.addClass`

in fact, what if you used the bootstrap validation css? https://getbootstrap.com/docs/3.3/css/#forms-control-validation

this css should be on the page, you'd just need to add the right css class names (i think `has-error` and `has-success`)

ocfweb/static/js/pages/account/recommend_username.js
```diff
@@ -0,0 +1,19 @@
+function recommend() {
+    $.ajax({
+         type: 'GET',
+         url: 'recommend',
+         data: {'real_name': $('#real_name').html()},
```
chriskuehl: same thing, can we use `.text()`?

ocfweb/static/js/pages/account/quick_username_validation.js
```diff
@@ -0,0 +1,41 @@
...
+                     $('#valid_username_notif').css('display', 'block');
+                     $('#valid_username_notif').html(data.msg);
```
chriskuehl: can we use `.text(data.msg)` here to avoid potential XSS vulnerabilities?

ocfweb/static/js/pages/account/quick_username_validation.js
```diff
@@ -0,0 +1,41 @@
...
+                     $('#invalid_username_notif').css('display', 'none');
+
+                 } else {
+                     $input.css('border', '5px solid #a94442');
+                     $('#invalid_username_notif').css('display', 'block');
+                     $('#invalid_username_notif').html(data.msg);
```
chriskuehl: can we use `.text(data.msg)` here to avoid potential XSS vulnerabilities?

ocfweb/static/js/pages/account/recommend_username.js
```diff
@@ -0,0 +1,19 @@
...
+         success: function(data) {
+             $('#recommendations').empty();
+             recommendations = data['recommendations'];
+             for (var i in recommendations) {
+                 $('#recommendations').append(
+                     '<li class="list-group-item list-group-item-success">'+ recommendations[i] +'</li>'
```
chriskuehl: can we construct a jquery object here to avoid XSS vulnerabilities?

e.g. (untested, but i think this is right)

```javascript
$('recommendations').append(
    $('<li>')
        .addClass('list-group-item list-group-item-success')
        .text(recommendations[i])
);
```

also -- what if we made these recommendations clickable (click the recommendation to fill in the username box)?

ocfweb/account/register.py
tliu22: Modifying the exception messages sounds like a better idea. Made a separate ocflib PR for that.

ocfweb/account/templates/account/register/index.html
```diff
@@ -118,9 +119,19 @@ <h4>OCF &hearts; Python, Linux, Open Source</h4>
                 </p>
             </div>
 
-            <hr />
+            <div class="well well-sm">
+                <strong>Available Account Names</strong>
+                <ul class="list-group list-group-flush" id="recommendations" style="margin: 7px"></ul>
+                <button type="button" class="btn btn-secondary" onclick="recommend()">More</button>
+            </div>
+            <hr class="visible-sm" />
```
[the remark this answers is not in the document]
tliu22: 32960829-421ef38a-cb7b-11e7-8373-13a136975e17
32960830-425f315c-cb7b-11e7-8d25-5815353efd81

ocfweb/account/recommender.py
tliu22: What do you think about keeping the first letter of each name intact (for "Tony Liu", keep "t" and "l")?
And also how bout only removing vowels (though that might greatly limit the total number of possible recommendations)?

tliu22:
I shifted the recommendations over from the sidebar, so it's easier to notice and doesn't end up at bottom of page for mobile. I also removed the "More" button for recommendations and switched to a scrollable list instead (seems more convenient). 
screenshot_2017-12-02_15-26-29

I'm not too familiar with HTML/JS practices on UI element padding, so a lot of my values are fixed. Please let me know if there's a better way of doing them.

And as for the recommendations function, I randomized which recs to show because I figure if a username doesn't like a particular rec, he probably won't like adjacent recs (adjacent with respect to the unvalidated_recs list in recommender.py).

abizer:
definitely should have squashed this first before merging

PR #346, opened 2017-11-16
add non-warranty clause

Baisang:
github is smart and suggested kpengboy as a requestor

### DIFF
--- a/ocfweb/docs/templates/docs/account_policies.html
+++ b/ocfweb/docs/templates/docs/account_policies.html
@@ -3,7 +3,7 @@
 {% block content %}
     <div class="ocf-content-block">
         <p>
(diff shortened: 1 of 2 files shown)


matthew-mcallister:
> github is smart and suggested kpengboy as a requestor

Reviewer?

IANAL but this is cool by me.

abizer:
the RT is real with baisang

kpengboy:
Also don't forget to update the annotation at the top to mention the amendment date

Baisang:
@kpengboy is that date okay? looking at bod minutes it appears to be the date we confirmed the policy during bod

kpengboy:
@Baisang I'll trust that you got the right date

PR #347, opened 2017-11-16
make tv/hours have a background image

abizer:
this looks nice, IMHO

demo at nyx.ocf.berkeley.edu:8934/tv/hours


ocfweb/static/scss/pages/_tv.scss
```diff
@@ -4,7 +4,14 @@ $tv-hours-small: 5;
 
 .page-tv {
```
kpengboy: Does it work to add these attributes to the `body` rule instead?

abizer: You mean for the TV's base.html `body`? It does IIRC but .page-tv is on a container div immediately inside body which IMHO is slightly cleaner.

matthew-mcallister:
Isn't this originally an SVG? Can we use that instead of the render?

abizer:
I pulled this from RT where it's a PNG. Raymond hasn't uploaded an SVG yet.

raymondnumbergenerator:
i uploaded the .ai file to rt but i think the image won't save properly as an svg

PR #348, opened 2017-11-18
tv/hours: fix background positioning

raymondnumbergenerator:
the top of the background was getting cut off on ocf-tv @abizer 

check:
http://supernova.ocf.berkeley.edu:8513/tv/hours/

image

### DIFF
--- a/ocfweb/static/scss/pages/_tv.scss
+++ b/ocfweb/static/scss/pages/_tv.scss
@@ -2,17 +2,22 @@ $base-font-size: 6vw;
 $tv-hours-large: 12;
 $tv-hours-small: 5;
 
+.page-tv-hours {
+    body {
+        background-image: url('../img/tv/background.png');
+        background-attachment: fixed;
+        background-repeat: no-repeat;
+        background-position: center center;
+        background-size: cover;
+    }
+}
+
 .page-tv {
 
-    background-image: url('../img/tv/background.png');
-    background-attachment: fixed;
-    background-repeat: no-repeat;
-    background-size: cover;
-    height: 100vh;
-
-    .page-tv-hours {
+    .tv-hours {
 
         @extend .text-center;
+        padding-top: 10vh;
 
         font-size: $base-font-size;
 
--- a/ocfweb/tv/templates/tv/base.html
+++ b/ocfweb/tv/templates/tv/base.html
@@ -1,6 +1,6 @@
 <!doctype html>
 
-<html>
+<html class="{{base_css_classes}}">
 
   {% include 'partials/head.html' %}
   <meta http-equiv='refresh' content='600' />
--- a/ocfweb/tv/templates/tv/hours.html
+++ b/ocfweb/tv/templates/tv/hours.html
@@ -2,7 +2,7 @@
 {% load tv_formatting %}
 
 {% block tv-body %}
-<div class="page-tv-hours">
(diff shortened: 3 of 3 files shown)


raymondnumbergenerator:
my bad forgot to do pre-commit

ocfweb/static/scss/pages/_tv.scss
```diff
@@ -9,10 +9,12 @@ $tv-hours-small: 5;
     background-repeat: no-repeat;
     background-size: cover;
     height: 100vh;
+    margin-top: -60px;
```
kpengboy: You'll want to use the variable `$navbar-height` rather than hardcoding its value.

kpengboy:
That would be a better idea actually. We can then just override `padding-top` in the class too.

ocfweb/tv/templates/tv/base.html
```diff
@@ -1,10 +1,10 @@
 <!doctype html>
 
-<html>
+<html class="{{base_css_classes}}">
```
kpengboy: Wait...we shouldn't set the background on the `<html>` element, we should be setting it on the `<body>` element.

raymondnumbergenerator: isn't having a class on the `<html>` consistent with every other page?
setting the background on `<body>` doesn't solve the issue with body having padding at the top because of the navbar

ocfweb/static/scss/pages/_tv.scss
```diff
@@ -2,17 +2,26 @@ $base-font-size: 6vw;
...
 
     background-image: url('../img/tv/background.png');
     background-attachment: fixed;
     background-repeat: no-repeat;
+    background-position: center center;
     background-size: cover;
-    height: 100vh;
+}
+
+.page-tv-body {
```
kpengboy: Here's a slightly easier way to do this, that we can make use of since we're using Sass:

```css
.page-tv-hours {
    /* Stuff that gets applied to everything on this page goes here */
    ...

    body {
        /* Stuff that gets applied to the <body> element and below goes here, e.g.: */
        background-image: url('../img/tv/background.png');
        background-attachment: fixed;
        ...
    }
}
```

Then you won't separately need a `.page-tv-body` class anymore.

See https://github.com/ocf/ocfweb/blob/ea795318a1b19298f03080ff2c6cc9c7d1322b84/ocfweb/static/scss/pages/about/_lab-open-source.scss#L16 for an example.

ocfweb/tv/templates/tv/base.html
kpengboy: > isn't having a class on the `<html>` consistent with every other page?

Yes, but we never set the `background` attribute on an `<html>` element. The proper way to add a background for a page is to put it on the `<body>` element.

> setting the background on `<body>` doesn't solve the issue with body having padding at the top because of the navbar

Whether there is padding or not shouldn't affect the positioning of the background, since padding is space inside the element's box. If the body padding is still an issue, you can manually override it to zero.

kpengboy:
@raymondnumbergenerator just a friendly reminder about this

kpengboy:
@raymondnumbergenerator another reminder about this

raymondnumbergenerator:
requesting to close this, should be obsolete with the lab map now
i'll talk to sahil about merging lab map

PR #353, opened 2017-11-23
Proposal: add temporary net neutrality modal on ocfweb pages

chriskuehl:
This seemed to have general support when I mentioned it on IRC yesterday (technically I proposed changing our desktop home pages, but this also works).

This shows the modal in the screenshot below on all ocfweb pages, only on the first page view. I added a custom title (wanted to include "OCF" in the text, to maybe make it seem less spammy) and turned off GA tracking on the modal. Also the "countdown" theme seemed to be the least confusing, so I went with that one.

Thoughts?

### DIFF
--- a/ocfweb/templates/base.html
+++ b/ocfweb/templates/base.html
@@ -230,5 +230,14 @@
 
         {% include "partials/google-analytics.html" %}
         {% block inline_js %}{% endblock %}
+
+        <!-- TODO: disable after December 14 vote -->
+        <script>
+            var _bftn_options = {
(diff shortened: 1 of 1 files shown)


chriskuehl:
Just to be clear, this also shows on our lab intro page:

kkuehler:
Looks great

chriskuehl:
sounded like everyone was on board yesterday and three +1s on here, gonna go ahead and merge this. someone can revert this if they think it needs more discussion.

PR #355, opened 2017-11-28
hiding regular hours by default

karenytu:
I changed ocf.berkeley.edu/docs/services/lab/ and made the regular weekly hours hidden by default. The reasoning for this change is because Sahil said that people would ignore the actual hours above the regular weekly hours and get confused about lab opening times...

Screenshots of the page on load and after button click are attached.
hours_hidden
hours_shown

### DIFF
--- a/ocfweb/docs/templates/docs/lab.html
+++ b/ocfweb/docs/templates/docs/lab.html
@@ -71,7 +71,9 @@
                 </div>
 
 
-                <h3>Regular Weekly Hours</h3>
+		<h3>Regular Weekly Hours</h3>
+		<button class = "btn btn-primary" id = "weekly_hours_btn" onclick="toggle_weekly_hours()" style="opacity:0.3;">Show</button>
+		<div id = "weekly_hours" style="display:none;">
                 <p>The regular OCF hours are:</p>
                 <div class="row">
                     <div class="col-sm-8">
@@ -86,6 +88,7 @@
                         </table>
                     </div>
                 </div>
+		</div>
 
                 {% if holidays %}
                     <h3 id="holidays">{{semester}} Holidays</h3>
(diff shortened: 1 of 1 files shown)


ocfweb/docs/templates/docs/lab.html
```diff
@@ -71,7 +71,9 @@ <h2 id="hours">Hours</h2>
                 </div>
 
 
-                <h3>Regular Weekly Hours</h3>
+		<h3>Regular Weekly Hours</h3>
+		<button class = "btn btn-primary" id = "weekly_hours_btn" onclick="toggle_weekly_hours()" style="opacity:0.3;">Show</button>
+		<div id = "weekly_hours" style="display:none;">
```
chriskuehl: weird spacing around some of the = signs

ocfweb/docs/templates/docs/lab.html
```diff
@@ -131,4 +134,19 @@ <h4 id="walkthrough">Having trouble finding the entrance?</h4>
...
+		{
+			change.innerHTML = "Hide";
+			$("#weekly_hours").show();
+		}
+		else {
+			change.innerHTML = "Show";
+			$("#weekly_hours").hide();
+		}
+	}
+    </script>
```
chriskuehl: Can we use something from Bootstrap instead of reinventing it ourselves?

https://getbootstrap.com/docs/3.3/javascript/#collapse-example looks like basically what you want. Plus we get nice transitions this way.

chriskuehl: (to clarify, we already have all the bootstrap css/js, so it should be as simple as adding the right classes and hooking things up; the example there should help)

kpengboy:
@karenytu just a friendly reminder about this

kkuehler:
@karenytu bump

PR #356, opened 2017-11-28
Removed white box from "cool services"

edricx:
Removed unnecessary white box around the contents of the "cool services" section in https://www.ocf.berkeley.edu/about/staff.

@sahilhasan

### DIFF
--- a/ocfweb/static/scss/pages/about/_about-staff.scss
+++ b/ocfweb/static/scss/pages/about/_about-staff.scss
@@ -207,8 +207,8 @@ $color-blue3: #29D9C2;
         }
 
         > div > div > div {
-            background-color: $color-white;
-            color: black;
+            /* background-color: $color-white; */
+            color: white;
             text-align: center;
             padding-top: 15px;
             padding-bottom: 5px;
--- a/ocfweb/static/scss/pages/about/_about-staff.scss
+++ b/ocfweb/static/scss/pages/about/_about-staff.scss
@@ -170,7 +170,7 @@ $color-blue3: #29D9C2;
 
     .ocf-services {
         background-color: $color-blue3;
-        color: white;
+        color: black;
         padding-bottom: 30px;
 
         h2 {
@@ -207,7 +207,7 @@ $color-blue3: #29D9C2;
         }
 
         > div > div > div {
-            color: white;
+            color: black;
             text-align: center;
             padding-top: 15px;
             padding-bottom: 5px;
--- a/ocfweb/static/scss/pages/about/_about-staff.scss
+++ b/ocfweb/static/scss/pages/about/_about-staff.scss
@@ -207,7 +207,6 @@ $color-blue3: #29D9C2;
         }
 
         > div > div > div {
-            /* background-color: $color-white; */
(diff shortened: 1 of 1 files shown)


ocfweb/static/scss/pages/about/_about-staff.scss
```diff
@@ -207,8 +207,8 @@ $color-blue3: #29D9C2;
         }
 
         > div > div > div {
-            background-color: $color-white;
-            color: black;
+            /* background-color: $color-white; */
```
matthew-mcallister: Don't comment this out, just delete it.

abizer:
I feel like the white on (green?) is a bit hard to read but otherwise lgtm

edricx:
@abizer I changed the text to white since the header sentence ("OCF staff help provide...") was white originally, but that entire section can be changed to black text. Might actually make more sense because each of the colored sections on the page would then alternate between black and white text.

I was thinking of also getting rid of the white box around OCF Staff Meetings, but since its in some kind of special container the alignment with the left column would look weird. Left as is.

matthew-mcallister:
Please change the text to black. White text on this background is just plain [hard to read](https://leaverou.github.io/contrast-ratio/#white-on-%2329D9C2).

chriskuehl:
...but it looks better as it is now? :(

chriskuehl:
can we just match the white box colors if that's the concern? this was put together in like one hour at 2am one night before a recruiting event, it doesn't surprise me things are a little mismatched.

but i really think it looks better now than with these changes.

edricx:
All the text can be changed to black in the "cool services" section for better readability: https://leaverou.github.io/contrast-ratio/#black-on-%2329D9C2. 

Personally I think the original white box is out of place. I don't see why the three subsections in the box need to be separated as ideas/elements distinct from the rest of the section, so I don't think they should be separated visually either by putting them in a box.

chriskuehl:
@edricx please post screenshots when you make these kinds of changes in PRs

matthew-mcallister:
Simulated:

screenshot_2017-11-28_12-39-21

I fail to see how this is an improvement...

edricx:
I'm not at the OCF and I couldn't get my local version of ocfweb to work. I
linked that contrast testing site which shows what the text would look like.

edricx:
I don't understand why the white box was there in the first place, since
none of the other text or elements on the page are wrapped in a white box
(besides one subsection in the section above, although I'm not touching
that at the moment since the alignment would need to be fixed). The design
of the page is centered around having these different colored rows and
every other row has text directly on the colored background. IMO the white
box feels inconsistent with the rest of the sections. The change to black
text addresses the readability issue in my initial edit. Open to hearing
thoughts/reasons on why the white box would be better though.

matthew-mcallister:
The most obvious criticism I can think of is that graphics don't always look that great on colored backgrounds, especially not when they're really close in color to the background color.

chriskuehl:
Yeah, definitely -1 on this change. This looks worse and is much harder to read.

black text on white >>> white text on dark green

PR #357, opened 2017-11-29
session tracking -> ocfweb

abizer:
same as the puppet PR, this has been working for the last few days but I'm going to let it continue testing for a few more

feedback is appreciated

### DIFF
--- a/ocfweb/api/session_tracking.py
+++ b/ocfweb/api/session_tracking.py
@@ -107,7 +107,7 @@ def _close_sessions(host):
         )
 
 
-@cache()
+@cache(600)
 def _get_desktops():
     """Return IPv4 and 6 address to fqdn mapping for OCF desktops from LDAP."""
 
--- a/ocfweb/api/session_tracking.py
+++ b/ocfweb/api/session_tracking.py
@@ -51,12 +51,7 @@ def log_session(request):
         state = State[body.get('state')]  # triggers KeyError
         user = body.get('user')
 
-        if not user:
-            # could also check LDAP, but that would result in
-            # thousands of extra requests per day to firestorm
-            raise ValueError('No user specified')
-
-        if state is State.cleanup:
+        if state is State.cleanup or not user:
             _close_sessions(host)
         elif state is State.active and _session_exists(host, user):
             _refresh_session(host, user)
--- a/ocfweb/api/session_tracking.py
+++ b/ocfweb/api/session_tracking.py
@@ -50,6 +50,7 @@ def log_session(request):
         user = body.get('user')
 
         if state is State.cleanup or not user:
+            # sessions also get periodically cleaned up: https://git.io/vpwg8
             _close_sessions(host)
         elif state is State.active and _session_exists(host, user):
             _refresh_session(host, user)
@@ -114,6 +115,7 @@ def _get_desktops():
     for e in hosts_by_filter('(type=desktop)'):
         host = e['cn'][0] + '.ocf.berkeley.edu'
         v4 = e['ipHostNumber'][0]
+        v6 = ipv4_to_ipv6(ip_address(v4))
         desktops[v4] = host
-        desktops[ipv4_to_ipv6(v4)] = host
+        desktops[v6] = host
     return desktops
--- a/ocfweb/api/session_tracking.py
+++ b/ocfweb/api/session_tracking.py
@@ -7,6 +7,7 @@ from django.http import HttpResponse
 from django.http import HttpResponseBadRequest
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
+from ipware.ip import get_real_ip
 from ocflib.infra.hosts import hosts_by_filter
 from ocflib.infra.net import is_ocf_ip
 from ocflib.lab.stats import get_connection
@@ -30,7 +31,7 @@ def log_session(request):
     the functionality that used to be in ocf/labstats.
     """
 
-    remote_ip = request.META['REMOTE_ADDR']
+    remote_ip = get_real_ip(request)
 
     if not is_ocf_ip(ip_address(remote_ip)):
         return HttpResponse('Not Authorized', status=401)
@@ -40,8 +41,9 @@ def log_session(request):
 
         host = _get_desktops().get(remote_ip)
         user = body.get('user')
+        state = body.get('state')
 
-        if user == '' or user is None:
+        if not user:
             raise ValueError('Invalid user "{}"'.format(user))
 
         if host is None:
@@ -49,6 +51,8 @@ def log_session(request):
 
         if _session_exists(host, user):
             _refresh_session(host, user)
+        elif state == 'inactive':
+            _close_sessions(host)
         else:
             _new_session(host, user)
 
@@ -102,7 +106,7 @@ def _close_sessions(host):
         )
 
 
-@cache(3600)
+@cache()
 def _get_desktops():
     return {e['ipHostNumber'][0]: e['cn'][0] + '.ocf.berkeley.edu'
             for e in hosts_by_filter('(type=desktop)')}
--- a/ocfweb/api/session_tracking.py
+++ b/ocfweb/api/session_tracking.py
@@ -8,15 +8,13 @@ from django.http import HttpResponse
 from django.http import HttpResponseBadRequest
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
-from ipware.ip import get_real_ip
+from ipware import get_client_ip
 from ocflib.infra.hosts import hosts_by_filter
+from ocflib.infra.net import ipv4_to_ipv6
 from ocflib.infra.net import is_ocf_ip
 from ocflib.lab.stats import get_connection
 
 from ocfweb.caching import cache
-from ocfweb.caching import periodic
-
-CLEANUP_TIMEOUT = 3
 
 State = Enum('State', ['active', 'cleanup'])
 
@@ -36,7 +34,7 @@ def log_session(request):
     Desktops have a cronjob that calls this endpoint: https://git.io/vpIKX
     """
 
-    remote_ip = get_real_ip(request)
+    remote_ip, _ = get_client_ip(request)
 
     if not is_ocf_ip(ip_address(remote_ip)):
         return HttpResponse('Not Authorized', status=401)
@@ -108,24 +106,14 @@ def _close_sessions(host):
         )
 
 
-@periodic(60)
-def _cleanup_sessions():
-    """Periodically clean up sessions that don't die naturally.
-
-    For example, if a desktop crashes or is reset.
-    """
-
-    with get_connection() as c:
-        c.execute(
-            'UPDATE `session` SET `end` = `last_update` WHERE '
-            '`end` IS NULL AND `last_update` < '
-            ' ADDDATE(NOW(), -{} MINUTE)'.format(CLEANUP_TIMEOUT),
-        )
-
-
 @cache()
 def _get_desktops():
-    """Return IP address to fqdn mapping for OCF desktops from LDAP."""
+    """Return IPv4 and 6 address to fqdn mapping for OCF desktops from LDAP."""
 
-    return {e['ipHostNumber'][0]: e['cn'][0] + '.ocf.berkeley.edu'
-            for e in hosts_by_filter('(type=desktop)')}
+    desktops = {}
+    for e in hosts_by_filter('(type=desktop)'):
+        host = e['cn'][0] + '.ocf.berkeley.edu'
+        v4 = e['ipHostNumber'][0]
+        desktops[v4] = host
+        desktops[ipv4_to_ipv6(v4)] = host
+    return desktops
--- a/ocfweb/api/session_tracking.py
+++ b/ocfweb/api/session_tracking.py
@@ -2,7 +2,6 @@ import json
 from functools import partial
 from ipaddress import ip_address
 
-import pymysql
 from django.conf import settings
 from django.http import HttpResponse
 from django.http import HttpResponseBadRequest
@@ -14,14 +13,12 @@ from ocflib.infra.net import is_ocf_ip
 from ocflib.lab.stats import get_connection
 
 from ocfweb.caching import cache
-from ocfweb.caching import periodic
-
-HOST_TIMEOUT = 5  # minutes
 
 get_connection = partial(
     get_connection,
     user=settings.OCFSTATS_USER,
     password=settings.OCFSTATS_PASSWORD,
+    db=settings.OCFSTATS_DB,
 )
 
 
@@ -44,10 +41,10 @@ def log_session(request):
         user = body.get('user', None)
 
         if user is '' or user is None:
-            raise ValueError('Invalid user {}'.format(user))
+            raise ValueError('Invalid user "{}"'.format(user))
 
         if host is None:
-            raise ValueError('Invalid host {}'.format(host))
+            raise ValueError('Invalid host "{}"'.format(host))
 
         if _session_exists(host, user):
             _refresh_session(host, user)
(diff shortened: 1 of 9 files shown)


ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+@csrf_exempt
+def log_session(request):
+    """Primary API endpoint for session tracking.
+
+    Desktops have a cronjob that calls this endpoint, replacing
+    the functionality that used to be in ocf/labstats.
+    """
+
+    if not is_ocf_ip(ip_address(request.META['REMOTE_ADDR'])):
+        return HttpResponse('Not Authorized', status=401)
```
chriskuehl: i don't think this ip is ever going to be accurate? you'll probably see the IP of some load balancer

needs either ocfweb to use proxy protocol or [rpaf](https://github.com/yelp/wsgi-mod-rpaf) or something

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ValueError('Invalid host "{}"'.format(host))
+
+        if _session_exists(host, user):
+            _refresh_session(host, user)
+        else:
+            _new_session(host, user)
+
+        return HttpResponse(status=200)
+
+    except Exception as e:
```
chriskuehl: please be less greedy here. what exceptions are we guarding against and in what code?

as written, this is likely to mask real errors.

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if host is None:
+            raise ValueError('Invalid host "{}"'.format(host))
+
+        if _session_exists(host, user):
+            _refresh_session(host, user)
+        else:
+            _new_session(host, user)
+
+        return HttpResponse(status=200)
```
chriskuehl: i think you want 204

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+    if not is_ocf_ip(ip_address(request.META['REMOTE_ADDR'])):
+        return HttpResponse('Not Authorized', status=401)
+
+    try:
+        body = json.loads(request.body.decode('utf-8'))
+
+        host = _get_desktops().get(request.META['REMOTE_ADDR'], None)
+        user = body.get('user', None)
+
+        if user is '' or user is None:
```
chriskuehl: `==` not `is` to compare strings

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    if not is_ocf_ip(ip_address(request.META['REMOTE_ADDR'])):
+        return HttpResponse('Not Authorized', status=401)
+
+    try:
+        body = json.loads(request.body.decode('utf-8'))
+
+        host = _get_desktops().get(request.META['REMOTE_ADDR'], None)
+        user = body.get('user', None)
```
chriskuehl: default for `dict.get` is `None`

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+        if user is '' or user is None:
+            raise ValueError('Invalid user "{}"'.format(user))
+
+        if host is None:
+            raise ValueError('Invalid host "{}"'.format(host))
+
+        if _session_exists(host, user):
+            _refresh_session(host, user)
+        else:
+            _new_session(host, user)
```
chriskuehl: is there any way to upsert this in raw sql? this is sort of yucky and race-prone as written, and uses multiple txns

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+    """Close old sessions for a particular host."""
+
+    with get_connection() as c:
+        c.execute(
+            'UPDATE `session` SET `end` = NOW(), `last_update` = NOW() '
+            'WHERE `host` = %s AND `end` IS NULL', (host,),
+        )
+
+
+@cache()
```
chriskuehl: can you make this at least somewhat periodic to hopefully reduce confusion when we add new desktops? even refreshing it once every 5 minutes or something should be fine

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,111 @@
...
+def _get_desktops():
+    with ldap_ocf() as c:
+        c.search(OCF_LDAP_HOSTS, '(type=desktop)', attributes=['cn', 'ipHostNumber'])
```
chriskuehl: can you simplify this a bit with `hosts_by_filter`? https://github.com/ocf/ocflib/blob/0e09714ad0b8a10c708ab07c35cc726b3890d26c/ocflib/infra/hosts.py#L18

ocfweb/api/session_tracking.py
jvperrin: Couldn't we set a `X-Forwarded-For` header at the load balancer in nginx and then use that header here? I think we'd also need to have Apache on death forward the IP too though, so that's a bit more awkward.

chriskuehl: yeah, that's what rpaf is :P

jvperrin: For context, the comment I accidentally deleted while trying to cancel my [pending review](https://i.fluffy.cc/t5JgC7f8vmjnmzXWHLbfhFLCnMz4FgH4.png) on my phone was this:

> i don't think this ip is ever going to be accurate? you'll probably see the IP of some load balancer
> 
> needs either ocfweb to use proxy protocol or [rpaf](https://github.com/Yelp/wsgi-mod-rpaf) or something

That'll teach me to review pull requests on my phone... _sigh_

ocfweb/api/session_tracking.py
abizer: Would `ValueError` be appropriate? The two main things I was thinking of were the ones I raise and `JSONDecodeError`, which subclasses `ValueError`.

ocfweb/api/session_tracking.py
abizer: yeah that's a good point, I'll add the things to death for testing

ocfweb/api/session_tracking.py
abizer: The closest thing I can think of would be like `INSERT ... ON DUPLICATE KEY UPDATE` but that doesn't solve this problem, I don't think. Perhaps a solution closer to https://github.com/ocf/ocfweb/blob/0c284bb5759da461c49b60118bc34c57cf581126/ocfweb/account/vhost_mail.py#L381-L391 like having each function use the same connection and committing the transaction at the end of the function?

ocfweb/api/session_tracking.py
chriskuehl: instead of raising ValueError, can you just return a 400 bad request there?

there are lots of things that raise valueerrors

ocfweb/api/session_tracking.py
chriskuehl: okay. sad we can't just do it in sql, but that's okay. the txn approach is better but not critical (ok with shipping this without it, it's not really a "critical" database where races matter much)

ocfweb/api/session_tracking.py
chriskuehl: (my main concern is that you'll catch a legitimate exception and it'll return a bad request, which nobody will ever notice, and then we'll have a real error for a long time with nobody noticing or fixing it)

abizer: that's certainly true, this was just broken for several hours and I didn't notice because the errors were being sent to the client and from there promptly to /dev/null.

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,108 @@
...
+
+def _new_session(host, user):
+    """Register new session in when a user logs into a desktop."""
+
+    _close_sessions(host)
+
+    with get_connection() as c:
+        c.execute(
+            'INSERT INTO `session` (`host`, `user`, `start`, `last_update`) '
+            'VALUES (%s, %s, NOW(), NOW())', (host, user),
```
dkess: Are we not using the django ORM?

chriskuehl: we are not

jvperrin:
Looks like the build is failing on this because ocf/puppet/pull/252 isn't merged (need the dev config for jenkins builds), so we should try and merge that first to make sure this works, but I think something needs to be done to close sessions when the user logs out, because currently I don't see anything that actually does that, and I don't really think we want to make the session closing code stricter, 3 minutes seems like a decent amount of time to use for that.

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,108 @@
...
+@require_POST
+@csrf_exempt
+def log_session(request):
+    """Primary API endpoint for session tracking.
+
+    Desktops have a cronjob that calls this endpoint, replacing
+    the functionality that used to be in ocf/labstats.
+    """
+
+    remote_ip = request.META['REMOTE_ADDR']
```
jvperrin: I think you'll want to use `django-ipware` like we do in the [vhost request form](https://github.com/ocf/ocfweb/blob/6aa0e1ce31497358f153ee29527a603b706a40de/ocfweb/account/vhost.py#L74) to get the user's real IP instead of getting the IP of the proxy host. If that doesn't work, you'd probably want to use [Yelp/wsgi-mod-rpaf](https://github.com/yelp/wsgi-mod-rpaf) or something with some more work to get all the proxy layers between to pass the correct headers. We'd probably want to do this anyway some day so that we can get a good idea of which IPs are actually hitting ocfweb from inside the containers themselves, and having this as a middleware removes the duplication that having to retrieve the correct IP everywhere individually would include.

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,112 @@
...
+
+        host = _get_desktops().get(remote_ip)
+        user = body.get('user')
+        state = body.get('state')
+
+        if not user:
+            raise ValueError('Invalid user "{}"'.format(user))
+
+        if host is None:
+            raise ValueError('Invalid host "{}"'.format(host))
```
kpengboy: Maybe this should log the host IP instead?

jvperrin: Agreed, this should log the IP, logging `None` is not useful.

abizer: will fix

abizer:
the build is failing because the test to ensure nothing errors doesn't like http 405. I'll add some better tests in addition to possibly whitelisting that error code as well.

ocfweb/api/session_tracking.py
```diff
@@ -106,7 +108,24 @@ def _close_sessions(host):
         )
 
 
+@periodic(60)
+def _cleanup_sessions():
+    """Periodically clean up sessions that don't die naturally.
+
+    For example, if a desktop crashes or is reset.
```
jvperrin: I still think this makes a lot more sense as a cronjob, especially since it's more reliable than the ocfweb worker (sometimes we have version mismatches that cause problems) and because this has nothing to do with django, all it does is execute a SQL query. Also, this is already in [ocf_stats/files/stats/bin/close-old-sessions(https://github.com/ocf/puppet/blob/master/modules/ocf_stats/files/stats/bin/close-old-sessions) and is running as a cronjob.

ocfweb/api/session_tracking.py
```diff
@@ -38,21 +43,18 @@ def log_session(request):
...
+        if not host:
             raise ValueError('Invalid host "{}"'.format(host))
 
-        if _session_exists(host, user):
-            _refresh_session(host, user)
-        elif state == 'inactive':
+        state = State[body.get('state')]
+        user = body.get('user')
+
+        if state in (State.inactive, State.cleanup) or not user:
```
jvperrin: Given `inactive` and `cleanup` are just treated the same here, I don't see why 3 states are needed.

ocfweb/api/session_tracking.py
abizer: I kinda disagree, because keeping the logic here keeps almost everything that has to do with session tracking together, rather than one part living as a solitary cronjob way over in puppet, where it is easily forgotten. Also, in this implementation, even if the worker dies, there won't be any problems with the end times being off by however long the worker was dead, because it just sets `end` to `last_update` after the timeout. Seeing a large number of unclosed sessions would also be an indicator that the worker has died, although, that's a side effect.

jvperrin: I do agree having all the logic for this in one place is good, but I still think that using the ocfweb worker as a cron replacement feels wrong, like we're overloading ocfweb to do things that it just doesn't need to be doing, especially since there are other similar cronjobs in ocf_stats for periodically updating MySQL. Either way, we should only have one copy of this, either here, or in puppet, and there are currently 2 copies.

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,131 @@
...
+from enum import Enum
+from functools import partial
+from ipaddress import ip_address
+
+from django.conf import settings
+from django.http import HttpResponse
+from django.http import HttpResponseBadRequest
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST
+from ipware.ip import get_real_ip
```
kpengboy: Does this still need to be updated?

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,131 @@
...
+    Desktops have a cronjob that calls this endpoint: https://git.io/vpIKX
+    """
+
+    remote_ip = get_real_ip(request)
+
+    if not is_ocf_ip(ip_address(remote_ip)):
+        return HttpResponse('Not Authorized', status=401)
+
+    try:
+        host = _get_desktops().get(remote_ip)
```
kpengboy: Wait, how are we dealing with IPv6 addresses?

jvperrin: That's a very good question, I don't think they're handled correctly here.

ocfweb/api/session_tracking.py
abizer: ?

ocfweb/api/session_tracking.py
abizer: I'll remove it from here, it's a smaller change than removing the existing cronjob.

abizer:
@dkess @jvperrin @kpengboy @kkuehler 

plz catch any other bugs I may have missed

ocfweb/api/session_tracking.py
```diff
@@ -0,0 +1,121 @@
...
+
+@require_POST
+@csrf_exempt
+def log_session(request):
+    """Primary API endpoint for session tracking.
+
+    Desktops have a cronjob that calls this endpoint: https://git.io/vpIKX
+    """
+
+    remote_ip, _ = get_client_ip(request)
```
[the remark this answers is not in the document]
jvperrin: The problem with that is that then we'd have to trust the request is legitimate, and anyone could forge lab stats requests from outside the lab without even having access to the desktops.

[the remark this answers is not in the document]
dkess: But that means right now, anyone logged into a desktop can send "trusted" fake data if they know the API endpoint. So we should do some verification step anyways.

I could look into using client certs for this. Passwords could work too.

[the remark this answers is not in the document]
jvperrin: I believe this used to use a client cert and that stopped being used (@chriskuehl), but using a password would probably be good? I don't think we're so worried about people in the lab faking hours though, I mean they could just walk over to a desktop and log in if they want to do that (or if they are staff they could SSH in or edit the mysql lab stats db anyway).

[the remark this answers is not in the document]
abizer: I think the current solution increases the level of annoyance you'd have to go through to fake stats enough that it's not worth trying to do from this end, and easier to just add entries to the db. This is also much more overengineered than it needs to be, lol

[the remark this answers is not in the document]
kpengboy: I also think that the current solution is fine.

PR #339, opened 2017-11-06
add more accurate toner calculation

NickImpicciche:

PR #358, opened 2017-11-30
Toner calculation on webpage

noloerino:
Made changes on #339 to show toner numbers on website

### DIFF
--- a/ocfweb/stats/printing.py
+++ b/ocfweb/stats/printing.py
@@ -71,6 +71,63 @@ def _toner_changes():
     ]
 
 
+# Toner numbers can be significantly noisy, including significant diffs
+# whenever toner gets taken out and put back in whenever there is jam. Because
+# of this it's hard to determine if a new toner is inserted into a printer to
+# reduce this noise we only count diffs that are smaller than a cutoff which
+# empirically seems to be more accurate
+def _toner_used_by_printer(printer, cutoff=.05, since=date(2017, 8, 20)):
+    with stats.get_connection() as cursor:
+        cursor.execute(
+            '''
+            CREATE TEMPORARY TABLE ordered1
+                (PRIMARY KEY (position))
+                AS (
+                    SELECT * FROM (
+                        SELECT
+                            T.*,
+                            @rownum := @rownum + 1 AS position
+                            FROM (
+                                (
+                                    SELECT * FROM printer_toner_public
+                                    WHERE printer = %s AND
+                                    date > %s
+                                    ORDER BY date
+                                ) AS T,
+                                (SELECT @rownum := 0) AS r
+                            )
+                    ) AS x
+                )
+        ''', (printer, since.strftime('%Y-%m-%d')),
+        )
+        cursor.execute('''
+            CREATE TEMPORARY TABLE ordered2
+                (PRIMARY KEY (position))
+                AS (SELECT * FROM ordered1)
+        ''')
+        cursor.execute('''
+            CREATE TEMPORARY TABLE diffs
+            AS (SELECT
+                B.date AS date,
+                B.value/B.max - A.value/A.max as pct_diff
+                FROM
+                    ordered1 as A,
+                    ordered2 as B
+                WHERE
+                    B.position = A.position + 1)
+        ''')
+        cursor.execute(
+            '''
+            SELECT SUM(pct_diff) as toner_used
+            FROM
+            diffs
+            WHERE
+            ABS(pct_diff)<%s
+        ''', (cutoff),
+        )
+        return cursor.fetchone()['toner_used']
+
+
 @periodic(120)
 def _pages_per_day():
     with stats.get_connection() as cursor:
--- a/ocfweb/stats/printing.py
+++ b/ocfweb/stats/printing.py
@@ -66,15 +66,21 @@ def _semester_histogram():
 
 @periodic(3600)
 def _toner_changes():
-    return [(printer, _toner_used_by_printer(printer)) for printer in ACTIVE_PRINTERS]
+    return [
+        (printer, _toner_used_by_printer(printer))
+        for printer in ACTIVE_PRINTERS
+    ]
 
 
-# Toner numbers can be significantly noisy, including significant diffs
-# whenever toner gets taken out and put back in whenever there is jam. Because
-# of this it's hard to determine if a new toner is inserted into a printer to
-# reduce this noise we only count diffs that are smaller than a cutoff which
-# empirically seems to be more accurate
 def _toner_used_by_printer(printer, cutoff=.05, since=date(2017, 8, 20)):
+    """Returns toner changes for a printer since a given date.
+
+    Toner numbers can be significantly noisy, including significant diffs
+    whenever toner gets taken out and put back in whenever there is jam. Because
+    of this it's hard to determine if a new toner is inserted into a printer to
+    reduce this noise we only count diffs that are smaller than a cutoff which
+    empirically seems to be more accurate
+    """
     with stats.get_connection() as cursor:
         cursor.execute(
             '''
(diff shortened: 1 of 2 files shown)


ocfweb/stats/printing.py
```diff
@@ -65,10 +66,7 @@ def _semester_histogram():
 
 @periodic(3600)
 def _toner_changes():
-    return [
-        (printer, _toner_changes_for_printer(printer))
-        for printer in ALL_PRINTERS
-    ]
+    return [(printer, _toner_used_by_printer(printer)) for printer in ACTIVE_PRINTERS]
```
matthew-mcallister: I think this looked better as a multi-liner.

ocfweb/stats/printing.py
```diff
@@ -125,7 +123,10 @@ def _toner_used_by_printer(printer, cutoff=.05, since=date(2017, 8, 20)):
             ABS(pct_diff)<%s
         ''', (cutoff),
         )
-        return float(cursor.fetchone()['toner_used'])
+        result = cursor.fetchone()['toner_used']
+        # The none check is here as a kind of failsafe for if the request is run on a printer
+        # that is no longer in use, although this should no longer happen in the first place
+        return [None if result is None else float(result)]
```
matthew-mcallister: If `result` should never be None, then add an assert statement and get rid of the comment.

ocfweb/stats/printing.py
```diff
@@ -16,6 +16,7 @@
 
 
 ALL_PRINTERS = ('papercut', 'pagefault', 'logjam', 'deforestation')
+ACTIVE_PRINTERS = ('papercut', 'pagefault')
```
chriskuehl: what about reading these from ldap? e.g. like this:

```python
>>> import ocflib.infra.hosts
>>> {host['cn'][0] for host in ocflib.infra.hosts.hosts_by_filter('(type=printer)')}
{'papercut', 'pagefault'}
```

unfortunately we can't do this for old printers (they're not in ldap anymore), but it could work for current printers

(low priority since we'd still need to update ALL_PRINTERS anyway when we get a new printer)

ocfweb/stats/printing.py
```diff
@@ -65,10 +66,67 @@ def _semester_histogram():
...
-    ]
+    return [(printer, _toner_used_by_printer(printer)) for printer in ACTIVE_PRINTERS]
+
+
+# Toner numbers can be significantly noisy, including significant diffs
+# whenever toner gets taken out and put back in whenever there is jam. Because
+# of this it's hard to determine if a new toner is inserted into a printer to
+# reduce this noise we only count diffs that are smaller than a cutoff which
+# empirically seems to be more accurate
+def _toner_used_by_printer(printer, cutoff=.05, since=date(2017, 8, 20)):
```
chriskuehl: can you put this comment in a docstring so that it will show up in pydoc? (some IDEs and similar can also show this i think)

e.g.

```python
def _toner_used_by_printer(...):
    """Return toner changes for a printer.

    Toner numbers can be significantly noisy...
    """
```

ocfweb/stats/printing.py
```diff
@@ -65,10 +66,67 @@ def _semester_histogram():
...
+                    B.position = A.position + 1)
+        ''')
+        cursor.execute(
+            '''
+            SELECT SUM(pct_diff) as toner_used
+            FROM
+            diffs
+            WHERE
+            ABS(pct_diff)<%s
+        ''', (cutoff),
```
chriskuehl: `(cutoff)` is sort of misleading, that's not a tuple. the python syntax around tuples sucks:

```python
>>> type(())
<class 'tuple'>
>>> type(('a'))
<class 'str'>
>>> type(('a', 'b'))
<class 'tuple'>
```

This often causes confusion in cases like this where people try to iterate it, e.g.

```python
>>> for x in ('hello'):
...     print(x)
... 
h
e
l
l
o
```

i'd either change to just `cutoff` without parentheses, or `(cutoff,)` if you want it to be a tuple

(guessing the mysql api allows it to be a string if this is working? i didn't know that)

ocfweb/stats/printing.py
```diff
@@ -65,10 +66,67 @@ def _semester_histogram():
...
+        )
+        result = cursor.fetchone()['toner_used']
+        # The none check is here as a kind of failsafe for if the request is run on a printer
+        # that is no longer in use, although this should no longer happen in the first place
+        return [None if result is None else float(result)]
```
chriskuehl: i'm not really a fan of this kind of error handling. it seems like a real error to run this function on a no-longer-used printer, right? if so, i'd expect the function to error when i call it.

generally i'd prefer to fail loud so that we're alerted that something is broken and can fix things. an easy way to do that without having a yucky error message (like `TypeError: float() argument must be a string or a number, not 'NoneType'`) is to just add an assertion before it:

```python
assert result is not None, 'No data exists for printer: {}'.format(printer)
return float(result)
```

ocfweb/stats/printing.py
```diff
@@ -95,6 +153,7 @@ def _pages_per_day():
     return pages_printed
 
 
+# Deprecated in favor of _toner_used_by_printer by virtue of giving meaningless data
 def _toner_changes_for_printer(printer):
```
chriskuehl: can we delete this function if nothing is using it?

ocfweb/stats/printing.py
noloerino: My bad, screenshot here: https://imgur.com/a/IeWHB

chriskuehl:
so, the purpose of having that entire table was so we could see things like "when did we last make a toner change" or "have we been changing toner more this semester", and we're kinda losing that data.

i wonder if this change should add an additional table rather than replace the existing one

noloerino:
Question, since this is my first time committing: how do I implement suggested changes? Do I just manually make the changes on my copy and then push them, or is there something I can pull from?

I think the reason Sahil/Nick wanted the table to be changed is that the old table was essentially meaningless because jamming/unjamming the printer would cause weird spikes in the data completely unrelated to actual toner cartridge changes, so I think they just wanted a count of how much toner has been used this semester. I'm not completely sure on that though.

ocfweb/stats/printing.py
noloerino: where would this error be handled?

matthew-mcallister:
@noloerino If you push changes to your branch, they'll show up here.

ocfweb/stats/printing.py
chriskuehl: you don't need to handle it if we never expect it to happen. "programming bugs" like these can't really be handled in a sane way except by degrading the experience (e.g. 500ing the page or not showing the affected component) and notifying staff to fix it.

if it *were* to start happening, that would mean there's a bug in ocfweb and we'd automatically start getting emailed about it and know to fix it.

ocfweb/stats/printing.py
matthew-mcallister: For future reference, you can upload pictures to GitHub when you make a comment.

matthew-mcallister:
I think the "Toner changes" heading is going to be out-of-date if we actually show the cumulative use instead of a table of changes.

ocfweb/stats/printing.py
noloerino: Is it ok to leave the result in a singleton list like this? I left it like that since the HTML on the stats page loops through a list of toner changes to generate rows, and even though I could've changed that, I didn't want to in case we have the table display toner changes for multiple semesters/different time periods in the future.

noloerino:
What should the heading be then?

ocfweb/stats/printing.py
chriskuehl: personally i'd drop the list, usually when we try to be defensive and leave code around in case somebody wants to use it later, it just ends up being technical debt.

if people want to find the old code, they can also grab it from the git history.

matthew-mcallister:
> What should the heading be then?

It's not really important as long as it doesn't look like a relic. Something like "Toner usage" would be appropriate and is more general. Also consider what @chriskuehl said: we may want to revive the existing table if we can find a way for it to be more useful.

noloerino:
How about this?
screen shot 2017-11-30 at 14 51 14

PR #359, opened 2017-12-06
Fix JS path

tliu22:
- Fix JS Path
- Clean up username feedback code

### DIFF
--- a/ocfweb/account/templates/account/register/index.html
+++ b/ocfweb/account/templates/account/register/index.html
@@ -58,8 +58,7 @@
                         <div class="col-sm-7">
                             <p>Your account name must consist of between three and 16 lowercase letters (no spaces, numbers, underscores, or other symbols), and it <strong><strong>MUST</strong> be based on your real name or initials.</strong></p>
                             {{form.ocf_login_name|bootstrap}}
-                            <div class="alert alert-success" role="alert" id="valid_username_notif" style="display: none; padding: 7px"></div>
-                            <div class="alert alert-danger" role="alert" id="invalid_username_notif" style="display: none; padding: 7px"></div>
+                            <div class="alert" role="alert" id="username_feedback" style="display: none; padding: 7px"></div>
                         </div>
                         <div class="col-sm-5">
                             <div class="well well-sm">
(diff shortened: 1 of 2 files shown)


ocfweb/account/templates/account/register/index.html
```diff
@@ -58,8 +58,7 @@
                         <div class="col-sm-7">
                             <p>Your account name must consist of between three and 16 lowercase letters (no spaces, numbers, underscores, or other symbols), and it <strong><strong>MUST</strong> be based on your real name or initials.</strong></p>
                             {{form.ocf_login_name|bootstrap}}
-                            <div class="alert alert-success" role="alert" id="valid_username_notif" style="display: none; padding: 7px"></div>
-                            <div class="alert alert-danger" role="alert" id="invalid_username_notif" style="display: none; padding: 7px"></div>
+                            <div class="alert" role="alert" id="username_feedback" style="display: none; padding: 7px"></div>
```
jvperrin: It would be better if this CSS was in a stylesheet instead of inline here, but this is fine for now.

ocfweb/account/templates/account/register/index.html
```diff
@@ -139,6 +138,6 @@ <h4>OCF &hearts; Python, Linux, Open Source</h4>
     </div>
 {% endblock %}
 {% block inline_js %}
-        <script src="/static/js/pages/account/quick_username_validation.js"></script>
-        <script src="/static/js/pages/account/recommend_username.js"></script>
+        <script src="{% static 'js/pages/account/quick_username_validation.js' %}"></script>
+        <script src="{% static 'js/pages/account/recommend_username.js' %}"></script>
```
chriskuehl: i wonder if we should combine these two scripts into one

the size isn't an issue, but multiple requests can suck on bad connections (e.g. phones, where people often make new accounts)

tliu22: Sure, should I just throw under a new register.js?

chriskuehl: sounds reasonable to me!

PR #360, opened 2017-12-06
Move styling to _register.scss

tliu22:
- Move styling to _register.scss
- Change variable names to be more consistent

### DIFF
--- a/ocfweb/static/scss/pages/account/_register.scss
+++ b/ocfweb/static/scss/pages/account/_register.scss
@@ -4,9 +4,8 @@
     }
 }
 
-.username-feedback {
-    display: none;
-    padding: 7px;
+.recommendation {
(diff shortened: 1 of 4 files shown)


ocfweb/static/scss/pages/account/_register.scss
```diff
@@ -3,3 +3,18 @@
...
+.recommendations-list {
+    margin: 7px;
+    overflow-y: scroll;
+    max-height: 150px;
+}
+
+.username-feedback {
+    display: none;
+    padding: 7px;
+}
```
chriskuehl: this stuff all needs to go in the `.page-register` block above, otherwise it's still part of the global css

(we only ship one css file currently, the way we namespace is based on the `.page-blah` rules

PR #361, opened 2017-12-06
Consolidate register scripts into single file

tliu22:
-Consolidate quick_username_validation.js and recommend_username.js into register.js


ocfweb/static/js/pages/account/register.js
```diff
@@ -38,3 +39,30 @@ $(document).ready(function() {
...
+                         text: recommendation,
+                     })
+                 );
+             }
+         }
+     });
+}
+$(document).ready(function() {
+    recommend();
+});
```
jvperrin: Instead of having two `$(document).ready` blocks, I'd combine the two, so that this function call is in the block above. Otherwise, I think it could get confusing if someone is reading this file, since they won't expect there to be two different places that functions are called from.

jvperrin:
@tliu22 I think you pushed to the wrong remote? Your new commit is on the testing branch on this repo, but this pull request is for merging this branch from your fork, not from this repo.

tliu22:
Thanks for deleting that branch. Accidentally set origin to ocf/ocfweb instead of tliu22/ocfweb.

PR #365, opened 2018-01-26
Remove request reason field from vhost form

kpengboy:
The request reason field was originally added to the form back when anyone could fill it out. Its intended purpose was to discourage frivolous requests. Today, the vhost form is only accessible by those who are eligible for vhosting, and this field serves no useful purpose.

### DIFF
--- a/ocfweb/account/templates/account/vhost/index.html
+++ b/ocfweb/account/templates/account/vhost/index.html
@@ -58,7 +58,6 @@
 
                 {{form.requested_own_domain|bootstrap}}
                 {{form.requested_subdomain|bootstrap}}
-                {{form.requested_why|bootstrap}}
 
                 <p><strong>Website requirements:</strong></p>
                 <p>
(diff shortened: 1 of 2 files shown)


chriskuehl:
It was not added to "prevent frivolous requests", it was added because IST asked us to collect a reason why they couldn't just use their userdir.

chriskuehl:
Also, the form has *always* only been accessible by eligible people (only groups, originally).

kpengboy:
> It was not added to "prevent frivolous requests", it was added because IST asked us to collect a reason why they couldn't just use their userdir.

I remembered Dara saying it was to discourage frivolous requests.

> Also, the form has always only been accessible by eligible people (only groups, originally).

I stand corrected.

chriskuehl:
It's probably a little of both. IST definitely argued for userdirs and I recall this being part of why Dara asked me to add that field, but who knows... searching RT is difficult :P. In any case, I don't think it matters -- I think it is fine to delete The reason for all of them is effectively the same.

PR #366, opened 2018-01-30
Lab Reservation Infrastructure

EnnFour:
This is the pull request for the LRI. Follow-up to ocf.io/rt/6938


jvperrin:
This failed because the `staff_hours.yaml` file was malformed (started with a comma). I'll re-run the tests.

jvperrin:
Nice, it passed this time. I swear, people cannot edit yaml files...

ocfweb/reservations/mail_templates/officer_notification.jinja
```diff
@@ -0,0 +1,9 @@
+Hello all,
+
+The OCF has received a lab reservation request!
+
+The request was issued by {{request.real_name}} ({{request.contact_email}}) on behalf of the student group {{request.student_group}}.
+They wish to use our lab from {{request.starttime}} until {{request.endtime}} on {{request.date}}.
+Their stated reason is: {{request.reason}}
+
+Would a staffer kindly create an RT ticket and place it in the bod queue so that a vote can be held at the next meeting?
```
chriskuehl: since these are plain-text email templates, please wrap lines to ~72 chars

ocfweb/reservations/mail_templates/user_notification.jinja
```diff
@@ -0,0 +1,8 @@
+Hello {{request.real_name}},
+
+The OCF has received your request to reserve the lab on {{request.date}} from {{request.starttime}} to {{request.endtime}} and is pending staff approval.
+Your request is currently being processed. Should we encounter any difficulty or require any additional information, we will send you an email.
+Once we've made a decision, we'll send you an email to let you know.
+
+If, for whatever reason, you need to contact us about this reservation request, please don't hesitate to send an email to awelty@ocf.berkeley.edu.
+Thanks for flying OCF!
```
chriskuehl: ditto

ocfweb/reservations/mail_templates/user_notification.jinja
```diff
@@ -0,0 +1,8 @@
+Hello {{request.real_name}},
+
+The OCF has received your request to reserve the lab on {{request.date}} from {{request.starttime}} to {{request.endtime}} and is pending staff approval.
```
chriskuehl: and *it* is pending approval?

ocfweb/reservations/mail_templates/user_notification.jinja
```diff
@@ -0,0 +1,8 @@
...
+Your request is currently being processed. Should we encounter any difficulty or require any additional information, we will send you an email.
+Once we've made a decision, we'll send you an email to let you know.
+
+If, for whatever reason, you need to contact us about this reservation request, please don't hesitate to send an email to awelty@ocf.berkeley.edu.
```
chriskuehl: awelty?

ocfweb/reservations/reserve.py
```diff
@@ -0,0 +1,214 @@
...
+from django.shortcuts import render
+from jinja2 import Environment
+from jinja2 import PackageLoader
+from ocflib.account.search import user_attrs_ucb
+
+from ocfweb.auth import calnet_required
+from ocfweb.component.forms import wrap_validator
+
+SENDMAIL_PATH = '/usr/sbin/sendmail'
+MAIL_FROM = 'Open Computing Facility <help@ocf.berkeley.edu>'
```
chriskuehl: we shouldn't be redefining these, `ocflib.misc.mail` knows how to send email already

ocfweb/reservations/reserve.py
```diff
@@ -0,0 +1,214 @@
...
+
+            return render(
+                request,
+                'reservations/pending.html',
+            )
+
+    else:
+        form = RequestForm()
+
+    return render(
```
chriskuehl: indent this for consistency with the above? or deindent the thing above?

personally i would write this all as one big if-statement

```python
if not user_attrs_ucb(calnet_uid):
    ...
elif request.method == 'POST':
    ...
else: 
    form = RequestForm()
    return render(...)
```

imo this makes the program flow more obvious, as you don't need to check the other branches to see if they actually do return or not

ocfweb/reservations/reserve.py
```diff
@@ -0,0 +1,214 @@
...
+        {
+            'form': form,
+            'status': status,
+            'title': 'Request the OCF Lab Space',
+        },
+    )
+
+
+def reservation_requested(request):
+    return render(request, 'reservations/pending.html', {'title': 'Account request successful'})
```
chriskuehl: reservation request?

ocfweb/reservations/reserve.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def send_mail(to, subject, body, sender=MAIL_FROM):
+    """Send a plain-text mail message.
+    `body` should be a string with newlines, wrapped at about 80 characters."""
```
chriskuehl: this function already exists in `ocflib.misc.mail`, we shouldn't duplicate it

ocfweb/reservations/reserve.py
```diff
@@ -0,0 +1,214 @@
...
+        stdin=subprocess.PIPE,
+    )
+    p.communicate(msg.as_string().encode('utf8'))
+
+
+def send_request_to_officers(request):
+    body = JINJA_MAIL_ENV.get_template(
+        'reservations/mail_templates/officer_notification.jinja',
+    ).render(request=request)
+    send_mail('welty025@berkeley.edu', 'New Lab Reservation Request', body)
```
chriskuehl: fix email

ocfweb/reservations/mail_templates/officer_notification.jinja
chriskuehl: why not just send directly to an RT queue like we do for vhost requests?

ocfweb/reservations/templates/mail_templates/user_notification.jinja
```diff
@@ -0,0 +1,8 @@
+Hello {{request.true_name}},
+
+The OCF has received your request to reserve the lab on {{request.date}} from {{request.starttime}} to {{request.endtime}} and is pending staff approval.
+Your request is currently being processed. Should we encounter any difficulty or require any additional information, we will send you an email.
+Once we've made a decision, we'll send you an email to let you know.
+
+If, for whatever reason, you need to contact us about this reservation request, please don't hesitate to send an email to awelty@ocf.berkeley.edu.
+Thanks for flying OCF!
```
chriskuehl: wrap lines

also -- duplicate?

ocfweb/reservations/templates/reservations/cant-find-in-LDAP.html
```diff
@@ -0,0 +1,23 @@
+{% extends "base.html" %}
```
chriskuehl: don't capitalize in file names

ocfweb/reservations/templates/reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+        <div class="col-md-8 ocf-content-block">
+
+            <form method="post" action="{% url 'request_reservation' %}">
+                {% csrf_token %}
+                <p>
+                    This form is <strong>only</strong> for student groups desiring to
+                    use the OCF's lab space. If you want to create an individual account
+                    please visit <a href="{% url 'register' %}">this link</a>. If you
+                    want to request an account for a student group, please visit
+                    a staff member in the OCF lab during
```
chriskuehl: fix copy, this refers to creating accounts

ocfweb/reservations/templates/reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+                         In addition, I agree not to hold the University
+                        of California, the Associated Students of the University of
+                        California, or the Open Computing Facility (OCF)
+                        responsible for lost files or other misfortunes which may
+                        result from my use of the OCF's facilities.
+
+                        Additionally, I certify that I am eligible for an OCF
+                        account according to the
+                        <a href="{% url 'doc' 'membership/eligibility' %}">OCF account eligibility policy</a>.
+                    </strong></p>
```
chriskuehl: this entire template seems to refer to account creation?

ocfweb/reservations/urls.py
```diff
@@ -0,0 +1,15 @@
+from django.conf.urls import url
```
chriskuehl: i would vote for calling this stuff `lab_reservations` instead of just `reservations` (and updating the routes accordingly)

ocfweb/reservations/reserve.py
```diff
@@ -0,0 +1,214 @@
...
+    :param real_name:
+    :param contact_email:
+    :param student_group:
+    :param reason:
+    :param date:
+    :param starttime:
+    :param endtime:
+    :param handle_warnings: one of WARNINGS_WARN, WARNINGS_SUBMIT
+        WARNINGS_WARN: don't create request, return warnings
+        WARNINGS_SUBMIT: submit for staff approval
```
chriskuehl: where are you actually using this `handle_warnings` stuff?

ocfweb/reservations/mail_templates/officer_notification.jinja
EnnFour: Fixed

EnnFour: At present, we desire to send an email to officers. This feature will happen in the future, however.

ocfweb/reservations/mail_templates/user_notification.jinja
EnnFour: I used personal emails to test so as not to spam. These should all be fixed now.

ocfweb/reservations/reserve.py
EnnFour: Fixed

ocfweb/reservations/reserve.py
EnnFour: Nowhere, and removed. Thanks for pointing this out.

ocfweb/reservations/reserve.py
EnnFour: I believe the indentation is correct. I can combine the `if not user_attrs_ucb(calnet_uid):` and the `if request.method == 'POST':` into an if/elif/else block if you still desire.

ocfweb/reservations/templates/reservations/index.html
EnnFour: I intended the wording here to be exactly as is.

ocfweb/reservations/templates/reservations/index.html
EnnFour: If a student wants to request a reservation with us, I want them to be able to make an OCF account. I feel this is reasonable and the wording here was exactly as I intended as well, despite the parallels it has to account creation.

ocfweb/reservations/urls.py
EnnFour: I've done this -- good call

ocfweb/reservations/reserve.py
chriskuehl: it is "correct" in that the flow is correct, but weird because half of your one logical branch is indented differently to the other half

ocfweb/reservations/templates/reservations/index.html
chriskuehl: honestly don't understand what you're saying -- it seems like you deleted all this anyway?

EnnFour: Could you clarify? I don't see anything deleted in this file

ocfweb/reservations/templates/reservations/index.html
chriskuehl: looks like it was just renamed. i'll re-open my comments on the new file.

ocfweb/reservations/templates/reservations/index.html
chriskuehl: looks like it was just renamed. i'll re-open my comments on the new file.

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+
+                <p>{{form.date|bootstrap}}</p>
+                <p>{{form.starttime|bootstrap}}</p>
+                <p>{{form.endtime|bootstrap}}</p>
+
+
+                <div class="well">
+                    <p><strong>
+                        Please read our lab reservation policy
+                        <a href=https://www.ocf.berkeley.edu/docs/services/lab/lab-reservation-policy/>
```
chriskuehl: missing quotes on the href attribute

also, please don't do absolute links like this, use the `{% url %}` template helpers -- this helps ensure all links are up-to-date and that we don't have any broken links (they would fail at test time and never make it into production)

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+                         here</a>; you are responsible for
+                        knowing and abiding by it.
+                    </strong></p>
+
+                    <p><strong>
+                        I have read, understood, and agreed to follow the policy linked above.
+                         In addition, I agree not to hold the University
```
chriskuehl: indentation

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+                        of California, the Associated Students of the University of
+                        California, or the Open Computing Facility (OCF)
+                        responsible for lost files or other misfortunes which may
+                        result from my use of the OCF's facilities.
+
+                        Additionally, I certify that I am eligible for an OCF
+                        account according to the
+                        <a href="{% url 'doc' 'membership/eligibility' %}">OCF account eligibility policy</a>.
```
chriskuehl: why not just require a group account if that's the idea?

this entire template is pretty confusingly written

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+                    <a class="btn btn-default" href="{% url 'calnet_logout' %}">Cancel</a>
+                </p>
+            </form>
+        </div>
+
+        <!-- This stuff is the sidebar. -->
+        <div class="col-md-4">
+            <hr class="visible-sm" />
+            <div class="welcome-penguin">
+                <h2>Welcome to the OCF!</h2>
```
chriskuehl: ? fix copy-pasted sidebar, this is not for new accounts, right?

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,92 @@
...
+
+            <form method="post" action="{% url 'request_reservation' %}">
+                {% csrf_token %}
+                <p>
+                    This form is <strong>only</strong> for student groups desiring to
+                    use the OCF's lab space. If you want to create an individual account
+                    please visit <a href="{% url 'register' %}">this link</a>. If you
+                    want to request an account for a student group, please visit
+                    a staff member in the OCF lab during
+                    <a href="{% url 'staff-hours' %}">staff hours</a>.
```
chriskuehl: still think this is bizarre, why are you providing links to a bunch of totally unrelated forms that we already link to...?

ocfweb/lab_reservations/templates/lab_reservations/index.html
abizer: Apparently, we want to have this page open to the public and not behind a group account login, because groups without OCF group accounts should be allowed to reserve the lab. I don't think making a group account is a signficant hurdle for the kinds of groups we'd approve of using the lab, however.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+
+def reservation_requested(request):
+    return render(request, 'lab_reservations/pending.html', {'title': 'Reservation request successful'})
+
+
+def send_request_to_officers(request):
+    body = JINJA_MAIL_ENV.get_template(
+        'lab_reservations/mail_templates/officer_notification.jinja',
+    ).render(request=request)
+    send_mail('officers@berkeley.edu', 'New Lab Reservation Request', body)
```
abizer: I still think it's a better idea to just have this get sent to bod@ immediately with the requestor set appropriately, without going through officers@ first, and to integrate that "feature" in right now as opposed to waiting for someone to do it in the future.

chriskuehl: totally agreed -- although keep in mind i think bod@ has a special rule where requestors aren't emailed about replies (not sure if it's still there or not, but we tried implementing it after some embarrassing accidental reply to a bod ticket). might still be okay for this use-case (e.g. it can be moved to `help` after approval or whatever), just wanted to mention.

abizer: If that's the case, it may be a better long-term option to repurpose the lending queue as the requests queue. Aside from borrowing stuff and asking to reserve the lab, I imagine in the future we'll also want to have an HPC request form, and having a single requests queue would be a good way to manage all of that together.

ocfweb/lab_reservations/templates/lab_reservations/index.html
EnnFour: After further discussion with Sahil, we've decided to follow this recommendation and remove the information I provided at the top of the page.

ocfweb/lab_reservations/templates/lab_reservations/index.html
EnnFour: Fixed

ocfweb/lab_reservations/reserve.py
EnnFour: My most recent commit tackles this issue. Instead of an email to officers, instead a ticket is created in the bod queue.

EnnFour:
@chriskuehl @abizer I feel like all the main issues you guys had with my code are fixed. Could you both please review this updated code and comment any concerns you have?

ocfweb/lab_reservations/mail_templates/officer_notification.jinja
```diff
@@ -0,0 +1,9 @@
+Hello all,
+
+The OCF has received a lab reservation request!
+
+The request was issued by {{request.real_name}}
+({{request.contact_email}}) on behalf of the student group
+{{request.student_group}}. They wish to use our lab from {{request.starttime}}
+until {{request.endtime}} on {{request.date}}. Their stated reason is:
+{{request.reason}}
```
chriskuehl: (not blocking)

This is sort of nit-picky, but I think it's usually easier for people to parse form emails like this when the fields are just put in a list (similar to what we do for vhost requests). For example:


> Lab Reservation Request:
> - Submitter: *Chris Kuehl*
> - Email: *ckuehl@ocf.berkeley.edu*
> - Student Group: *ckuehl appreciation club*
> - Time: *6pm to 9pm*
> - Date: *2018-02-18*
>
> Request Reason:
> *details*

I don't really see the point of wrapping a machine-generated email in superfluous English that just needs to be parsed out.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+    :param date:
+    :param starttime:
+    :param endtime:
+    """
+
+    def to_dict(self):
+        return {
+            field: getattr(self, field)
+            for field in self._fields
+        }
```
chriskuehl: There's a built-in `_asdict()` method on namedtuples you can use:

```python
>>> import collections
>>> Point = collections.namedtuple('Point', ('x', 'y'))
>>> Point(x=32, y=16)._asdict()
OrderedDict([('x', 32), ('y', 16)])
```

It should be possible to just call the `namedtuple` constructor here without inheriting from it:

```python
NewReservationRequest = namedtuple(
    'NewReservationRequest',
    (
        'real_name',
        'contact_email',
        'student_group',
        'reason',
        'date',
        'starttime',
        'endtime',
    ),
)
```

I don't think we're really losing anything by losing the docstring, since it just repeats the parameters. Plus if you do this, you can ignore my slots comment above.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+    'NewReservationRequest', [
+        'real_name',
+        'contact_email',
+        'student_group',
+        'reason',
+        'date',
+        'starttime',
+        'endtime',
+    ],
+)):
```
chriskuehl: If you're going to extend a nametuple, it's best practice to set `__slots__ = ()` again in order to make it immutable.

Here's why:

```python
# Not using inheritence
>>> Point = collections.namedtuple('Point', ('x', 'y'))
>>> p = Point(x=32, y=16)
>>> p.z = 14
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Point' object has no attribute 'z'
# ^-- good, can't jam arbitrary attributes onto it

# Trying again, this time using inheritence
>>> class Point(collections.namedtuple('Point', ('x', 'y'))):
...     pass
... 
>>> p = Point(x=32, y=16)
>>> p.z = 14
>>>
# ^-- bad, can jam arbitrary attributes on it (and mutate them later!)

# The fix is to explicitly set the slots:
>>> class Point(collections.namedtuple('Point', ('x', 'y'))):
...     __slots__ = ()
... 
>>> p = Point(x=32, y=16)
>>> p.z = 14
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Point' object has no attribute 'z'
# ^-- good, can't jam arbitrary attributes onto it again
```

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class RequestForm(forms.Form):
+
+    error_css_class = 'error'
+    required_css_class = 'required'
+
+    real_name = forms.CharField(
+        label='Full Name',
+        widget=forms.TextInput(attrs={'placeholder': 'Jane Doe'}),
```
chriskuehl: (not blocking)

I think we've traditionally used "Oski Bear" and "oski@berkeley.edu" as placeholders. Depending on how much school spirit you have, these may be better placeholders :)

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+    contact_email = forms.EmailField(
+        label='Contact e-mail',
+        validators=[wrap_validator(ocflib.misc.validators.valid_email)],
+        widget=forms.EmailInput(attrs={'placeholder': 'janedoe@berkeley.edu'}),
+    )
+
+    verify_contact_email = forms.EmailField(
+        label='Confirm contact e-mail',
```
chriskuehl: `s/e-mail/email/` for consistency with the rest of our website

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+        widget=forms.EmailInput(attrs={'placeholder': 'janedoe@berkeley.edu'}),
+    )
+
+    student_group = forms.CharField(
+        label='Student Group',
+        widget=forms.TextInput(attrs={'placeholder': 'OCF'}),
```
chriskuehl: Maybe spell out the placeholder (`Open Computing Facility`) to encourage people submitting the form to do the same. If people submit the form with an acronym/initialism, it may not be clear to us which group they're actually referring to.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+        min_length=3,
+        max_length=32,
+    )
+
+    reason = forms.CharField(
+        label='Reason for reservation',
+        widget=forms.TextInput(attrs={'placeholder': ''}),
```
chriskuehl: should this be a text area rather than just a single-line text input?

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+    verify_contact_email = forms.EmailField(
+        label='Confirm contact e-mail',
+        widget=forms.EmailInput(attrs={'placeholder': 'janedoe@berkeley.edu'}),
+    )
+
+    student_group = forms.CharField(
+        label='Student Group',
+        widget=forms.TextInput(attrs={'placeholder': 'OCF'}),
+        min_length=3,
+        max_length=32,
```
chriskuehl: 32 seems low to me, any reason we can't raise this? Even OCF's full name is 23 characters, and looking through LDAP at group accounts, many are longer than 32 (the longest is 77).

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+    reason = forms.CharField(
+        label='Reason for reservation',
+        widget=forms.TextInput(attrs={'placeholder': ''}),
+    )
+
+    date = forms.DateField(
+        label='Date of reservation (mm/dd/yy)',
+        widget=forms.DateInput(attrs={'placeholder': '02/28/18'}),
```
chriskuehl: yyyy-mm-dd would be better imo

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+    starttime = forms.TimeField(
+        label='Starting time of reservation (xx:xx)',
+    )
+
+    endtime = forms.TimeField(
+        label='Ending time of reservation (xx:xx)',
+    )
```
chriskuehl: I assume these are 24-hour times? It's probably worth specifying. It's sort of obvious to people in technical fields because we use 24-hour clocks all the time, but it may be confusing to others ("why won't it let me enter 'PM'???")

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+            'required': 'You must agree to our policies.',
+        },
+    )
+
+    def clean_verify_contact_email(self):
+        email = self.cleaned_data.get('contact_email')
+        verify_contact_email = self.cleaned_data.get('verify_contact_email')
+
+        if email and verify_contact_email:
+            if email != verify_contact_email:
```
chriskuehl: Not sure I'm following this logic correctly, but it seems like since this is a non-required field, I could just completely leave the verify email field blank?

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+                'title': 'Unable to read account information',
+            },
+        )
+
+    if request.method == 'POST':
+        form = RequestForm(request.POST)
+        if form.is_valid():
+            req = NewReservationRequest(
+                real_name=form.cleaned_data['real_name'],
+                contact_email=form.cleaned_data['contact_email'],
```
chriskuehl: Is there anything that validates this email address? We have a function in ocflib which is good at this (it even checks the DNS name). You wouldn't believe how many typos of `berkeley.edu` it's caught, even from Berkeley students.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+                endtime=form.cleaned_data['endtime'],
+            )
+
+            send_request_to_officers(req)
+            send_request_confirmation(req)
+
+            return render(
+                request,
+                'lab_reservations/pending.html',
+            )
```
chriskuehl: After a POST, it should ideally redirect to a new page. Otherwise the user can continually press "refresh" and spam us with emails. Plus the "back" button won't really behave correctly.

https://en.wikipedia.org/wiki/Post/Redirect/Get

You can see how we do it for the vhost submission form: https://github.com/ocf/ocfweb/blob/3ff4f1b4df48527b2421d7e81d7b60e96ef0e8be/ocfweb/account/vhost.py#L138

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+
+def reservation_requested(request):
+    return render(request, 'lab_reservations/pending.html', {'title': 'Reservation request successful'})
+
+
+def send_request_to_officers(request):
+    body = JINJA_MAIL_ENV.get_template(
+        'lab_reservations/mail_templates/officer_notification.jinja',
+    ).render(request=request)
+    send_mail('bod@ocf.berkeley.edu', 'New Lab Reservation Request', body, sender=request.contact_email)
```
chriskuehl: Maybe add some additional info to the subject? If we get multiple submissions at a time, it would help if the subject was `Lab Reservation Request: *Group Name*`, for example, rather than just many submissions with the same subject.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def send_request_confirmation(request):
+    body = JINJA_MAIL_ENV.get_template(
+        'lab_reservations/mail_templates/user_notification.jinja',
+    ).render(request=request)
+    send_mail(request.contact_email, '[OCF] Your recent reservation request has been submitted!', body)
```
chriskuehl: `recent` is superfluous

ocfweb/lab_reservations/templates/lab_reservations/cant-find-in-ldap.html
```diff
@@ -0,0 +1,23 @@
...
+            us from reading information about. If you get in touch with us, we
+            should be able to help.
+        </p>
+        <p>
+            If you believe this is an error, please <a href="{% url 'doc' 'contact' %}">contact us</a>
+            with your CalNet UID: <tt>{{calnet_uid}}</tt>.
+        </p>
+
+        <p>
+            Please logout <a href="{% url 'calnet_logout' %}">here</a>.
```
chriskuehl: `log out` is two words when used as a verb (similar to `log in`)

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,84 @@
...
+
+{% block content %}
+    <div class="row">
+        <div class="col-md-8 ocf-content-block">
+
+            <form method="post" action="{% url 'request_reservation' %}">
+                {% csrf_token %}
+
+
+                <hr />
```
chriskuehl: Could you include a screenshot? It seems weird to me to have a horizontal line after nothing (the CSRF token is a hidden form field).

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,84 @@
...
+
+                <p>{{form.real_name|bootstrap}}</p>
+
+                <hr />
+                Please provide a UC Berkeley e-mail address. We will use this
+                to send you a confirmation email in addition to any necessary
+                follow-ups (including an approval email).
```
chriskuehl: inconsistent use of e-mail and email

ocfweb/lab_reservations/reserve.py
kkuehler: Or something like `datetime.now().strftime("%Y-%m-%d")` looks better than an arbitrary date.

ocfweb/lab_reservations/reserve.py
kkuehler: Err... I don't know if 24 hour format is our convention, but I'd say use 12 hour format -- that's what most people are used to. You can pass `input_formats` to a `TimeField` to change the formatting. [Here](http://blog.tkbe.org/archive/date-filter-cheat-sheet/) is a cheat sheet. `%g` and `%A` will serve your purpose.

ocfweb/lab_reservations/reserve.py
EnnFour: I felt like a single line justification wouldn't be sufficient information for BOD to vote on it in some instances.

chriskuehl: i agree, that's why i suggested to make it a textarea rather than a single-line text input. are you saying it's already a textarea?

ocfweb/lab_reservations/mail_templates/user_notification.jinja
```diff
@@ -0,0 +1,12 @@
...
+
+The OCF has received your request to reserve the lab on {{request.date}}
+from {{request.starttime}} to {{request.endtime}} and it is pending staff
+approval. Your request is currently being processed. Should we encounter
+any difficulty or require any additional information, we will send you
+an email. Once we've made a decision, we'll send you an email to let
+you know.
+
+If, for whatever reason, you need to contact us about this reservation
+request, please don't hesitate to send an email to help@ocf.berkeley.edu.
```
jvperrin: I think mentioning here that they can just reply to this email (since it's sent from help@ocf.berkeley.edu) would be nice, maybe something like `please don't hesitate to reply to this email or send a new email to help@ocf.berkeley.edu`?

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,186 @@
...
+    ],
+)):
+    """Request for reservation request.
+    :param real_name:
+    :param contact_email:
+    :param student_group:
+    :param reason:
+    :param date:
+    :param starttime:
+    :param endtime:
```
jvperrin: I don't really think giving all the parameters to the class here gives any information, I mean they're just a couple lines this if anyone wants to actually look at them. Instead, maybe include here more details on what this class represents, when one is created, information like that, and remove this information at least since it's not useful and just boilerplate.

(this is essentially what @chriskuehl said below about the docstring, but I typed all of this out, so I'm going to leave it. I'd agree with him that using the `namedtuple` constructor without inheriting from it and using `_asdict()` would be cleaner here)

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,186 @@
...
+class RequestForm(forms.Form):
+
+    error_css_class = 'error'
+    required_css_class = 'required'
+
+    real_name = forms.CharField(
+        label='Full Name',
+        widget=forms.TextInput(attrs={'placeholder': 'Oski Bear'}),
+        min_length=3,
+        max_length=32,
```
jvperrin: Does this even need a min/max? I don't think adding one here really does anything besides annoy those people who might happen to have really long names if they hit the limit. I think requiring some input to be here is good, but in general I don't think we need to restrict the length of name that someone enters in this form.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,186 @@
...
+    verify_contact_email = forms.EmailField(
+        label='Confirm contact email',
+        widget=forms.EmailInput(attrs={'placeholder': 'oski@berkeley.edu'}),
+    )
+
+    student_group = forms.CharField(
+        label='Student Group',
+        widget=forms.TextInput(attrs={'placeholder': 'Open Computing Facility'}),
+        min_length=3,
+        max_length=100,
```
jvperrin: Same as with the `real_name` field above this, I don't see why this needs any kind of limit.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,186 @@
...
+    # (alumni etc. might not be readable in LDAP but can still auth via CalNet)
+    if not user_attrs_ucb(calnet_uid):
+        return render(
+            request,
+            'lab_reservations/lab_reservations/cant-find-in-ldap.html',
+            {
+                'calnet_uid': calnet_uid,
+                'title': 'Unable to read account information',
+            },
+        )
```
jvperrin: Why do we need to be able to look them up in LDAP/CalNet at all? They enter their own name and contact email, so shouldn't we leave this out entirely? The decorator about this method ensures they've logged in anyway, so I really don't think we need this part at all. Same with the `cant-find-in-ldap.html` template, it seems like it's left over from the account creation request page/form to me.

ocfweb/lab_reservations/reserve.py
EnnFour: Tested on supernova: By leaving the field blank, a popup occurs asking users to "please fill out this field."
image

ocfweb/lab_reservations/reserve.py
EnnFour: Within the RequestForm class, we have a line `validators=[wrap_validator(ocflib.misc.validators.valid_email)],`

Does this assuage any concerns?

ocfweb/lab_reservations/reserve.py
EnnFour: I done goofed. Should be fixed in a couple minutes.

ocfweb/lab_reservations/reserve.py
abizer: Is that something that might be browser-specific?

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,168 @@
...
+
+            send_request_to_officers(req)
+            send_request_confirmation(req)
+
+            return redirect(reverse('request_reservation_success'))
+
+    else:
+        form = RequestForm()
+
+    return render(
```
kpengboy: If you're going to have an `else` block you might as well move this statement inside it

EnnFour: If we move it inside the else block we need another return statement for the case where the first if statement passes and the second fails.

ocfweb/lab_reservations/mail_templates/officer_notification.jinja
```diff
@@ -0,0 +1,11 @@
+The OCF has received a lab reservation request!
+
+Submitter: {{request.real_name}}
+Email: {{request.contact_email}}
+Student Group: {{request.student_group}}
```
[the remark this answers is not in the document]
EnnFour: Fixed

ocfweb/lab_reservations/mail_templates/user_notification.jinja
chriskuehl: ++

ocfweb/lab_reservations/reserve.py
chriskuehl: still need to do this?

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,168 @@
...
+        'group',
+        'reason',
+        'date',
+        'starttime',
+        'endtime',
+    ],
+)):
+
+    def to_dict(self):
+        return self._asdict()
```
chriskuehl: a little nitpicky, but i'd just have consumers call `_asdict()` as this is a standard method (i don't need to look at the class implementation to see what it does, i'll know it does the same thing as `_asdict` on every other namedtuple)

this is all a little muddied by the fact that it's an underscored method (which usually means "private", but we frequently ignore in namedtuples)

ok to keep it either way, up to you

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,168 @@
...
+        widget=forms.DateInput(attrs={'placeholder': datetime.now().strftime('%Y-%m-%d')}),
+    )
+
+    starttime = forms.TimeField(
+        label='Starting time of reservation (24 hour, i.e. 20:00)',
+    )
+
+    endtime = forms.TimeField(
+        label='Ending time of reservation (24 hour, i.e. 22:00)',
+    )
```
chriskuehl: ideally for the dates and times, we'd use a fancy jquery date picker widget or something so that the non-technical masses don't have to figure out how to convert 8pm to 24-hour time, but imo this is plenty good enough to be merged

ocfweb/lab_reservations/reserve.py
chriskuehl: yeah, i must have missed that. assuming it works, i think this is fine.

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,168 @@
...
+    return render(request, 'lab_reservations/pending.html', {'title': 'Reservation request successful'})
+
+
+def send_request_to_officers(request):
+    body = JINJA_MAIL_ENV.get_template(
+        'lab_reservations/mail_templates/officer_notification.jinja',
+    ).render(request=request)
+    send_mail(
+        'bod@ocf.berkeley.edu',
+        'New Lab Reservation Request: ' + str(request.group),
```
chriskuehl: might be cleaner to use `format`:

```python
'New Lab Reservation Request: {}'.format(request.group)
```

this will get even cleaner if we ever get python3.6 and f-strings :)

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,85 @@
...
+            <form method="post" action="{% url 'request_reservation' %}">
+                {% csrf_token %}
+
+
+                <p>{{form.real_name|bootstrap}}</p>
+
+                <hr />
+                Please provide a UC Berkeley email address. We will use this
+                to send you a confirmation email in addition to any necessary
+                follow-ups (including an approval email).
```
chriskuehl: kinda feels like these labels should be in a `<p>` tag?

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,85 @@
...
+
+                <hr />
+
+                <p>{{form.group|bootstrap}}</p>
+                <p>{{form.reason|bootstrap}}</p>
+
+                <hr />
+
+                Please enter all times in 24-hour format (i.e. 13:00
+                instead of 1:00 PM))
```
chriskuehl: extra )

ocfweb/lab_reservations/templates/lab_reservations/pending.html
```diff
@@ -0,0 +1,9 @@
+{% extends "base.html" %}
+
+{% block content %}
+    <div class="ocf-content-block">
+        <p>Your lab reservation request has successfully
+           been submitted and is pending approval!</p>
```
chriskuehl: nit: i'd encourage writing multi-line elements like:

```html
<p>
    Your lab reservation request has been successfully
    submitted and is pending approval!
</p>
```

for the same reason behind most of our python style guidelines: it reduces diff churn, which makes git blame, merge conflicts, etc. better

ocfweb/lab_reservations/templates/lab_reservations/index.html
```diff
@@ -0,0 +1,85 @@
...
+                <p><img src="{% static 'img/penguin.svg' %}" width="100" height="132" /></p>
+                <h4>OCF &hearts; Python, Linux, Open Source</h4>
+                <p>
+                    The OCF is provided for students, by students. Sound interesting?
+                    <br />
+                    <a href="{% url 'about-staff' %}">Learn about getting involved!</a>
+                </p>
+            </div>
+
+            <hr />
```
chriskuehl: weird horizontal line at the end? not sure if this looks better on the rendered page or something, didn't see a screenshot (might have missed it in all the PR comments)

ocfweb/lab_reservations/reserve.py
jvperrin: bump, I don't think we should require calnet login for this

ocfweb/lab_reservations/reserve.py
```diff
@@ -0,0 +1,169 @@
...
+            'lab_reservations/lab_reservations/cant-find-in-ldap.html',
+            {
+                'calnet_uid': calnet_uid,
+                'title': 'Unable to read account information',
+            },
+        )
+
+    if request.method == 'POST':
+        form = RequestForm(request.POST)
+        if form.is_valid():
```
jvperrin: So if the form isn't valid here, doesn't this not re-render the form because the `else` is indented to be only on a get request? What happens if the form is invalid but made a POST request?

ocfweb/lab_reservations/templates/lab_reservations/cant-find-in-ldap.html
```diff
@@ -0,0 +1,23 @@
...
+        <p>
+            If you believe this is an error, please <a href="{% url 'doc' 'contact' %}">contact us</a>
+            with your CalNet UID: <tt>{{calnet_uid}}</tt>.
+        </p>
+
+        <p>
+            Please log out <a href="{% url 'calnet_logout' %}">here</a>.
+        </p>
+    </div>
+{% endblock %}
```
jvperrin: Yeah I think this whole page can be removed, we shouldn't need to look up accounts in calnet/LDAP since they provide their own contact information, etc. in the form.

ocfweb/lab_reservations/templates/lab_reservations/index.html
jvperrin: I do think we should allow other groups to use the lab, for instance I'm not sure the FAFSA people we let use the lab had a group account associated with them, yet they were a great example of a group that should be using this.

ocfweb/lab_reservations/templates/lab_reservations/pending.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+
+{% block content %}
+    <div class="ocf-content-block">
+        <p>Your lab reservation request has successfully
+           been submitted and is pending approval!
```
jvperrin: Indentation here is a bit strange, I'd probably do something like this instead to make it clearer:

```
<p>
    Your lab reservation request has successfully
    been submitted and is pending approval!
</p>
```

ocfweb/lab_reservations/urls.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+urlpatterns = [
+
+    # reservation creation form
+    url(r'^request/$', request_reservation, name='request_reservation'),
+
+    # reservation pending
+    url(r'^request/pending/$', request_reservation_success, name='request_reservation_success'),
+
```
jvperrin: I'd remove the extra newline here and on line 8

ocfweb/lab_reservations/reserve.py
EnnFour: If the form is invalid but makes a POST request, we just return the same page with everything still filled out (since the return on line 138 is not in any of these conditionals). Looks something like this: 
image

jvperrin: Ah yeah, I missed that section, thanks.

PR #367, opened 2018-01-31
Add password reset link to login page

ethanhs:
Fixes rt#6836

### DIFF
--- a/ocfweb/login/templates/login/ocf/login.html
+++ b/ocfweb/login/templates/login/ocf/login.html
@@ -19,7 +19,7 @@
                     {% csrf_token %}
                     <div style="max-width: 300px;">
                         {{ form|bootstrap }}
-            <small style="float: right; margin-top: 6pt;">
+            <small class="reset-link">
                 <a href="{% url 'change_password' %}">Forgot your password?</a>
             </small>
                     </div>
--- a/ocfweb/login/templates/login/ocf/login.html
+++ b/ocfweb/login/templates/login/ocf/login.html
@@ -19,12 +19,12 @@
                     {% csrf_token %}
                     <div style="max-width: 300px;">
                         {{ form|bootstrap }}
-            <small class="reset-link">
-                <a href="{% url 'change_password' %}">Forgot your password?</a>
-            </small>
+                        <small class="reset-link">
+                            <a href="{% url 'change_password' %}">Forgot your password?</a>
+                        </small>
                     </div>
                     <input class="btn btn-primary" type="submit" value="Login" />
-            </form>
+                </form>
             </div>
             <div class="col-sm-4">
                 <h3>OCF Accounts</h3>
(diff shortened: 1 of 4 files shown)


ocfweb/login/urls.py
```diff
@@ -1,12 +1,15 @@
...
+from ocfweb.account.urls import urlpatterns as account
 from ocfweb.login.calnet import login as calnet_login
 from ocfweb.login.calnet import logout as calnet_logout
 from ocfweb.login.ocf import login
 from ocfweb.login.ocf import logout
 
 
 urlpatterns = [
+    url(r'^account/', include(account)),
+
```
jvperrin: Why do all these changes need to be made? I believe the link will just work without all of this.

jvperrin:
I was messing with this a bit before, and I thought putting the link in a position like this would work pretty well.  I thought it looked nice having the link in the bottom right since it worked well with the login button, but it could probably be shifted down a bit so that it aligns with the login button vertically instead of floating a bit above it:

reset-link-position

I did that with the following (which I think should be cleaned up a bit to not use as much inline CSS):

    <div style="max-width: 300px;">
        {{ form|bootstrap }}
        <small style="float: right;">
            <a href="{% url 'change_password' %}">Forgot your password?</a>
        </small>
    </div>

ocfweb/login/urls.py
ethanhs: Oh, you are right, it is already included in the main urlpatterns.

ethanhs:
@jvperrin I got it to shift down nicely if I add `margin-top: 6pt;`.
image

However, is it worth it to add the complexity of a new scss for login for just one dom element?

ocfweb/login/templates/login/ocf/login.html
```diff
@@ -19,9 +19,12 @@
                     {% csrf_token %}
                     <div style="max-width: 300px;">
                         {{ form|bootstrap }}
+            <small class="reset-link">
+                <a href="{% url 'change_password' %}">Forgot your password?</a>
+            </small>
```
chriskuehl: indented weird (plus, hard tabs?)

ethanhs: Hm, that is strange I'll fix that. When you say hard tabs, are you suggesting I use them in future or something else?

chriskuehl: yeah, we generally use 4 space indents at the ocf (unless the language has a well-known convention, e.g. overwhelming puppet convention is 2 space indents, so we use that)

ethanhs:
Any more changes needed for this?

jvperrin:
Nope, I think that's it! I've sent you an invitation to join to OCF organization, then you should be able to merge your own PR. Thanks for the change!

ethanhs:
Thank you @jvperrin!

PR #369, opened 2018-02-13
Add clarification to approve docs for non-student groups

dkess:
This just documents that OID should be 0 when approving non student group accounts.


sahilhasan:
but then how can I make fun of mdcha :(

PR #370, opened 2018-02-21
Parallelize tests run and bump requirements

jvperrin:
This cuts down teest time from around 120 seconds using one core on `reaper` to around 25 seconds using all cores at once. My only real concern with this is that if a lot of builds of ocfweb run at once that it'll overwhelm a machine, but I think it'll be ok, especially since all tests finish in under 30 seconds anyway, so any high load should be over relatively quickly.


pytest.ini
```diff
@@ -1,3 +1,8 @@
 [pytest]
 DJANGO_SETTINGS_MODULE=ocfweb.settings
-addopts=--durations=10
+addopts=--durations=10 -n auto
```
ethanhs: Probably want to document how people can lower this, as pytest will by default gobble up all of the available threads (which people may not like).

chriskuehl: I would vote to avoid duplicating pytest's documentation, but linking to it may be appropriate.

jvperrin: Agreed, mentioning here that this will use all the available CPU threads (and linking to xdist's docs somewhere) is a good idea.

PR #371, opened 2018-02-24
ocf tv that shows desktop usage

shex1627:
http://nyx.ocf.berkeley.edu:8934/tv/ 

image


shex1627:
It also provides a api that gives 
total desktops(public only)
and list of desktops in use(public only)

http://nyx.ocf.berkeley.edu:8934/api/lab/desktops 
sample json file
{"total_desktops": 28, "desktops_in_use": ["cyclone", "madcow", "invasion", "venom", "asteroid", "hurricane", "firewhirl"]}

abizer:
some bugs/issues remain to be fixed in this

ocfweb/tv/templates/tv/hours.html
```diff
@@ -1,7 +1,5 @@
-{% extends 'tv/base.html' %}
 {% load tv_formatting %}
```
jvperrin: I don't think you need this extra load here if you have it in the parent template, but I could be wrong on that.

ocfweb/api/lab.py
```diff
@@ -0,0 +1,41 @@
...
+@cache()
+def _enumerate_desktops():
+    return list_desktops(public_only=True)
+
+
+@periodic(15)
+def _get_active_desktops():
+    """List which desktops are currently in use."""
+
+    # https://github.com/ocf/ocflib/blob/master/ocflib/lab/ocfstats.sql#L70
```
jvperrin: I'd suggest making this a permalink or posting the relevant section here, otherwise this link will potentially point to the wrong line in the future.

ocfweb/api/lab.py
```diff
@@ -0,0 +1,41 @@
...
+            'WHERE `end` IS NULL;',
+        )
+
+    return {hostname_from_domain(i['host']) for i in c}
+
+
+def desktop_usage(request):
+    desktops = _enumerate_desktops()
+
+    # get the list of public-only dekstops in use
```
jvperrin: `dekstops` -> `desktops`

Also I'd call these "publically-available", since staff can (and does) use them too, but this is a nitpick.

ocfweb/tv/templates/tv/desktop_usage.html
```diff
@@ -0,0 +1,16 @@
+<div class="page-tv-hours">
+  <p>There are <span id="desktop_num">_</span> desktops available.</p>
+  <script>
+    $(document).ready(function() {
+      function update() {
+        $.getJSON("{% url 'desktop_usage' %}", function(res) {
+          document.getElementById("desktop_num").innerHTML =
+          (res.total_desktops - res.desktops_in_use.length);
```
jvperrin: Since you already include jQuery, I think this can be made simpler, something like `$('#desktop_num').text((res.total_desktops - res.desktops_in_use.length))`. Also I don't think you need to return the full list of desktops in use, but I'm guessing that's planned to be used in the future for a desktop map or something that shows which desktops are being used currently.

ocfweb/tv/templates/tv/desktop_usage.html
```diff
@@ -0,0 +1,16 @@
...
+        });
+      }
+
+      update();
+      setInterval(update, 15000);
```
jvperrin: Nice, I think this is a decent amount of time to give between requests. You could even go lower if you want, but I think this is good.

ocfweb/api/api.py
```diff
@@ -1,18 +0,0 @@
-from django.http import JsonResponse
-from ocflib.lab.hours import Day
-
-
-def hours(request):
```
chriskuehl: was this file meant to be deleted?

ocfweb/api/lab.py
```diff
@@ -0,0 +1,41 @@
...
+    # https://github.com/ocf/ocflib/blob/master/ocflib/lab/ocfstats.sql#L70
+    # we don't use users_in_lab_count_public because we're looking for
+    # desktops in use, and the view does COUNT(DISTINCT users)
+    with get_connection() as c:
+        c.execute(
+            'SELECT `host` FROM `session_duration_public` '
+            'WHERE `end` IS NULL;',
+        )
+
+    return {hostname_from_domain(i['host']) for i in c}
```
chriskuehl: i'd avoid the use of `i` as a variable name here, i/j/k are all traditionally used for iterating over indices and not elements. just do `{hostname_from_domain(session['host']) for session in c}` ?

ocfweb/api/lab.py
```diff
@@ -0,0 +1,41 @@
...
+from ocflib.infra.hosts import hostname_from_domain
+from ocflib.lab.stats import get_connection
+from ocflib.lab.stats import list_desktops
+
+from ocfweb.caching import cache
+from ocfweb.caching import periodic
+
+
+@cache()
+def _enumerate_desktops():
```
chriskuehl: i would give this a (long) TTL, or even make it periodic, as this list is not actually static (just almost static)

ocfweb/api/lab.py
```diff
@@ -0,0 +1,41 @@
...
+        c.execute(
+            'SELECT `host` FROM `session_duration_public` '
+            'WHERE `end` IS NULL;',
+        )
+
+    return {hostname_from_domain(i['host']) for i in c}
+
+
+def desktop_usage(request):
+    desktops = _enumerate_desktops()
```
chriskuehl: i'd suggest `list_desktops` instead of `enumerate_desktops`; you're not really enumerating anything here, and it might confuse people who are used to the python function `enumerate` which numbers a list

ocfweb/api/lab.py
```diff
@@ -0,0 +1,41 @@
...
+from ocfweb.caching import periodic
+
+
+@cache()
+def _enumerate_desktops():
+    return list_desktops(public_only=True)
+
+
+@periodic(15)
+def _get_active_desktops():
```
chriskuehl: what is an active desktop? not clear from the name if this means any available desktop or a desktop currently in use.

can we just call it `get_desktops_currently_used` or something?

ocfweb/api/lab.py
chriskuehl: imo this comment is sort of redundant if you use good variable names anyway

ocfweb/tv/templates/tv/base.html
```diff
@@ -1,9 +1,8 @@
 <!doctype html>
 
 <html>
-
   {% include 'partials/head.html' %}
-  <meta http-equiv='refresh' content='600' />
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
```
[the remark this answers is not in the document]
chriskuehl: ++

ocfweb/tv/templates/tv/desktop_usage.html
chriskuehl: +1 avoid setting innerhtml directly, it is easy to cause an xss vulnerability

ocfweb/tv/templates/tv/desktop_usage.html
```diff
@@ -0,0 +1,16 @@
+<div class="page-tv-hours">
+  <p>There are <span id="desktop_num">_</span> desktops available.</p>
+  <script>
+    $(document).ready(function() {
+      function update() {
+        $.getJSON("{% url 'desktop_usage' %}", function(res) {
```
chriskuehl: 4 space indents

ocfweb/tv/templates/tv/tv.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'tv/base.html' %}
+{% load tv_formatting %}
+
+{% block tv-body %}
+{% include 'tv/hours.html' %}
+{% include 'tv/desktop_usage.html' %}
```
chriskuehl: do these things really need to be split out into partials?

i think there's an unnecessary amount of indirection going on here

also, indentation

ocfweb/api/api.py
jvperrin: I was confused on that too, but I don't see it being used anywhere that I can tell, so I think it's ok?

ocfweb/api/lab.py
jvperrin: Yeah, agreed on that too, just removing the comment would be a good solution if you change the function name to be more clear.

ocfweb/tv/templates/tv/hours.html
abizer: Unfortunately, parents do not pass these onto their children :(

ocfweb/tv/templates/tv/desktop_usage.html
shex1627: I am thinking about updating it 1-5 seconds on the ocfside and 5-10 seconds on js. So people use this website would have a maximum of 15 second lag. If we go with 1s + 5s, would that be a burden/spam to our server? Imo fast update is pretty important for in-lab users waiting to get a desktop they cannot see.

ocfweb/tv/templates/tv/base.html
```diff
@@ -1,9 +1,8 @@
 <!doctype html>
 
 <html>
-
   {% include 'partials/head.html' %}
-  <meta http-equiv='refresh' content='600' />
```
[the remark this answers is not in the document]
shex1627: @abizer  
what is your take on this?

ocfweb/api/api.py
shex1627: yeah it is deleted because we don't see it get to be used anywhere.

ocfweb/api/lab.py
shex1627: I think the list of desktops will not change unless we add/delete a desktop. The situation I can think of which we need to update is when a desktop stops working (from like hardware failures), if so, is list_desktops able to detect the change?

ocfweb/tv/templates/tv/base.html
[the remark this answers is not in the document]
shex1627: Jquery doesn't load if I put 
<script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
before </body>.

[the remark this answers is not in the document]
jvperrin: Does it give any errors or anything? What do you mean by it doesn't load?

You might need to put the script tag on the desktop page after the jQuery load for it to work, but in general you want to put all the JS at the bottom of the page so it doesn't block rendering. On a small page like this one it shouldn't matter much, but for larger pages or ones with more JS it'll matter more.

ocfweb/api/lab.py
chriskuehl: Right, I agree it won't change unless we add/delete a desktop. That happens sometimes :)

ocfweb/tv/templates/tv/desktop_usage.html
jvperrin: I think it should be fine with what you have right now, although we might want to make the SQL query a bit faster/easier maybe?

ocfweb/api/lab.py
```diff
@@ -0,0 +1,40 @@
...
+    """List which desktops are currently in use."""
+
+    # https://github.com/ocf/ocflib/blob/master/ocflib/lab/ocfstats.sql#L70
+    # we don't use users_in_lab_count_public because we're looking for
+    # desktops in use, and the view does COUNT(DISTINCT users)
+    with get_connection() as c:
+        c.execute(
+            'SELECT `host` FROM `session_duration_public` '
+            'WHERE `end` IS NULL;',
+        )
```
jvperrin: I feel like making a new SQL view to show free desktops would make this easier, then you could just get that view directly and could also get the count of the free desktops from just the count of the elements in that view.

ocfweb/tv/templates/tv/tv.html
abizer: I split this apart like this for a future PR where we could add a key to `lab_status.yaml` to display a message on the TV if that became necessary, e.g.

```
{% block tv-body %}
  {% if tv_message %}
    {{ tv_message }}
  {% else %}
    {% include 'tv/hours.html' %}
```
etc.

ocfweb/api/lab.py
shex1627: Another way we could create the view is that:
column1: host (list of all desktops)
column2: if_login (1/True means logged in, 0/False means not)

shex1627:
Another way we could create the view is that:
column1: host (list of all desktops)
column2: if_login (1/True means logged in, 0/False means not) 
####
maybe we can later add fields like:
if_working, login_duration, staff_name (if applies)
####
those could be helpful if we want to create a lab map.

abizer:
merging this so impending lab map work can build on it

PR #374, opened 2018-04-11
error404 page

ivyn:
A non-plaintext non-ugly 404 page with links to the main site.


ocfweb/error404/er404.py
```diff
@@ -0,0 +1,9 @@
+from django.shortcuts import render
+
+
+# Function for rendering the error 404 page
```
chriskuehl: consider using a docstring instead of a comment; if you use a docstring, it gets jammed under `__doc__` for the function, which enables tools like `pydoc` and sphinx to generate useful documentation

these are standardized in https://www.python.org/dev/peps/pep-0257/

ocfweb/error404/er404.py
```diff
@@ -0,0 +1,9 @@
...
+def error404(request):
+    return render(
+            request,
+            '404.html')
```
chriskuehl: probably better to just write this all on one line? `return render(request, '404.html')`

ocfweb/error404/templates/404.html
```diff
@@ -0,0 +1,124 @@
+<!doctype html>
+<!--
+      Think this site is hideous? We could use your help!
+      More info at https://www.ocf.berkeley.edu/about/staff :-)
+
+      This is ocfweb version 2018-03-06-T21-08-09-git4c65c52
+      https://github.com/ocf/ocfweb/
+-->
+{% load staticfiles %}
+<html class="page-home">
```
chriskuehl: shouldn't have page-home for something other than the home page

ocfweb/error404/templates/404.html
```diff
@@ -0,0 +1,124 @@
+<!doctype html>
+<!--
+      Think this site is hideous? We could use your help!
+      More info at https://www.ocf.berkeley.edu/about/staff :-)
+
+      This is ocfweb version 2018-03-06-T21-08-09-git4c65c52
+      https://github.com/ocf/ocfweb/
+-->
```
chriskuehl: sort of confused why this duplicates our base template; is it possible to build this so it extends from the base? (see how other pages do it for an example)

ocfweb/error404/templates/404.html
```diff
@@ -0,0 +1,124 @@
...
+        <meta charset="utf-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+
+            <meta name="description" content="The Open Computing Facility is an all-volunteer student organization dedicated to free and open-source computing for all UC Berkeley students.">
+
+
+
+
+            <meta property="og:image" content="https://static.ocf.berkeley.edu/img/penguin-opengraph.png"/>
```
chriskuehl: this kind of thing should use the asset_url helpers and not hard-code

ocfweb/error404/templates/404.html
```diff
@@ -0,0 +1,124 @@
...
+                                <a href="/login/login/?next=/" class="ocf-account-dropdown">
+                                    <span class="glyphicon glyphicon-user" aria-hidden="true"></span>
+                                    Log In
+                                </a>
+                            </li>
+
+                    </ul>
+                </div>
+            </div>
+        </div>
```
chriskuehl: definitely don't want to duplicate all the header stuff, let's see if we can find a way to extend the base

(going to stop commenting on this file)

ocfweb/settings.py
```diff
@@ -16,6 +16,7 @@
     'dev.ocf.berkeley.edu',
     'dev-www.ocf.berkeley.edu',
     'ocfweb.ocf.berkeley.edu',
+    'supernova.ocf.berkeley.edu'
```
chriskuehl: this shouldn't be necessary since it's not enforced in debug mode, can you talk more about how you're hitting this?

ocfweb/settings.py
```diff
@@ -33,6 +34,7 @@
     'ocfweb.docs',
     'ocfweb.login',
     'ocfweb.main',
+    'ocfweb.error404',
```
chriskuehl: don't think we want an entire app just for this one page? maybe we could keep it under `main` or make a separate app for all errors?

ocfweb/settings.py
bzhen: Agreed -- fixed this!

ocfweb/error404/templates/404.html
bzhen: Thank you for all the helpful comments, Chris! We cleaned this file (a lot) and extended the base.

bzhen:
All comments reviewed and implemented!

bzhen:
Yes, I tested this. However, I ran into some issues when setting the DEBUG
variable to False. Right now, the 404 page seems to only show for the url:
ocf.berkeley.edu/404 -- looking into this now.

kpengboy:
Also, in the future, for nontrivial changes I would recommend bumping an unapproved PR that hasn't gotten recent attention for reviews prior to merging it. People might still have feedback.

jvperrin:
I do agree, but that being said I don't think that would have helped much in this case since all the tests passed, and the changes looked good, I think the issue was just that we hadn't worked with custom error pages before so it wasn't clear what the best/easiest way to do things was. Also the 500ing on 404 error pages wasn't clear either since you need to be in non-`DEBUG` mode to even see that behavior, so it wasn't as easy to test.

Anyway, yes, bumping non-reviewed pull requests on Slack/GitHub is generally good.

PR #377, opened 2018-04-27
changed accounts graphs to highcharts

Helen-Yang:
Changed stats/accounts graphs from matplotlib to highcharts

accountgraphs

### DIFF
--- a/ocfweb/stats/accounts.py
+++ b/ocfweb/stats/accounts.py
@@ -1,25 +1,38 @@
+import time
 from collections import defaultdict
 from datetime import date
 from datetime import timedelta
 
-from django.http import HttpResponse
 from django.shortcuts import render
-from matplotlib.dates import DateFormatter
-from matplotlib.figure import Figure
 from ocflib.infra.ldap import ldap_ocf
 from ocflib.infra.ldap import OCF_LDAP_PEOPLE
 
 from ocfweb.caching import cache
-from ocfweb.caching import periodic
-from ocfweb.component.graph import plot_to_image_bytes
 
 
 def stats_accounts(request):
+    account_data = _get_account_stats()
     return render(
         request,
         'stats/accounts.html',
         {
             'title': 'Account Statistics',
+            'total': [
+                {
+                    'name': 'Accounts',
+                    'showInLegend': False,
+                    'animation': False,
+                    'data': account_data['cumulative_accounts'],
+                },
+            ],
+            'group': [
+                {
+                    'name': 'Group Accounts',
+                    'showInLegend': False,
+                    'animation': False,
+                    'data': account_data['cumulative_group_accounts'],
+                },
+            ],
         },
     )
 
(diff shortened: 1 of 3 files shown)


ocfweb/stats/accounts.py
```diff
@@ -32,7 +46,7 @@ def _get_account_stats():
 
     # Some accounts don't have creation times, so we assume that as UIDs
     # increase, so does time. We use this assumption to fill in the gaps.
-    start_date = date(1989, 1, 1)
+    start_date = date(1995, 8, 21)
```
abizer: what's the reasoning behind changing this date?

ocfweb/stats/accounts.py
```diff
@@ -66,63 +80,8 @@ def _get_account_stats():
         cumulative_group_accounts.append(group_total)
 
         cur += one_day
-
+    dates = [time.mktime(d.timetuple())*1000 for d in dates]
```
abizer: where does the 1000 come from? Is there a better way to implement this, such as inside the loop where `dates` is populated to begin with?

ocfweb/stats/accounts.py
Helen-Yang: The data was all zeroes before that time so I changed it to make the graph look better visually.

ocfweb/stats/accounts.py
Helen-Yang: Highcharts wants time in milliseconds. It's inside the loop now.

ocfweb/stats/templates/stats/accounts.html
```diff
@@ -1,14 +1,94 @@
...
     <div class="ocf-content-block">
         {% stats_navbar %}
-        <p>
-            <img class="img-responsive" src="{% url 'cumulative_accounts_graph' %}" />
-        </p>
-        <p>
-            <img class="img-responsive" src="{% url 'cumulative_group_accounts_graph' %}" />
-        </p>
+    <div id="total"></div>
+    <div id="group"></div>
```
jvperrin: I think indenting these to match the other content in the `div` would make sense:

```
<div class="ocf-content-block">
    {% stats_navbar %}
    <div id="total"></div>
    <div id="group"></div>
</div>
```

ocfweb/stats/templates/stats/accounts.html
```diff
@@ -1,14 +1,94 @@
...
+
+{% block inline_js %}
+    <script src="https://code.highcharts.com/5.0.7/highcharts.js"></script>
+    <script>
+        $(document).ready(function () {
+            Highcharts.setOptions({
+                global: {
+                    useUTC: false
+                }
+            });
```
jvperrin: I think having a more standardized template for this instead of duplicating the same first part across multiple pages (like in the pages printed template I've linked below) would be good, but that's maybe more of an enhancement after this PR's changes. If you'd like to do it then I think that would make this a bit cleaner though:

https://github.com/ocf/ocfweb/blob/96216ddd76631356d9057fd54169d463941c80a5/ocfweb/stats/templates/stats/printing/pages-printed.html#L30-L74

ocfweb/stats/templates/stats/accounts.html
Helen-Yang: Fixed this, consolidated the highcharts options for the 2 graphs, and replaced time on the tooltips with years. I think it looks fine on mobile.

PR #378, opened 2018-05-16
ocf.io/tv/labmap

raymondnumbergenerator:
currently in production at ocf.io/raymondng/labmap


kpengboy:
Ah I guess these comments don't render properly because the file is too large. Perhaps the folks at GitHub never tested that?

kpengboy:
It will be easier to see what the comments refer to if you can split this file into an HTML file, SVG file, and Javascript file.

raymondnumbergenerator:
thanks, i'll clean it up sometime soon

i think i have the original svg lying around somewhere but it's so far removed from the one that's actually on the labmap since i've written a bunch of stuff on top of it

PR #379, opened 2018-05-16
Fixed previous error404 commit

bzhen:
There were some complications with loading static pages when DEBUG was set to False to test the 404 page on supernova. I was able to get around this and load the static files by adding `--insecure` when running `manage.py runserver`.

### DIFF
--- a/ocfweb/context_processors.py
+++ b/ocfweb/context_processors.py
@@ -14,7 +14,7 @@ from ocfweb.environment import ocfweb_version
 
 
 def get_base_css_classes(request):
-    if request.resolver_match.url_name:
+    if request.resolver_match and request.resolver_match.url_name:
         page_class = 'page-' + request.resolver_match.url_name
         yield page_class
 
--- a/ocfweb/templates/404.html
+++ b/ocfweb/templates/404.html
@@ -14,6 +14,6 @@
         height: 1000px;
     }
 </style>
-     <img src="{% static 'img/404penguin.png' %}" alt="Mountain View" style='height: 100%; width: 100%'/>
+     <img src="{% static 'img/404penguin.png' %}" alt="404 Penguin" style='height: 100%; width: 100%'/>
 </div>
 {% endblock %}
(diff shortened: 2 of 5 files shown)


ocfweb/context_processors.py
```diff
@@ -14,13 +14,19 @@
...
+        # Need this for the 404 page
+        yield ''
+    elif request.resolver_match.url_name:
         page_class = 'page-' + request.resolver_match.url_name
         yield page_class
 
         for arg in request.resolver_match.args:
             page_class += '-' + re.sub(r'[^a-zA-Z_\-]', '-', arg)
             yield page_class
+    else:
```
abizer: do what @jvperrin suggested earlier, which is also what we did when we looked at this together a couple weeks ago, rather than ending up raising the rather under-descriptive `Exception`

ocfweb/context_processors.py
```diff
@@ -14,13 +14,19 @@
 
 
 def get_base_css_classes(request):
-    if request.resolver_match.url_name:
+    if not request.resolver_match:
+        # Need this for the 404 page
+        yield ''
+    elif request.resolver_match.url_name:
```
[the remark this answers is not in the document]
bzhen: Yeah, that works too if I get rid of the else case. I wasn't sure if `get_base_css_classes(request)` would cause `ocf_template_processor(request)` to error if I only returned `None`.

ocfweb/templates/404.html
```diff
@@ -14,6 +14,6 @@
         height: 1000px;
     }
 </style>
```
kpengboy: My previous comment still stands: stylesheets should be specified in the SCSS directory, and not in the middle of a `<div>`.

bzhen: Will add the stylesheet improvements later!

jvperrin: I do think this would be good to do, however currently it is a bit harder to do this since some kind of identifying class needs to be added to the body of the email pages that isn't currently added, so some of the Django side of things needs to be changed to accomplish that.

PR #381, opened 2018-06-05
Remove repetitive code

glee-:
a small change


raymondnumbergenerator:
The reason those 2 lines are separate is because the lab can be in nightmode without being closed all day. During regular days, the lab will turn dark but still display hours. It should only change to "closed all day" when it's actually closed all day.

glee-:
Thanks for the clarification

PR #384, opened 2018-06-17
Add internal firewall docs

kpengboy:
Does it make sense to you all? Suggestions for improvement?

### DIFF
--- a/ocfweb/docs/docs/staff/backend/internal-firewalls.md
+++ b/ocfweb/docs/docs/staff/backend/internal-firewalls.md
@@ -10,7 +10,8 @@ are those which are not running user code and are not staff VMs, specifically
 those in the IP range 5-90 (excluding the ones in the DMZ listed below). The DMZ
 consists of all other machines, including:
 
- * Those running user code (tsunami, werewolves, death, and desktops)
+ * Those running user code (i.e. the public SSH login, apphosting, and web
+   servers, and desktops)
  * Staff VMs
  * External machines plugged into the OCF network (e.g. staffers' laptops)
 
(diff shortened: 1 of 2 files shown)


kpengboy:
@kkuehler @jvperrin @abizer Thanks for the feedback. Any comments on the revised version?

ocfweb/docs/docs/staff/backend/internal-firewalls.md
```diff
@@ -0,0 +1,126 @@
...
+traffic to the OCF from outside the OCF network, internal firewalls are
+responsible for regulating network traffic between different machines within the
+OCF.
+
+OCF machines are broadly classed into internal and [DMZ][dmz]. Internal machines
+are those which are not running user code and are not staff VMs, specifically
+those in the IP range 5-90 (excluding the ones in the DMZ listed below). The DMZ
+consists of all other machines, including:
+
+ * Those running user code (tsunami, werewolves, death, and desktops)
```
chriskuehl: I think it would be better to avoid server names here, especially for e.g. werewolves which may get renamed.

tsunami and death may be okay since they're unlikely to change.

ocfweb/docs/docs/staff/backend/internal-firewalls.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Debugging
+
+### `iptables` commands
+
+Note that all of the following commands need to be run as root.
+
+The `iptables` command allows you to inspect and debug IPv4 firewall rules:
+
+ * `iptables -L`: list firewall rules
```
chriskuehl: `iptables-save` is a useful command (and usually more useful, imo) for inspecting iptables state

ocfweb/docs/docs/staff/backend/internal-firewalls.md
jvperrin: I agree, werewolves is being renamed quite soon in fact.

ocfweb/docs/docs/staff/backend/internal-firewalls.md
kpengboy: Thanks for the suggestion; I've added `iptables -S`, which is a superset.

kpengboy:
Any final comments?

PR #386, opened 2018-07-06
Add security.txt for OpenBugBounty

kpengboy:
We need to add a file like this to reregister for [OpenBugBounty](https://www.openbugbounty.org)

### DIFF
--- a/ocfweb/main/openbugbounty.py
+++ b/ocfweb/main/openbugbounty.py
@@ -1,0 +1,11 @@
+from django.http import HttpResponse
+
+SECURITY_TXT = """\
+OpenBugBounty: https://openbugbounty.org/bugbounty/ucbocf/
+"""
+
+
+def openbugbounty_security(request):
+    """Serve the OpenBugBounty security.txt file."""
+
+    return HttpResponse(SECURITY_TXT, content_type='text/plain')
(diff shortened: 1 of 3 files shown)


ocfweb/main/openbugbounty.py
```diff
@@ -0,0 +1,11 @@
+from django.http import HttpResponse
+
+SECURITY_TXT = """\
+OpenBugBounty: https://openbugbounty.org/bugbounty/ucbocf/
+"""
```
chriskuehl: can we also include something like:

```
Contact: security@ocf.berkeley.edu
```

in case people would like to casually report issues without using OpenBugBounty? it's good to have OpenBugBounty as an option, but presumably we don't want to make people jump through hoops if they'd rather just send a message.

ocfweb/main/openbugbounty.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def openbugbounty_security(request):
+    """Serve the OpenBugBounty security.txt file."""
+
```
chriskuehl: why empty line?

ocfweb/main/openbugbounty.py
```diff
@@ -0,0 +1,11 @@
+from django.http import HttpResponse
```
chriskuehl: can we update the naming here (the module and function), maybe to `security_dot_txt` or something instead? `security.txt` isn't an openbugbounty thing.

https://securitytxt.org/

kpengboy:
@chriskuehl Thanks for the feedback.

PR #389, opened 2018-08-19
Upgrade jQuery from 1.12.4 to 3.3.1

jvperrin:
Seems to work fine, I used [jquery-migrate](https://github.com/jquery/jquery-migrate) to make sure that any deprecation warnings were resolved (hence all the changes from stuff like `click(function() {})` to `on('click', function() {})`). We also (thankfully) don't have that much JS to test, so that's nice.


jvperrin:
In the course of investigating this, I also discovered that on the mail virtual hosting page, the edit button under all the vhost email addresses doesn't actually do anything currently, so to change the source email, the whole thing needs to be recreated with the new email.

PR #391, opened 2018-08-28
Update Puppet logo on lab about page.

redplanks:


redplanks:
Join staff

Lab desktop homepage

PR #395, opened 2018-09-09
quick wording fix on staff_hours.html

MrMinos:
Simple wording fix; no humans would ever have been confused by this and be paralyzed but some humans would have had to read the sentence twice, myself included. 

This gets rid of that mild inconvenience.

Previous sentence "...so it's a good idea to check this page before coming in for cancellations." could be read as the person is coming to the lab for cancellations rather than staff hour.

The new sentence "...so it's a good idea to check this page for cancellations before coming in to the lab." should be more clear.


keur:
Thanks

MrMinos:
Woops, apologies for merging ahead. I'll keep in mind next time.

PR #399, opened 2018-10-09
update provisioning instructions for new puppetserver ca syntax

abizer:


keur:
Thank you. And thank you to the puppet team. As you can plainly see, it was _necessary_ for them to cut support for the `puppet cert` command

kpengboy:
(And the `puppet ca` command before that)

PR #404, opened 2018-10-18
Upgrade pre-commit hooks, add a few additional hooks

jvperrin:
The title pretty much explains it, this upgrades the existing hooks, and then adds a few more hooks from the standard hooks repo (and fixes some things caught by the hooks).


dkess:
Nice changes, though is there any way around referring to commit hashes in those links? I'm worried that they'll fall out of date any time we change these files...

PR #406, opened 2018-10-19
Auto-update ocf.io/servers

ziyaointl:
Updated servers.py to dynamically fetch servers from LDAP.
Hypervisor/vm relationships are still hard coded, but should be easy to auto-update once we figure out how to run puppet queries in python.

Discourse link: https://leprosy.ocf.berkeley.edu/forum/t/auto-update-ocf-io-servers/162

### DIFF
--- a/ocfweb/docs/views/servers.py
+++ b/ocfweb/docs/views/servers.py
@@ -7,7 +7,7 @@ from cached_property import cached_property
 from django.shortcuts import render
 from ocflib.infra.hosts import hosts_by_filter
 
-from ocfweb.caching import periodic
+from ocfweb.caching import cache
 
 PUPPETDB_URL = 'https://puppetdb:8081/pdb/query/v4'
 PUPPET_CERT_DIR = '/etc/ocfweb/puppet-certs'
@@ -108,7 +108,7 @@ def ldap_to_host(item):
     return hostname, Host(hostname, item['type'], description, ())
 
 
-@periodic(300)
+@cache()
 def get_hosts():
     ldap_output = hosts_by_filter('(|(type=server)(type=desktop)(type=printer))')
     servers = dict(ldap_to_host(item) for item in ldap_output if not is_hidden(item))


ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,73 @@ def has_munin(self):
...
...a", "gnats", "leprosy", "malaria", "nuke", "oilspill", "panic", "pileup", "riot", "sarin", "shipwreck", "virus", "vortex", "war", "zerg", "walpurgisnacht", "mudslide", "pompeii", "locusts", "fire", "rapture", "smallpox", "aliens", "cloudburst", "coldwave", "coma", "dev-firestorm", "emp", "meltdown", "meteorite", "old-vampires", "quasar", "ragnarok", "revolution", "sauron", "skynet", "tempest" ], "certname": "pandemic.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [ "nyx", "dementors", "democracy", "whiteout", "reaper", "segfault", "monsoon", "fraud", "gridlock", "lethe", "anthrax", "supernova", "firestorm", "pestilence", "flood", "maelstrom", "thunder", "death", "tsunami", "lightning", "vampires", "werewolves", "biohazard", "cataclysm", "dev-dementors", "dev-flood", "dev-whiteout", "doom", "hozer-69", "hozer-70", "hozer-71", "hozer-73", "hozer-74", "limniceruption", "matrix", "miasma", "zombies" ], "certname": "riptide.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [], "certname": "gnats.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [], "certname": "smallpox.ocf.berkeley.edu" } ]')  # noqa: E501
```
abizer:
```suggestion
    # XXX: FOR TESTING PURPOSES ONLY, REMOVE THIS BEFORE MERGING
    puppet_query_output = eval('[ { "name": "vms", "environment": "abizer_aux", "value": [], "certname": "jaws.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [ "fallingrocks", "cataclysm", "dev-flood", "dev-maelstrom", "dev-pestilence", "dev-tsunami", "dev-werewolves", "dev-whiteout", "doom", "fraud", "hozer-62", "limniceruption", "lowgpa", "miasma" ], "certname": "hal.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [], "certname": "mudslide.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [ "pox", "alamo", "apocalypse", "blackrain", "fallout", "falsevacuum", "fireball", "fukushima", "gnats", "leprosy", "malaria", "nuke", "oilspill", "panic", "pileup", "riot", "sarin", "shipwreck", "virus", "vortex", "war", "zerg", "walpurgisnacht", "mudslide", "pompeii", "locusts", "fire", "rapture", "smallpox", "aliens", "cloudburst", "coldwave", "coma", "dev-firestorm", "emp", "meltdown", "meteorite", "old-vampires", "quasar", "ragnarok", "revolution", "sauron", "skynet", "tempest" ], "certname": "pandemic.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [ "nyx", "dementors", "democracy", "whiteout", "reaper", "segfault", "monsoon", "fraud", "gridlock", "lethe", "anthrax", "supernova", "firestorm", "pestilence", "flood", "maelstrom", "thunder", "death", "tsunami", "lightning", "vampires", "werewolves", "biohazard", "cataclysm", "dev-dementors", "dev-flood", "dev-whiteout", "doom", "hozer-69", "hozer-70", "hozer-71", "hozer-73", "hozer-74", "limniceruption", "matrix", "miasma", "zombies" ], "certname": "riptide.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [], "certname": "gnats.ocf.berkeley.edu" }, { "name": "vms", "environment": "production", "value": [], "certname": "smallpox.ocf.berkeley.edu" } ]')  # noqa: E501
```

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,73 @@ def has_munin(self):
...
+        """Pop Host with hostname from servers, change its type, and add to misc
+        """
+        servers.pop(hostname, host_type)
+        type_dict[hostname] = Host.from_ldap(hostname, type=host_type)
+    change_host_type('overheat', 'raspi', misc)
+    change_host_type('tornado', 'nuc', misc)
+    change_host_type('jaws', 'hypervisor', hypervisors)
+    misc['blackhole'] = Host(
+        'blackhole', 'network',
+        'Managed Cisco Catalyst 2960S-48TS-L Switch.', [],
```
abizer: now an Arista 7050SX

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,74 @@ def has_munin(self):
...
+    )
+
+    return list(hypervisors.values()) + list(servers.values()) \
+        + list(desktops.values()) + list(misc.values())
```
kpengboy: NB: it's generally preferred to use parentheses instead of backslashes when possible, i.e.
```python
return (list(hypervisors.values()) + list(servers.values())
        + list(desktops.values()) + list(misc.values()))
```

You can alternatively write this line more compactly using unpacking:
```python
return [*hypervisors.values(), *servers.values(), *desktop.values(), *misc.values()]
```

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,74 @@ def has_munin(self):
...
+
+    servers = create_hosts(hosts_by_filter('(type=server)'))
+    desktops = create_hosts(hosts_by_filter('(type=desktop)'))
+    misc = create_hosts(hosts_by_filter('(type=printer)'))
+    hypervisors = {}
+
+    # Add children to hypervisors
+    for h in list(servers.values()):
+        children = []
+        for child_hostname in get_children(h.hostname):
```
kpengboy: This will cause you to query PuppetDB once for every server in the servers list. That's a lot of queries - slow!

You could fix this by implementing some sort of caching for `get_children`. But an even better way would be to rearchitect this code so it queries PuppetDB only once, gets all the information it needs, and then processes the response to extract the needed information.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,74 @@ def has_munin(self):
...
-        Host('pagefault', 'printer', '', []),
-        Host('papercut', 'printer', '', []),
-        Host.from_ldap('overheat', type='raspi'),
-        Host.from_ldap('tornado', type='nuc'),
-
-        *(Host.from_ldap(desktop, type='desktop')
-          for desktop in sorted(list_desktops())),
-    ]
+    def is_hidden(host):
+        return host['cn'][0][:6] == 'hozer-' or host['cn'][0][:4] == 'dev-'
```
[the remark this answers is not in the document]
dkess: We want to keep staffvms hidden, so they should be excluded here.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
         return self.type in ('hypervisor', 'vm', 'server', 'desktop')
 
 
+def is_hidden(host):
+    return host['cn'][0].startswith('hozer-') or host['cn'][0].startswith('dev-')
```
kpengboy: Again, this also needs to check whether a host is a staff VM (`(type=staffvm)`) and hide it if so.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
...
-        Host('logjam', 'printer', '', []),
-        Host('pagefault', 'printer', '', []),
-        Host('papercut', 'printer', '', []),
-        Host.from_ldap('overheat', type='raspi'),
-        Host.from_ldap('tornado', type='nuc'),
-
-        *(Host.from_ldap(desktop, type='desktop')
-          for desktop in sorted(list_desktops())),
-    ]
+    def create_hosts(lst):
```
kpengboy: N.B.: I was actually suggesting you move this function outside the function body, rather than the previous one

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
         return self.type in ('hypervisor', 'vm', 'server', 'desktop')
 
 
+def is_hidden(host):
+    return host['cn'][0].startswith('hozer-') or host['cn'][0].startswith('dev-')
+
+
+TEMP_GET_CHILDREN = "facts { name = 'vms' }"
```
kpengboy: There's no need to name these constants `TEMP`. (After all, they're not temporary.) Rather, I would suggest a name like `PQL_GET_CHILDREN`.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
...
+                hostname = h['cn'][0]
+                hosts[hostname] = (Host(hostname, h['type'], description, ()))
+        return hosts
+
+    servers = create_hosts(hosts_by_filter('(type=server)'))
+    desktops = create_hosts(hosts_by_filter('(type=desktop)'))
+    misc = create_hosts(hosts_by_filter('(type=printer)'))
+    hypervisors = {}
+
+    physical_hosts = query_puppet(TEMP_IS_VIRT_FALSE)
```
kpengboy: I've found that you can directly get a list of hypervisors with the PQL query `resources[certname] { type = "Class" and title = "Ocf_kvm" }`. (Note that the capital O is not a typo.)

What this query does is that it checks whether a node has included the `ocf_kvm` Puppet class, which is the class that makes a given server a hypervisor. Nodes include `ocf_kvm` based on their configuration in hieradata. For example, [here](https://github.com/ocf/puppet/blob/33e2e00a32cd6f3d44940a6a0b6cf89e0e280c49/hieradata/nodes/jaws.yaml#L2) is the specific item that tells `jaws` to include `ocf_kvm` and thus make itself a hypervisor. If you have some free time, I encourage you to browse other files in the same directory to see which other nodes have that class included and which ones do not.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
...
+    all_children = query_puppet(TEMP_GET_CHILDREN)
+
+    # Add children to hypervisors
+    for h in list(servers.values()):
+        if h.hostname in physical_hosts:
+            # Populate a list of children
+            children = []
+            for child_hostname in all_children.get(h.hostname, []):
+                child = servers.get(child_hostname, False)
```
kpengboy: It's slightly more idiomatic to just say `child = servers.get(child_hostname)` here. If there is no entry with key `child_hostname`, then `child` will be set to `None`, which is false in boolean context.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
...
+                h.description,
+                children,
+            )
+
+    # Handle special cases
+    def change_host_type(hostname, host_type, origin_type_dict, target_type_dict):
+        """Pop Host with hostname from servers, change its type, and add to misc
+        """
+        del origin_type_dict[hostname]
+        target_type_dict[hostname] = Host.from_ldap(hostname, type=host_type)
```
kpengboy: I would suggest the following pattern:

```python
host = origin_type_dict.pop(hostname)
target_type_dict[hostname] = host._replace(type=host_type)
```

It would avoid making an extra LDAP query.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +56,90 @@ def has_munin(self):
...
+
+    change_host_type('overheat', 'raspi', servers, misc)
+    change_host_type('tornado', 'nuc', hypervisors, misc)
+    change_host_type('ransomware', 'hypervisor', servers, hypervisors)
+    change_host_type('dataloss', 'server', hypervisors, servers)
+    change_host_type('corruption', 'server', hypervisors, servers)
+    misc['blackhole'] = Host(
+        'blackhole', 'network',
+        'Arista 7050SX Switch.', [],
```
kpengboy: Should not have a period

kpengboy: Actually never mind

ocfweb/docs/views/servers.py
ziyaointl: Looks like hosts that have type 'staffvm' are not included in the '(type=server)' query. Is it possible that each host only has one 'type' value? If that's the case, then maybe we don't need to filter staffvms.

kpengboy: You're right about that, actually. Never mind.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +57,89 @@ def has_munin(self):
...
+    servers = create_hosts(hosts_by_filter('(type=server)'))
+    desktops = create_hosts(hosts_by_filter('(type=desktop)'))
+    misc = create_hosts(hosts_by_filter('(type=printer)'))
+    hypervisors = {}
+
+    hypervisor_hostnames = query_puppet(PQL_IS_HYPERVISOR)
+    all_children = query_puppet(PQL_GET_CHILDREN)
+
+    # Add children to hypervisors
+    for h in list(servers.values()):
```
dkess: No need to convert this to a list, Python will know to loop through the set of values.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +57,89 @@ def has_munin(self):
...
+        if h.hostname in hypervisor_hostnames:
+            # Populate a list of children
+            children = []
+            for child_hostname in all_children.get(h.hostname, []):
+                child = servers.get(child_hostname)
+                if child:
+                    del servers[child.hostname]
+                    children.append(Host(child.hostname, 'vm', child.description, ()))
```
dkess: You shouldn't need to construct a new `Host` object here, it should be fine to just append `child`.

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +57,89 @@ def has_munin(self):
...
+
+    # Handle special cases
+    def change_host_type(hostname, host_type, origin_type_dict, target_type_dict):
+        """Pop Host with hostname from servers, change its type, and add to misc
+        """
+        host = origin_type_dict.pop(hostname)
+        target_type_dict[hostname] = host._replace(type=host_type)
+
+    change_host_type('overheat', 'raspi', servers, misc)
+    change_host_type('tornado', 'nuc', servers, misc)
```
dkess: Nested functions are messy, and this one has a lot of arguments which makes it even more messy. I think the best way to handle this would be at the beginning, when you first interpret the `hosts_by_filter` output.

ocfweb/docs/views/servers.py
ziyaointl: The list acts as a copy of the elements in servers.values() since the dictionary servers is modified during iteration.

ocfweb/docs/views/servers.py
ziyaointl: A new `Host` object is constructed for the sole purpose of changing its type from a physical `server` to `vm` given that they are not differentiated in LDAP listings.

ziyaointl: Instead of creating a new `Host` object, maybe I could call the `_replace` method?

ocfweb/docs/views/servers.py
```diff
@@ -57,81 +57,89 @@ def has_munin(self):
...
-        Host('papercut', 'printer', '', []),
-        Host.from_ldap('overheat', type='raspi'),
-        Host.from_ldap('tornado', type='nuc'),
-
-        *(Host.from_ldap(desktop, type='desktop')
-          for desktop in sorted(list_desktops())),
-    ]
+    servers = create_hosts(hosts_by_filter('(type=server)'))
+    desktops = create_hosts(hosts_by_filter('(type=desktop)'))
+    misc = create_hosts(hosts_by_filter('(type=printer)'))
```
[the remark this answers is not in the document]
ziyaointl: I consolidated the queries. Please see my latest commit. I am not sure if I should get rid of the function to interpret the output of hosts_by_filter though, since I do need to filter out hidden hosts and convert the LDAP output to `Host` objects.

ocfweb/docs/views/servers.py
dkess: `_replace` will still return a new `Host` object (`namedtuple`s are immutable), but yes you should use that instead.

ocfweb/docs/views/servers.py
```diff
@@ -125,30 +130,16 @@ def get_hosts():
...
                     children.append(Host(child.hostname, 'vm', child.description, ()))
-            # Associate host with its children and move it to the hypervisors dictionary
+            # Associate host with its children and specify type
             del servers[h.hostname]
-            hypervisors[h.hostname] = Host(
+            servers[h.hostname] = Host(
                 h.hostname,
                 'hypervisor',
                 h.description,
                 children,
```
dkess: You should specify keyword arguments here instead of relying on order. That way if the `Host` object ever changes, you won't put the wrong fields in the wrong place.

ocfweb/docs/views/servers.py
```diff
@@ -125,30 +130,16 @@ def get_hosts():
...
-
-    change_host_type('overheat', 'raspi', servers, misc)
-    change_host_type('tornado', 'nuc', servers, misc)
-    misc['blackhole'] = Host(
-        'blackhole', 'network',
-        'Arista 7050SX Switch.', [],
-    )
-
-    return [*hypervisors.values(), *servers.values(), *desktops.values(), *misc.values()]
+    return [*servers.values()]
```
dkess: I'd use the `list` function here.

ocfweb/docs/views/servers.py
```diff
@@ -139,7 +139,18 @@ def get_hosts():
                 children,
             )
 
-    return [*servers.values()]
+    def host_key(h):
```
dkess: This doesn't have to be a nested function, you should pull it out.

dkess:
Changes look great! Just have some minor style nits. You should also make sure to resolve the merge conflicts.

dkess:
One more thing-- I've created a PR that includes the Puppet cert files in the ocfweb development environment (which includes supernova). This way, everyone can work on developing this page without worrying about where to put the certs. See https://github.com/ocf/puppet/pull/473 for this PR.

Instead of getting certs from the `/etc/ocfweb` folder, you should use `/etc/ocfweb/puppet-certs`.

ziyaointl:
@dkess Thanks for all the code suggestions! They are very helpful. ocf/puppet#473 is a welcoming change. I'll update the directory in the next commit.

dkess:
OK, one final thing and this PR will be good to go!

I ran this for myself and noticed that "death" and "tsunami" are listed as physical servers, even though they're actually VMs on riptide. This is happening because during the stretch upgrade 2 months ago, we forgot to rename them from "dev-death" to "death" and so on. So virsh (the puppet facts) reports death as having name "dev-death" even though its actual hostname is death. Same with tsunami.

You should hardcode these in as special exceptions. We'll be able to rename these machines in virsh next time they get rebooted.

jvperrin:
Yeah, sorry about that, should have renamed them as part of the upgrade/migration. You can also just leave it for now and it'll get fixed as part of the migration, it's not a big deal if they are temporarily shown as physical servers.

ocfweb/docs/views/servers.py
```diff
@@ -57,95 +62,96 @@ def has_munin(self):
...
+    hosts = {}
+    for h in lst:
+        if not is_hidden(h):
+            description = h.get('description', [''])[0]
+            hostname = h['cn'][0]
+            hosts[hostname] = (Host(hostname, h['type'], description, ()))
+    return hosts
+
+
+def host_key(h):
```
jvperrin: Rather than doing this, I'd just implement the __lt__(self, other_host) method in the Host class and then the sorting will work just fine without needing to specify a key on a sort.

ocfweb/docs/views/servers.py
```diff
@@ -61,6 +61,18 @@ def english_type(self):
...
         return self.type in ('hypervisor', 'vm', 'server', 'desktop')
 
+    def __key(self):
+        if self.type == 'hypervisor':
+            return 'a' + self.hostname
+        if self.type == 'server':
+            return 'b' + self.hostname
+        if self.type == 'desktop':
+            return 'z' + self.hostname
+        return 'c' + self.type + self.hostname
```
dkess: I'd use a tuple here, so instead of `'a' + hostname`, it would be `(1, hostname)`. Makes it more clear that this is for sorting.

Also, I'd use if/elif/else here, so all of the return statements are at the same level of indentation.

ocfweb/docs/views/servers.py
dkess: Hmm, I agree that this is a bit weird. Not sure if making a `__key` method is necessary, it might just be fine to make it a regular method (without the underscores). @ethanhs, thoughts on what best practice would be here?

dkess:
I've fixed the Jenkins problems, so automated tests work now.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +61,97 @@ def english_type(self):
...
+
+    def __lt__(self, other_host):
+        return self.__key() < other_host.__key()
+
+
+def is_hidden(host):
+    return host['cn'][0].startswith('hozer-') or host['cn'][0].startswith('dev-')
+
+
+PQL_GET_CHILDREN = "facts { name = 'vms' }"
```
abizer: `PQL_GET_VMS`

ocfweb/docs/views/servers.py
```diff
@@ -15,7 +17,7 @@ class Host(namedtuple('Host', ['hostname', 'type', 'description', 'children'])):
```
abizer: remove the TODO and `from_ldap` method if it's no longer being used

ocfweb/docs/views/servers.py
abizer: we don't really need to hide either of these types - they're useful for us as well to tell where the machines are located when we're trying to find them instead of guessing, sshing into every hypervisor, or going to puppetboard.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +61,97 @@ def english_type(self):
...
+    """
+    result = {}
+    for d in output:
+        hostname = d['certname'].split('.')[0]
+        result[hostname] = d.get('value')
+    return result
+
 
-@periodic(120)
+def create_hosts(lst):
```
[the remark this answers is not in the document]
abizer: similar to above, it would be preferable if this function operated on single items rather than a list

abizer:
another thought, this change introduces a dependency on puppetdb for ocfweb. what is the failure model if puppetdb is down or unresponsive? given how relatively infrequently we move around VMs, the output of a successful call should probably get cached forever rather than periodically refreshed (ocfweb is redeployed every night anyways) and if puppetdb isn't available, the entire page should fail gracefully rather than 502ing and spitting ocfweb failure emails at us.

ocfweb/docs/views/servers.py
```diff
@@ -29,13 +31,16 @@ def from_ldap(cls, hostname, type='vm', children=()):
 
     @cached_property
     def ipv4(self):
-        return str(dns.resolver.query(self.hostname, 'A')[0])
+        try:
+            return str(dns.resolver.query(self.hostname, 'A')[0])
+        except dns.resolver.NXDOMAIN:
+            return 'No IPv4 Address'
```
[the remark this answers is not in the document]
dkess: Yeah, I thought the same thing. Perhaps the method name should have an underscore before it so it's clear that this shouldn't be used anywhere else, but that can be outside of this pr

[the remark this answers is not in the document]
dkess: Yeah, I thought the same thing. Perhaps the method name should have an underscore before it so it's clear that this shouldn't be used anywhere else, but that can be outside of this pr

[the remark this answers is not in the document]
dkess: Yeah, I thought the same thing. Perhaps the method name should have an underscore before it so it's clear that this shouldn't be used anywhere else, but that can be outside of this pr

ocfweb/docs/views/servers.py
keur: This feels really hacky. I would not use a tuple either. I think a class to abstract this away would make it more simple.

```python
from enum import Enum

class Machine(Enum):
    HYPERVISOR = 1 
    SERVER = 2 
    OTHER = 3 
    DESKTOP = 4 

    @classmethod
    def from_string(cls, machine):
        if machine == 'hypervisor':
            return Machine.HYPERVISOR
        elif machine == 'server':
            return Machine.SERVER
        elif machine == 'desktop':
            return Machine.DESKTOP
        else:
            return Machine.OTHER

    def __lt__(self, other):
        return self.__class__ is other.__class__ and self.value < other.value
```

Now you can just do `Machine.from_string(self.type) < Machine.from_string(other_host.type)` in `__lt__`.

kpengboy: @keur what you propose could be done more succinctly with a dictionary mapping type strings to numbers. Additionally, you would have to teach `__lt__` to compare type strings and hostnames as well as the enum value. Which makes the final solution about equal in complexity.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +61,97 @@ def english_type(self):
...
+                    children.append(child._replace(type='vm'))
+            # Associate host with its children and specify type
+            del servers[h.hostname]
+            servers[h.hostname] = Host(
+                hostname=h.hostname,
+                type='hypervisor',
+                description=h.description,
+                children=children,
+            )
+    return sorted(servers.values())
```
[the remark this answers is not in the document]
ziyaointl: If we don't mutate the dictionary here we would end up with two extra lists: one corresponding to the `servers` dictionary in the code snippet you posted and another list to keep track of the vms we've processed so far so we won't add them twice. After all, we don't want to just display hypervisors and their vms but also other types of machines. I agree that mutating dictionaries is ugly, but I think the alternative would require too much more code and is less efficient.

ocfweb/docs/views/servers.py
ziyaointl: After coding up different approaches I found this one being possibly the easiest to understand. It's a combination of using a dictionary to store the order and using tuples as keys.
https://github.com/ocf/ocfweb/pull/406/commits/c83f8c81b703657e01f800cacd84cc6f80d2451a

ocfweb/docs/views/servers.py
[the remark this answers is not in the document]
abizer: mutating the dictionary is a classic case of premature optimization. the performance and efficiency implications of the additional list are microscopic (especially compared to the amount of CPU time it will take to query puppetdb and interpret the output), and it will probably get garbage collected anyways once the declaration goes out of scope. The code in the snippet I posted should be all that is necessary to implement this feature. Please reconsider the mutation approach.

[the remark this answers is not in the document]
ethanhs: Mutation is also a lot harder to reason through. So +1 to abizer's suggestion.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +57,93 @@ def english_type(self):
...
 
+    def __key(self):
+        """Key function used for comparison."""
+        ORDER = {
+            'hypervisor': 1,
+            'server': 2,
+            'desktop': float('inf'),
+        }
+        DEFAULT_ORDER = 3
+        return (ORDER.get(self.type, DEFAULT_ORDER), self.type, self.hostname)
```
abizer: all upper case is a style for global constants. leave this as `ranking` and `default` or something like that.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +57,93 @@ def english_type(self):
...
+PQL_GET_VMS = "facts { name = 'vms' }"
+PQL_IS_HYPERVISOR = 'resources[certname] { type = "Class" and title = "Ocf_kvm" }'
 
-        Host.from_ldap(
-            hostname='dev-fallingrocks',
-            type='server',
+
+def query_puppet(query):
+    """Accepts a PQL query, returns a parsed json result."""
+    PUPPETDB_URL = 'https://puppetdb:8081/pdb/query/v4'
```
abizer: move these to the top of the file. they are effectively magic constants.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +57,93 @@ def english_type(self):
...
+    PUPPET_CERT_DIR = '/etc/ocfweb/puppet-certs'
+    r = get(
```
abizer: `requests.get`

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +57,93 @@ def english_type(self):
...
+        PUPPETDB_URL,
+        cert=(
+            os.path.join(PUPPET_CERT_DIR, 'puppet-cert.pem'),
+            os.path.join(PUPPET_CERT_DIR, 'puppet-private.pem'),
```
abizer: change the import at the top of the file to `os`

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +57,93 @@ def english_type(self):
...
         ),
+        verify=os.path.join(PUPPET_CERT_DIR, 'puppet-ca.pem'),
+        params={'query': query},
+    )
+    return r.json() if r.status_code == 200 else None
+
+
+def format_query_output(item):
+    """Converts an item of a puppet query to tuple(hostname, query_value).
+    """
```
abizer: fix docstring formatting

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +61,97 @@ def english_type(self):
...
+        hostname = d['certname'].split('.')[0]
+        result[hostname] = d.get('value')
+    return result
+
 
-@periodic(120)
+def create_hosts(lst):
+    """Accepts a list of raw ldap output, returns a dictionary of host
+    objects indexed by hostname
+    """
```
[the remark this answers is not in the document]
abizer: if the docstring is just one sentence, the terminating triple quote should be on the same line

ocfweb/docs/views/servers.py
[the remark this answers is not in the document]
kpengboy: @abizer does your code correctly handle servers which are neither hypervisors nor VMs nor special cases?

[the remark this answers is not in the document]
abizer: what servers would those be? I have not tested that snippet, obviously, but I think it should cover the cases we have.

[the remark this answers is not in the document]
ziyaointl: @abizer For now I think we have Physical Servers and Printers that are in that category.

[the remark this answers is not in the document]
abizer: non-hypervisor physical machines are special cases, as are the three printers. They should be handled as such. Physical machines that aren't hypervisors (i.e. expressing the ocf_kvm class) won't even show up in the hypervisor query anyways, so it shouldn't be a problem.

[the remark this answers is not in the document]
ziyaointl: @abizer Here's a snippet implementing the special case handling you mentioned.
```python
hypervisors_hostnames = dict(format_query_output(item) for item in query_puppet(PQL_IS_HYPERVISOR))
all_children = dict(format_query_output(item) for item in query_puppet(PQL_GET_VMS))

hostnames_seen = []
servers_to_display = []
# Add children to hypervisors
for hypervisor_hostname in hypervisors_hostnames:
    children = []
    for child_hostname in all_children.get(hypervisor_hostname, []):
        child = servers.get(child_hostname)
        if child:
            children.append(child._replace(type='vm'))
            hostnames_seen.append(child.hostname)
    description = servers[hypervisor_hostname].description if hypervisor_hostname in servers else None
    servers_to_display.append(Host(
        hostname=hypervisor_hostname,
        type='hypervisor',
        description=description,
        children=children,
    ))
    hostnames_seen.append(hypervisor_hostname)

# Handle special cases
for host in servers.values():
    if (host.hostname not in hostnames_seen):
        servers_to_display.append(host)

servers_to_display.extend([
    Host(
        'blackhole', 'network',
        'Arista 7050SX Switch.', [],
    ),
    servers['overheat']._replace(type='raspi'),
    servers['tornado']._replace(type='nuc'),
])

return sorted(servers_to_display)
```

ziyaointl:
I ran some manual testing on the server and the new `get_hosts` function should be good to go.
Also, I noticed that test coverage has decreased according to coveralls. Do I have to write a few unit tests in this case?

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +60,96 @@ def english_type(self):
...
-    ]
+@periodic(300)
+def get_hosts():
+    ldap_output = hosts_by_filter('(|(type=server)(type=desktop)(type=printer))')
+    servers = dict(ldap_to_host(item) for item in ldap_output if not is_hidden(item))
+
+    hypervisors_hostnames = dict(format_query_output(item) for item in query_puppet(PQL_IS_HYPERVISOR))
+    all_children = dict(format_query_output(item) for item in query_puppet(PQL_GET_VMS))
+
+    hostnames_seen = []
```
kpengboy: I would use a set for this variable:

```python
hostnames_seen = set()
```

A set allows you to test membership in constant time instead of linear time.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +60,96 @@ def english_type(self):
...
+            hostname=hypervisor_hostname,
+            type='hypervisor',
+            description=description,
+            children=children,
+        ))
+        hostnames_seen.append(hypervisor_hostname)
+
+    # Handle special cases
+    for host in servers.values():
+        if (host.hostname not in hostnames_seen):
```
kpengboy: The parentheses in this line aren't necessary.

ocfweb/docs/views/servers.py
```diff
@@ -56,96 +60,96 @@ def english_type(self):
...
+            servers_to_display.append(host)
+
+    servers_to_display.extend([
+        Host(
+            'blackhole', 'network',
+            'Arista 7050SX Switch.', [],
```
kpengboy: I would recommend writing out the field names as you've done above, i.e.

```python
Host(
    hostname='blackhole',
    type='network',
    description='Arista 7050SX Switch.',
    children=[],
)
```

ocfweb/docs/views/servers.py
```diff
@@ -113,36 +113,43 @@ def get_hosts():
...
-            )
-    return sorted(servers.values())
+    for hypervisor_hostname in hypervisors_hostnames:
+        children = []
+        for child_hostname in all_children.get(hypervisor_hostname, []):
+            child = servers.get(child_hostname)
+            if child:
+                children.append(child._replace(type='vm'))
+                hostnames_seen.append(child.hostname)
+        description = servers[hypervisor_hostname].description if hypervisor_hostname in servers else None
```
dkess: `servers.get(hypervisor_hostname)`?

dkess:
I wouldn't worry about the coveralls decrease.

ocfweb/docs/views/servers.py
kpengboy: Well you'd still have to check for None and then get the `description` attribute if not None.

dkess: Ohhhh I see, I guess this is fine then

ocfweb/docs/views/servers.py
```diff
@@ -108,7 +108,7 @@ def ldap_to_host(item):
     return hostname, Host(hostname, item['type'], description, ())
 
 
-@periodic(300)
+@cache()
```
abizer: the parens aren't necessary

abizer:
before this gets merged: make sure you squash the commits into a logical series of changes, as opposed to merging the 42 commits directly as they stand.

ocfweb/docs/views/servers.py
ziyaointl: Deleting the parens generates TypeError: outer() missing 1 required positional argument: 'fn'

ziyaointl:
@abizer Is it okay to just squash this PR into one commit?

dkess:
I've enabled "squash and merge" for this repo so you can just do that.

ziyaointl:
Back when I committed my changes, the suite `test_available_domain` failed. Is there a way of restarting tests on Jenkins?

jvperrin:
I've restarted the tests. That one in particular appears to be flaky due to it depending on DNS which is unfortunate.

ziyaointl:
@abizer Could you approve this PR? Github shows me that you have requested changes despite all of your comments are marked resolved, which is kind of weird.

abizer:
congratulations on merging a pretty significant change to master!

PR #409, opened 2018-10-30
Add lab survey page

glee-b:
it is real now

### DIFF
--- a/ocfweb/about/templates/about/lab-survey.html
+++ b/ocfweb/about/templates/about/lab-survey.html
@@ -5,8 +5,8 @@
 {% block container %}
     <div class="container ocf-lab-content">
         <div class="row lab-intro">
-            <h3>Fill out the lab feedback survey!</h3>
-            <h2>This process takes less than 10 minutes!</h2>
+            <h3>Please fill out the lab feedback survey!</h3>
+            <h2>This process takes about 5 minutes.</h2>
         <br>
         <br>
         <p class="learnmore">


dkess:
Screenshot of what it looks like?

glee-:
https://i.fluffy.cc/jCcsSTdDMprrfv7wPmgQCWTGJ2gjjHlq.png
https://i.fluffy.cc/5c6CznJ08kgc3x343JGz5X94dkBrJCBJ.png

It's one long page that is to be scrolled down

glee-:
There's actually a small visual bug to do with how the survey is viewed as you click through it. It appears that a fix is not possible: https://stackoverflow.com/q/19009849/

abizer:
add a "please" to the main call to action, perhaps rewrite the second line because the average user session is like 5 minutes, no one is going to stick around for 10 minutes to fill out the survey if we tell them it's going to take that long. and do we really need to ask these questions about gender and transgender identification and such? if we don't have concrete reasons for asking people to give this kind of personal data about themselves we shouldn't ask for it at all, it's a privacy risk.

asaiacai:
The point for the identity questions is to show that our userbase reaches a wide audience and hopefully it will show that underrepresented people take advantage of our resources.  The form is anonymous, doesn't require people to login so no one's answers can be linked back to them.  Also the reason's for having these questions were discussed at BoD and have been in the ticket for a while now.

abizer:
I must not have been at that BoD meeting. Regardless, we should just be very careful with whatever data comes out of this survey, especially with respect to this kind of private, sensitive demographic data. The only guarantee users have that we won't deanonymize them is the promise that we won't try to correlate submission times with session statistics. We cannot lose or (accidentally) misuse what we don't have.

glee-:
Reworded

PR #415, opened 2018-10-31
Remove survey iframe scrollbar

daradib:
scrolling, like the other iframe attributes, is deprecated
but still supported by all major browsers.


dkess:
This actually breaks sometimes, preventing users from pressing the "next" button since it's not in the visible space. I will have to revert this.

daradib:
@dkess OK, thanks for reverting

PR #421, opened 2018-11-15
Change username validation warning to yellow

dkess:
The current interface doesn't really make it clear that it's OK for one's username to not exactly match our "similarity" filter. I'm open to suggestions about how to word this better, though.

Before:
screenshot_2018-11-14_16-36-51

After:
screenshot_2018-11-14_16-05-57


abizer:
>> when dkessler could instead have been dank@ocf

PR #431, opened 2018-12-16
Reverse list of officers so recent terms are on top

dkess:
This makes the officers page flow more nicely.

cue bikeshedding on why we shouldn't do this


dkess:
I think it makes more sense for the in-code data structure to start with older terms, since it makes sense for lists to have "newer" information at the end. I don't see why the presentation on the website has to match up with that.

PR #433, opened 2018-12-22
Remove "account creation" procedure docs

dkess:
Not sure what this page was ever meant to be for. If no one objects, I'm deleting this.

### DIFF
--- a/ocfweb/docs/docs/staff/procedures/accounts/creation.md
+++ b/ocfweb/docs/docs/staff/procedures/accounts/creation.md
@@ -1,3 +1,0 @@
-[[!meta title="Account creation"]]
-
-TODO: this page fell out of date even more


jvperrin:
I believe it was meant for docs about how the account creation system works, but I can't find past emails or tickets about it. It would still be useful to actually have, but as-is with no content I'd say deleting it is fine.

jvperrin:
Hmm, fair, I'm thinking of a different doc then since I'd agree that wouldn't fit under procedures.

kpengboy:
Long long ago, before account creation was automated, it contained
information on how to create an account:
https://github.com/ocf/wiki-old/blob/8e92307a0141e8613b9d7f8e1868739377dd1c3f/staff/procedures/accounts/creation.mdwn
The information on that page became obsolete so it was removed. Updating it
then became one of those eternal todos.

>

PR #434, opened 2018-12-27
Add prometheus metrics measuring

dkess:
This could help us with investigating the create celery errors, among other things. I assume this is fine, but gonna put this PR here in case there's anything I should be aware of.


dkess:
Here's some example output from a dev environment, where I loaded a view pages: https://i.fluffy.cc/MWQt4Rnq1D0zzMpwlz908twLZ5lC5G9r.html

I don't think anything is sensitive here, but we should still restrict this to be staff-only. I'll look into ways to do this.

PR #438, opened 2018-12-29
Update the Banning Policy

EnnFour:
Updates the banning policy to reflect the language voted on and passed at the BoD meeting on Dec 3, 2018.

### DIFF
--- a/ocfweb/docs/docs/membership/banning.md
+++ b/ocfweb/docs/docs/membership/banning.md
@@ -1,6 +1,6 @@
 [[!meta title="Banning policy"]]
 
-_As established by the Board of Directors on October 31, 2016._
+_As modified by the Board of Directors on December 3, 2018._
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/membership/banning.md
```diff
@@ -1,6 +1,6 @@
 [[!meta title="Banning policy"]]
 
-_As established by the Board of Directors on October 31, 2016._
+_As modified by the Board of Directors on December 3, 2018._
```
abizer: preferably keep both established and amended, e.g.

"As established by the Board... on October 31, 2016 and amended December 3, 2018."

kpengboy: I would be fine on the other hand without the establishment date (c.f. the account policy)

PR #439, opened 2019-01-03
docs: add kubernetes design doc

keur:


ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+master nodes share state to manage the cluster and schedule jobs to run on
+workers. It is considered best practice to run an odd number of masters, and
+currently our cluster has three masters.
+
+### Masters
+
+Kubernetes masters share state via a distributed key-value store. The reason
+for the key-value store is high availability: if one master goes down, a new
+master node will subsume control of the cluster in a leader election to ensure
+that no services incur downtime. [etcd][etcd-io] is an
```
abizer: I'd write this something like the following:

> Kubernetes masters share state via etcd, a distributed key-value store implementing the Raft protocol. One master is elected as a leader for the cluster, with the ability to commit writes to the KVS. etcd then reliably replicates this state across every master, so that if the leader fails, another master can be elected and no state will be lost in the process. The state stored in etcd is scheduling state, service locations, and metadata, not state of the services running on the cluster itself.

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+state, quorum must have _at least_ ⌊*N*/2⌋+1 members. Every new odd number in a
+cluster with *M* > 1 masters adds one more node of fault tolerance.  Therefore,
+adding an extra node to an odd numbered cluster gives us nothing. If interested
+read more [here][failure-tolerance].
+
+### Workers
+
+Workers are the brawn in the Kubernetes cluster. While master nodes are
+constantly sharing data, managing the control plane, and scheduling services,
+workers primarily run [pods][pod]. `kublet` is the service that executes pods
```
abizer: `kubelet`?

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+                                            |              Kubernetes Cluster               |
+                           HAProxy          |                                               |
+                          ----------        |                    Ingress         Pod        |
+                          |autocrat|        |                   ----------    ----------    |
+                          ----------        |                   |Worker1|     | ocfweb |    |
+                                            |                   ----------    ----------    |
+                                            |                                               |
+            *VIP*          HAProxy          |                    Ingress         Pod        |
+        ------------      ----------  ✘ SSL / Host: templates   ---------     -----------   |
+REQ --> |keepalived| -->  |deadlock|    -->   -------------->   |Worker1| --> |Templates|   |
```
abizer: +100 for artwork

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+appropriate pod where the request is finally processed.
+
+
+### Why didn't we use MetalLB?
+
+`MetalLB` was created so a bare-metal Kubernetes cluster could use `Type:
+LoadBalancer` in Service definitions. The problem is, in `L2` mode, it takes a
+pool of IPs and puts your service on a random IP in that pool. How one makes
+DNS work in this configuration is completely unspecified. We would need to
+dynamically update our DNS, which to me sounds like a myraid of outages waiting
```
abizer: myriad

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+forwards control plane traffic to the correct `kublet`.
+
+### Pods
+
+In the Kubernetes world, pods are the smallest computing unit. A pod is made up
+of one or more containers. The difference between a pod and a standalone
+container is best illustrated by an example. Consider [ocfweb][ocfweb]; it is
+comprised of several containers—the web container, static container, and worker
+container.  In Kubernetes, together these three containers form one pod, and it
+is pods that can be scaled up or down. A failure in any of these containers
```
dkess: These don't necessarily have to be in the same pod, and when we actually deploy ocfweb I'd consider putting static files in a separate pod.

I guess you can keep this here for now though, if that ends up happening I'll change it.

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+indicates a failure in the entire pod. It should be noted that when writing
+Kubernetes services we don't actually deal in pods but one further abstraction,
+deployments, which create pods for us.
```
dkess: Link to the `Deployment` docs

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+Do note that `puppetlabs-kubernetes` is still very much a work in progress. If
+you notice an issue in the module you are encouraged to write a patch and send
+it upstream.
+
+### Kubernetes PKI
+
+All the private keys and certs for the PKI are in the puppet private share, in
+`/opt/puppet/shares/private/kubernetes`. We won't go into detail of everything
+contained there, but much of Kubernetes and `etcd` communication is
+authenticated using client certificates. All the necessary items for workers
```
dkess: s/much of/all

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+are included in `os/Debian.yaml`, although adding a new master to the cluster
+requires re-running [kubetool][puppetlabs-kubetool] to generate new `etcd
+server` and `etcd peer` certs.
```
dkess: I'd add that as far as we know, refreshing the etcd certs requires taking down the cluster.

An aside that we may or may not want to document-- I believe it should, in theory, be possible to do all of this live and without downtime, by generating and signing keys using the old CA. We'd have to see exactly what the kubetool script is doing. If we ever want to scale to 5 masters, this is something that could be worth looking into.

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,190 @@
...
+communication. `puppetlabs-kubernetes` supports two choices: `weave` and
+`flannel`. Both solutions work out-the-box, and we've had success with
+`flannel` thus far so we've stuck with it.
+
+### Getting traffic into the cluster
+
+One of the challenges with running Kubernetes on bare-metal is getting traffic
+into the cluster. If you are running on `AWS`, `GCP`, or `Azure` you add `type:
+LoadBalancer` to your Service definitions and things just work. For bare-metal,
+this is an exercise left to the reader.
```
dkess: I don't think this paragraph makes a lot of sense to people new to Kubernetes. Perhaps add some more context on why it's difficult to get traffic into the cluster. The last two sentences are unnecessarily snarky.

ocfweb/docs/docs/staff/backend/kubernetes.md
dkess: I'd try to make it more clear that the ingress that handles the request from haproxy doesn't necessarily run on the same worker with the templates pod.

keur: Ah, okay. I just using the Pod abstraction instead of a worker here, but yeah people new to Kube might not get the implication. I'll have to add it to the text though

keur:
@dkess @abizer thanks for the feedback. I tried to address every comment and force-pushed.

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,204 @@
...
+3. Ensuring log integrity across all masters.
+
+One master is elected as a leader of the cluster. The leader has the ability to
+commit writes to the KVS. `etcd` then reliably replicates this state across
+every master, so that if the leader fails, another master can be elected and no
+state will be lost in the process. Do note that the state stored in `etcd` is
+scheduling state, service locations, and other cluster metadata; it does _not_
+keep state of the services running on the cluster.
+
+### Why the odd number of masters?
```
jvperrin: This seems more like a level 4 header, since it would be a subheader under the level 3 "Masters". Not sure what that actually looks like, but it feels like a subheader anyway

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,204 @@
...
+it upstream.
+
+### Kubernetes PKI
+
+All the private keys and certs for the PKI are in the puppet private share, in
+`/opt/puppet/shares/private/kubernetes`. We won't go into detail of everything
+contained there, but Kubernetes and `etcd` communication is authenticated using
+client certificates. All the necessary items for workers are included in
+`os/Debian.yaml`, although adding a new master to the cluster requires
+re-running [kubetool][puppetlabs-kubetool] to generate new `etcd server` and
```
jvperrin: How is kubetool re-run? Is this manually or automatically, where should it be run from and what arguments should it be given?

(not sure if all of this should be answered here, but I think a bit more detail on actually using kubetool would be good)

ocfweb/docs/docs/staff/backend/kubernetes.md
jvperrin: I think if you just add diagonal arrows in here to make it clear that the request could go to any of the pods from the ingress, then that would be fine. Same with the `keepalived` part, I think diagonal/sideways arrows would be good there to say it can go to any `HAProxy`/master, although I'm not sure if that would be confusing to suggest it's actually being split across masters or not. Maybe `keepalived` shouldn't be a separate step in the flow, but should just reside on one of the master servers? Like:

```
--------------------
|     deadlock:    |
|current keepalived|
|      master      |
--------------------
```

(or something similar)

Also `Worker1` -> `Worker2` on this line, `Worker1` is repeated twice, unless that was intentional because they could actually be scheduled onto the same worker.

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Why don't we copy Marathan and specify one port per service?
+
+In our current Marathon configuration, we give each service a port on the load
+balancer and traffic coming into that port is routed accordingly. First, in
+Kubernetes we would emulate this behavior using `NodePort` services, and all
+Kubernetes documentation discourages this. Second, it's ugly. Every time we add
+a new service we need to modify the load balancer configuration in Puppet. With
+our Kubernetes configuration we can add unlimited HTTP services without
+touching Puppet.
```
jvperrin: Where does HAProxy end up getting its config from instead?

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Why are we using NodePort services?
```
jvperrin: I feel like this doesn't quite fit, like it could be part of the previous section instead, since it's more of a follow-on from that than its own section.

ocfweb/docs/docs/staff/backend/kubernetes.md
```diff
@@ -0,0 +1,204 @@
...
+### OCF Kubernetes Configuration
+
+Currently, the OCF has three Kubernetes masters: (1) `deadlock`, (2) `coup`,
+and (3) `autocrat`. A Container Networking Interface (`cni`) is the last piece
+required for a working cluster. The `cni`'s purpose is to faciltate intra-pod
+communication. `puppetlabs-kubernetes` supports two choices: `weave` and
+`flannel`. Both solutions work out-the-box, and we've had success with
+`flannel` thus far so we've stuck with it.
+
+### Getting traffic into the cluster
```
jvperrin: I'd say maybe this is a level 2 header, since it's not really under bootstrapping any more, more just regular operation, and then the level 3 stuff under it makes sense since it fits under this header.

ocfweb/docs/docs/staff/backend/kubernetes.md
keur: HAProxy is configured in puppet. All it needs to know is where workers running ingress live (right now that is all nodes)

keur: Ingress workers route based on Host headers, which is why our current config only scales for HTTP services. If we wanted to expose TCP services, we would probably need to modify HAProxy

ocfweb/docs/docs/staff/backend/kubernetes.md
keur: The docs I linked explain how to run it (it's a docker container that can be run on any host) and what parameters to pass. I think `kubetool` warrants it's own set of documentation. One thing you pointed out that I missed is that it does need to be run manually.

PR #441, opened 2019-01-05
requirements: update them

abizer:
looks like the major version update is redis? `make dev` works for me, but it's 7am and I haven't slept yet so someone else may want to try this out and see if it all works


requirements-minimal.txt
```diff
@@ -4,6 +4,7 @@ django
 django-bootstrap-form
 django-ipware
 django-mathfilters
+django-prometheus
```
abizer: also this @dkess

abizer:
dkessler sees an empty spot in ~staff/legends near his name and desires to fill it with great breakage, I approve

abizer:
this breaks redis, reverted

PR #442, opened 2019-01-05
Add documentation on offsite backup retention

dkess:


ocfweb/docs/docs/staff/backend/backups.md
```diff
@@ -24,6 +24,10 @@ with only OCF officers to make sure the backups are kept as secure as possible,
 since they contain all of the OCF's important data.  The backups are already
 encrypted, but it doesn't hurt to add a little extra security to that.
 
+### Retention
+
+Off-site backups older than six months (180 days) are permanently deleted.
```
kpengboy: I would recommend mentioning and linking to the `prune-old-backups` script as well

PR #443, opened 2019-01-05
Allow markdown hard line breaks without trailing spaces

jvperrin:
I think the added test and docstring explain what this does pretty well, as well as the [GFM docs](https://github.github.com/gfm/#backslash-escapes) on this. Basically, I added this because a lot of editors people have (mine included) strip trailing whitespace, but it's actually used in markdown for hard line breaks (`<br>` tags), which is unfortunate. Now, with this change, adding a backslash before the newline at the end of a line will add in a line break (just like adding two spaces before the newline did before), but backslashes in the middle of a line will just function as normal.

Thoughts?

I also removed some of the duplication with the HTML comment mixins, since they can just share a class as far as I can tell for both inline and block lexers.

As a side note, backslash escaping with regexes is really annoying, because they need to be escaped in python and also escaped for the regex parser, and using a raw string in this case doesn't work because it contains a literal newline. That took a while to mess around with and figure out anyway.

### DIFF
--- a/ocfweb/component/markdown.py
+++ b/ocfweb/component/markdown.py
@@ -15,8 +15,8 @@ from ocfweb.caching import cache
 META_REGEX = re.compile(r'\[\[!meta ([a-z]+)="([^"]*)"\]\]')
 
 
-class HtmlCommentsInlineLexerMixin:
-    """Strip HTML comments inside lines."""
+class HtmlCommentsLexerMixin:
+    """Strip HTML comments as entire blocks or inside lines."""
 
     def enable_html_comments(self):
         self.rules.html_comment = re.compile(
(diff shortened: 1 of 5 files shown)


dkess:
I guess this is fine, but I'd rather us not write a very custom Markdown implementation. Can we use CommonMark instead? Looks like this is supported: https://commonmark.org/help/tutorial/03-paragraphs.html

jvperrin:
Sure, if we can swap out [mistune](https://github.com/lepture/mistune) with something using CommonMark ([mistletoe](https://github.com/miyuchina/mistletoe) perhaps?) then I'm all for that. But note that we already have some custom features, like stripping HTML comments and a custom format to link between docs. I'm not sure I'd call this very custom, since it's in GFM and as you mentioned, CommonMark.

jvperrin:
I do agree, I think having a library doing this for us would be much better, although the custom doc links are not something that can really be found elsewhere from what I can tell. The code syntax highlighting and the HTML comment stripping also don't appear to be common features, so if we did end up migrating to some other standard, I think we'd still end up needing to reimplement most of these features (not this one in particular since it's included in CommonMark, but the more complicated ones anyway).

jvperrin:
_semi-reluctantly merges_

PR #449, opened 2019-01-20
[WIP] More detailed HPC docs.

bzh-bzh:
Going to add more detailed docs on Singularity and Python workflows later, but figured this was a good start.

### DIFF
--- a/ocfweb/docs/docs/services/hpc.md
+++ b/ocfweb/docs/docs/services/hpc.md
@@ -53,8 +53,9 @@ generously funded by a series of grants from the [Student Tech Fund][stf].
 ## Slurm
 
 We currently use [Slurm][slurm] as our workload manager for the cluster.
-Slurm is a free and open source job scheduler that evenly distributes jobs across an HPC cluster,
-where each computer in the cluster is referred to as a node.
+Slurm is a free and open source job scheduler that evenly distributes jobs
+across an HPC cluster, where each computer in the cluster is referred to
+as a node.
 The only way to access our HPC nodes is through Slurm.
 
 Detailed documentation for how to access Slurm is [[here|doc services/hpc/slurm]].
--- a/ocfweb/docs/docs/services/hpc/slurm.md
+++ b/ocfweb/docs/docs/services/hpc/slurm.md
@@ -44,25 +44,25 @@ allocate more resources and time to your job, you must set one or more of these
 flags:**
 
 * `-n` / `--ntasks`:
-    - The number of tasks/processes to allocate. **Default is 1.**
+    - The number of tasks/processes to allocate. Default is 1.
 * `-c` / `--cpus-per-task`:
-    - The number of CPUs to allocate per task. **Default is 1.**
+    - The number of CPUs to allocate per task. Default is 1.
 * `--mem`:
     - The total amount of RAM to allocate. By default, the number supplied is
 assumed to megabytes. However, the prefixes `K`, `M`, `G`, and `T` can be
 appended to the number instead. For example, to allocate 5 gigabytes of ram,
-use `--mem=5G`. **Default is 100 megabytes.**
+use `--mem=5G`. Default is 100 megabytes.
 * `--gres` **(Optional)**:
     - Allocates some GPUs to your job. The format is `--gres=gpu:[optional
 type]:[number to allocate]`. For example, to allocate 2 GPUs of any type, you
 would include `--gres=gpu:2`. To allocate two Nvidia 1080Ti GPUs (our only type
-right now), you would include `--gres=gpu:nv1080:2`. **Default is no GPUs.**
+right now), you would include `--gres=gpu:nv1080:2`. Default is no GPUs.
 * `--t` / `--time` **(Optional)**:
     - The maximum amount of time your job can take before Slurm forcefully
 kills it. Acceptable time formats include "minutes", "minutes:seconds",
 "hours:minutes:seconds", "days-hours", "days-hours:minutes" and
 "days-hours:minutes:seconds". You cannot set the time limit greater than the
-**default, which is 2 days.**
+default, which is 2 days.
 
 ### Using `srun`
 


ocfweb/docs/docs/services/hpc.md
```diff
@@ -42,7 +42,7 @@ where you can ask questions and talk to us about anything HPC.
 As of Fall 2018, the OCF HPC cluster is composed of one server, with the
 following specifications:
 
-* 2 Intel Xeon [E5-2640v4][corruption-cpu] CPUs (10c/20t @ 2.4GHz)
+* 2 Intel Xeon [E5-2640v4][corruption-cpu] CPUs (40 cores @ 2.4GHz)
```
abizer: I'd revert this since nominally it isn't factually accurate - there's only 20 actual cores, 40 threads with SMT.

redplanks: It isn't accurate, but I wanted people to know what the maximum number of CPUs they could allocate in Slurm was.

ocfweb/docs/docs/services/hpc/slurm.md
```diff
@@ -0,0 +1,142 @@
...
+* `--mem`:
+    - The total amount of RAM to allocate. By default, the number supplied is
+assumed to megabytes. However, the prefixes `K`, `M`, `G`, and `T` can be
+appended to the number instead. For example, to allocate 5 gigabytes of ram,
+use `--mem=5G`. **Default is 100 megabytes.**
+* `--gres` **(Optional)**:
+    - Allocates some GPUs to your job. The format is `--gres=gpu:[optional
+type]:[number to allocate]`. For example, to allocate 2 GPUs of any type, you
+would include `--gres=gpu:2`. To allocate two Nvidia 1080Ti GPUs (our only type
+right now), you would include `--gres=gpu:nv1080:2`. **No default.**
```
kpengboy: I'd say "default is no GPUs"

ocfweb/docs/docs/services/hpc/slurm.md
```diff
@@ -0,0 +1,142 @@
...
+*Some terminology:* Slurm refers to a process as a "task". Even if a single
+process is using multiple threads/CPUs, it still counts as one task.
+
+**By default, without any flags, a job you submit will be allocated one CPU,
+100 MB of RAM, and no GPUs, and will run for at maximum 2 days. In order to
+allocate more resources and time to your job, you must set one or more of these
+flags:**
+
+* `-n` / `--ntasks`:
+    - The number of tasks/processes to allocate. **Default is 1.**
```
kpengboy: Is it desirable to make all of these defaults bold?

ocfweb/docs/docs/services/hpc/slurm.md
```diff
@@ -0,0 +1,142 @@
...
+
+However, to pass options into SLURM, you'll need to add some special comment
+lines, which are in the format `#SBATCH [command-line flag]=[value]`. **They
+must be after the shebang but before any non-comments**.
+
+For example, a batch script which uses 4 CPUs, 8 GB of RAM, and 1 GPU has its
+contents as:
+
+```
+ #!/bin/bash
```
kpengboy: Why the leading spaces?

redplanks: It'll trigger the pre-commit hook about "No top-level headers" otherwise.

ocfweb/docs/docs/services/hpc.md
keur: No do not put cores in scare quotes.

ocfweb/docs/docs/services/hpc.md
```diff
@@ -52,22 +52,12 @@ generously funded by a series of grants from the [Student Tech Fund][stf].
 
 ## Slurm
 
-We currently use [Slurm][slurm] as our workload manager for the cluster. We
-will soon post technical documentation about our Slurm configuration, but briefly,
+We currently use [Slurm][slurm] as our workload manager for the cluster. Briefly,
```
keur: just remove briefly from this sentence

ocfweb/docs/docs/services/hpc.md
abizer: probably better to make that intention explicit in the same section where you explain how many GPUs can be allocated, rather than in the spec sheet

ocfweb/docs/docs/services/hpc.md
```diff
@@ -52,22 +52,12 @@ generously funded by a series of grants from the [Student Tech Fund][stf].
...
 Slurm is a free and open-source job scheduler which helps distribute jobs from
```
keur: Does it help to distribute jobs or do the distributing itself? A bit confusing

How about: "Slurm is a free and open source job scheduler that evenly distributes jobs across an HPC cluster, where each computer in the cluster is referred to as a node."

ocfweb/docs/docs/services/hpc/slurm.md
```diff
@@ -0,0 +1,142 @@
...
+that includes options for your job, and the commands that your job will run.
+The batch file is then submitted, and after the job runs, log files with the
+job's output to `stdout` are put into your home directory.
+
+#### When to use `srun` versus `sbatch`
+
+If your job requires interactivity or inputs from the terminal, or you need a
+terminal to test or experiment, use `srun`. Otherwise, use `sbatch`, as you
+don't have to keep your terminal open until the job runs.
+
```
keur: Could mention that even noninteractive programs can use `srun` with `tmux` or `screen`.

ocfweb/docs/docs/services/hpc/slurm.md
```diff
@@ -0,0 +1,142 @@
...
+* `--t` / `--time` **(Optional)**:
+    - The maximum amount of time your job can take before Slurm forcefully
+kills it. Acceptable time formats include "minutes", "minutes:seconds",
+"hours:minutes:seconds", "days-hours", "days-hours:minutes" and
+"days-hours:minutes:seconds". You cannot set the time limit greater than the
+**default, which is 2 days.**
+
+### Using `srun`
+
+On `hpcctl`:
```
keur: This is the first mention of `hpcctl`. You ought to tell users how to get there first.

ocfweb/docs/docs/services/hpc/slurm.md
```diff
@@ -0,0 +1,145 @@
...
+On `hpcctl.ocf.berkeley.edu`:
+
+```
+srun [command-line flags] [command to run]
+```
+
+For example, to run a job that uses 4 CPUs, 8 GB of RAM, and 1 GPU:
+
+```
+bzh@hpcctl:~$ srun --ntasks=1 --cpus-per-task=4 --mem=8G --gres=gpu:1 echo "Hello world!"
```
kpengboy: When users are actually logged in they'll see `username@segfault`. Wonder if we should comment on this...

ocfweb/docs/docs/services/hpc.md
```diff
@@ -52,22 +52,12 @@ generously funded by a series of grants from the [Student Tech Fund][stf].
...
-* `squeue`: Used to view running and queued jobs.
-* `scancel`: Used to cancel jobs.
-* `sinfo`: Used to view status of compute nodes in a cluster.
-
-Berkeley Research Computing (BRC) has some useful documentation for using Slurm
-[here][brc_slurm], just keep in mind that there are some differences between
-their configuration and ours. To put it simply, all of your programs will be
-run through Slurm.
+We currently use [Slurm][slurm] as our workload manager for the cluster.
+Slurm is a free and open source job scheduler that evenly distributes jobs across an HPC cluster,
```
kpengboy: Nit: wrap

ocfweb/docs/docs/services/hpc/slurm.md
kpengboy: There should be another way to silence that, no?

redplanks: It's a [simple regex](https://github.com/ocf/ocfweb/blob/master/.pre-commit-config.yaml#L58), so I don't think so? Unless pre-commit has a mechanism to silence a check for a single file, but a quick search didn't reveal anything.

ocfweb/docs/docs/services/hpc/slurm.md
redplanks: I feel that HPC users are probably technically competent enough to not get confused by it, if they do notice.

PR #451, opened 2019-01-21
Howto redesign part 1

dkess:
This is the first phase of re-organizing staffdocs to be nicer to use. There are two big changes in this PR:

* I've re-organized the folder structure for staffdocs, so there is now a "howto" section with subfolders that represent categories.
* The howto section is displayed using a masonry layout

It looks like this:
2019-01-20-224841_1316x818_scrot

Note that this PR is not being merged into the master branch-- instead I'm using a branch called staffdocs-redesign. I don't want to productionize this until the URLs stabilize. In a future PR, I plan to merge some pages (like sorry/unsorry) together.


ocfweb/docs/urls.py
```diff
@@ -37,6 +38,47 @@
...
+    '/staff/scripts/how': 'staff/howto/other/how',
+    '/staff/scripts/ocf-tv': 'staff/howto/other/ocf-tv',
+    '/staff/scripts/pdf-open': 'staff/howto/other/pdf-open',
+    '/staff/procedures/gapps': 'staff/howto/staff-admin/gapps',
+    '/staff/procedures/granting-privileges': 'staff/howto/staff-admin/granting-privileges',
+    '/staff/procedures/accounts/alumni-reset': 'staff/howto/user-services/alumni-reset',
+    '/staff/scripts/approve': 'staff/howto/user-services/approve',
+    '/staff/scripts/paper': 'staff/howto/user-services/paper',
+    '/staff/procedures/vhost': 'staff/howto/user-services/vhost',
+
```
kpengboy: Extra line?

dkess:
Thanks for the feedback. I realize this patch has been sitting here for quite some time. However, this is kind of a good thing since I'm looking at the categories with a fresh set of eyes, and I agree that changes should be made.

Putting printing in its own category seems like a good decision, since it certainly seems tangential to most of our other infrastructure.

Infrastructure and Maintenance also look like they should be merged. I don't think I even remember what the difference was supposed to be for those.

User Services and Account Management have a slightly better line between them: user services are procedures specific to a request a user might make, either over email or in person. Account management is for general tools used when dealing with accounts.

I do plan to merge some articles here. If the number of articles in user services and account management are small enough, I would definitely consider merging those categories as well.

dkess:
Applied changes, here's what it looks like now:
2019-04-05-002105_1279x785_scrot

fydai:
Misclick sorry, I closed this and have no idea how to reopen it.

dkess:
fwiw, I am not sure if this will ever be merged. I would rather us use a static site generator for docs.

PR #454, opened 2019-01-25
Add Kubernetes support

keur:


kubernetes/ocfweb-web.yaml
```diff
@@ -0,0 +1,72 @@
...
+    matchLabels:
+      app: ocfweb-web
+  template:
+    metadata:
+      labels:
+        app: ocfweb-web
+    spec:
+      containers:
+        - name: ocfweb-web-deploy
+          image: "docker.ocf.berkeley.edu/ocfweb-web:keur-1"
```
kpengboy: This should be changed to the real value before merging, no? (Same for the other files)

keur: Yeah thanks

kubernetes/ocfweb-worker.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: ocfweb-worker-service
+spec:
+  selector:
+    app: ocfweb-worker
+  ports:
+    - port: 80
+      targetPort: 8000
```
dkess: I don't think this needs a `Service`.

keur: TrUwU

keur:
@jvperrin @chriskuehl if there is a better way to make `pre-commit` ignore multiple yaml streams in the same file, while also running everything else in the `check-yaml` hook, that would be ideal.

kubernetes/ocfweb-web.yaml.erb
```diff
@@ -0,0 +1,72 @@
...
+            path: /opt/share/kubernetes/secrets/ocfweb
+            type: Directory
+        - name: ocfweb-ocf
+          hostPath:
+            path: /etc/ocf
+            type: Directory
+      dnsPolicy: ClusterFirst
+      dnsConfig:
+        searches:
+          - "ocf.berkeley.edu"
```
cg505: @keur why is this necessary?

cg505:
also why are the tests failing?

kubernetes/ocfweb-web.yaml.erb
```diff
@@ -0,0 +1,72 @@
...
+      targetPort: 8000
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ocfweb-web-deployment
+  labels:
+    app: ocfweb-web
+spec:
+  replicas: 1
```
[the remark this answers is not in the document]
keur: good catch. this was a testing setting back when the cluster was on lightly provisioned VMs

kubernetes/ocfweb-web.yaml.erb
keur: It's when we use OCF hostnames (not fqdn). But it doesn't appear to be necessary. After glancing over the code I can't seem to find an instance where we use one.

keur:
@cg505 just needed a rebase

kubernetes/ocfweb-web.yaml.erb
dkess: I'd keep it just in case, it's probably used in various places in ocflib.

cg505: Do we need `dnsPolicy: ClusterFirst` though?

kubernetes/ocfweb-worker.yaml.erb
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ocfweb-worker-deployment
+  labels:
+    app: ocfweb-worker
+spec:
+  replicas: 1
```
cg505: :thinking:

keur: (🤔)[https://github.com/ocf/services/blob/090b1ef6c09d82b0f1713d47ffe7692a2fb50356/apps/ocfweb/worker/app.yaml#L8]

keur: also, i think the reasoning is since it just writes stuff to redis, even if the worker goes down the site is still live. also having multiple workers doing the same computations and persisting them to redis (one potentially overwriting the other) is not particularly helpful.

kubernetes/ocfweb-web.yaml.erb
keur: `ClusterFirst` is the default `dnsPolicy` so no we don't need it because it doesn't change anything. For clarity, when using `dnsConfig`, I always include it. I'm not particularly attached if you want to remove it.

kubernetes/ocfweb-static.yaml.erb
```diff
@@ -0,0 +1,52 @@
...
+          ports:
+            - containerPort: 8000
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: vhost-static-ingress
+spec:
+  rules:
+    - host: static.www.ocf.berkeley.edu
```
abizer: static.www.ocf?

cg505: wait...

dkess:
Is this ready for another review?

fydai:
Yes

kubernetes/ocfweb-web.yaml.erb
fydai: in particular, logs are sent to "syslog"

PR #459, opened 2019-02-05
Update django web hosting docs

jvperrin:
Add back the django webhosting docs (previously a mostly blank page that just said to use apphosting) that were removed in 3c51a6d3e460fd7423e93f7f022778ed777eb24a

### DIFF
--- a/ocfweb/docs/docs/services/web/django.md
+++ b/ocfweb/docs/docs/services/web/django.md
@@ -1,6 +1,133 @@
 [[!meta title="Django"]]
 
-**Note: If you are using a group account, you may wish to consider
-[[apphosting|doc services/webapps]] instead. The OCF is currently
-working on updating this documentation to better help users help
-set this up.**
+**Note: These instructions are primarily intended for individual user accounts.
+If you are using a group account, you may wish to consider
+[[apphosting|doc services/webapps]] instead.**
+
+Django is a popular web framework for Python applications. At the OCF, you can
+host it using our regular web hosting, though for groups with virtual hosting,
+we strongly recommend instead using our [[application hosting service|doc
+services/webapps]] instead, as it provides a first-class solution for webapps.
+
+## Running a Django app
+
+The steps below assume you have already created a Django application. If you
+haven't, take a look at [the Django documentation][django-docs] for help with
+that.
+
+1. Copy your Django application to your OCF account. For example, you can use
+   `scp` to upload it, or if you're using source control, clone it using `git`
+   (or another tool) into your home directory.
+
+   In the steps below, we'll assume you put it at `~/app` in your home
+   directory.
+
+2. Inside your project directory, create a virtualenv and install your
+   dependencies. For example, to create a Python 3 virtualenv, try:
+
+   ```bash
+   user@tsunami:~/app$ virtualenv -p python3 venv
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/services/web/django.md
```diff
@@ -1,6 +1,133 @@
...
+   (or another tool) into your home directory.
+
+   In the steps below, we'll assume you put it at `~/app` in your home
+   directory.
+
+2. Inside your project directory, create a virtualenv and install your
+   dependencies. For example, to create a Python 3 virtualenv, try:
+
+   ```bash
+   user@tsunami:~/app$ virtualenv -p python3 venv
```
ethanhs: maybe recommend `python3 -m venv myenv`?

ocfweb/docs/docs/services/web/django.md
```diff
@@ -1,6 +1,133 @@
...
+   other packages, install those too. You can also install from a
+   `requirements.txt` file, if you have one.
+
+   For full details on how to use pip, see [the pip documentation][pip-docs].
+
+3. Inside that same virtualenv, install the `flup6` package. You can just
+   copy-and-paste the below snippet:
+
+   ```bash
+   user@tsunami:~/app$ venv/bin/pip install flup6
```
ethanhs: Instead of running things prefixed with the venv a lot, perhaps we should tell them to source the venv?

ocfweb/docs/docs/services/web/django.md
jameslzhu: +1, does not require `pip install virtualenv` in Python 3

jvperrin: I would do this, but this requires `python3-venv` be installed, which it isn't currently on tsunami (although we could install it there perhaps). `virtualenv` on the OCF systems comes from the `virtualenv` package, which we do have installed, apparently as a dependency of `tox`, which is installed here: https://github.com/ocf/puppet/blob/1df64fb6b89d33d21c3eed1478e61659c5144534/modules/ocf/manifests/extrapackages.pp#L212

ocfweb/docs/docs/services/web/django.md
jvperrin: Maybe, I see upsides of this (it makes running commands simpler), and downsides (they might wonder why nothing is working when they don't have it sourced). Maybe mentioning sourcing as an option but still including the full path is a good compromise? Then I suppose you get the worst of the unsourced world in that you still have to type the full command and none of the benefit of sourcing... :/

ocfweb/docs/docs/services/web/django.md
ethanhs: Yeah I remember trying to use that a few days ago. I think we can just add it to extrapackages? (I am happy to do so). We already have `python3.7-venv` because I puppet'd that when I added 3.7 (which reminds me I need to finish setting up the jenkins build for that).

ocfweb/docs/docs/services/web/django.md
ethanhs: that is fair. I think mentioning sourcing is an option and then that you won't have to prefix commands with venv/bin/ would be a good compromise.

PR #460, opened 2019-02-05
Update flask web hosting docs

jvperrin:
Update flask web hosting docs (tested, but feel free to test it again with any sites or setups you have)


ocfweb/docs/docs/services/web/flask.md
```diff
@@ -4,81 +4,124 @@
...
-1.    Go into the directory you just created, and create two files: *.htaccess*
-      (note the dot), and *run.fcgi*.
+2. Make a virtualenv.  This will allow you to run `pip install` to install any
+   python packages that you need that the OCF does not already have. The
+   benefit of having a virtualenv is that you can decide the dependencies you
+   want, without being tied to the OCF defaults:
 
-* .htaccess:
+   ```shell
+   user@tsunami:~/app$ virtualenv -p python3 venv
```
ethanhs: Thank you for suggesting people use venvs :)

jameslzhu: Similar to #459, perhaps `python3 -m venv venv` instead?

jvperrin: As I mentioned there, on debian this apparently needs `python3-venv` to be installed, which tsunami doesn't currently have, although it sounds like it should be installed there.

Issue #453, opened 2019-01-25
Account commands returns output as stringified binary


dkess:
Not a huge deal, but we should be parsing this as UTF-8. This happens on https://www.ocf.berkeley.edu/account/commands/.

Screenshot:
![2019-01-24-184511_778x701_scrot](https://user-images.githubusercontent.com/749522/51721835-3014af80-2008-11e9-9451-b779db9519a3.png)

TonyLianLong:
We shall add a `decode` function to it.
Tested in repl:
```python3
>>> print(b'longlian printing quota:\n \xe2\x86\x92 90 remaining this semester\n \xe2\x86\x92 20 remaining today\n'.decode())
longlian printing quota:
 → 90 remaining this semester
 → 20 remaining today

>>> 
```

TonyLianLong:
Also why should it be parsed as UTF-8 since there is no `\u` in the string?

PR #469, opened 2019-02-11
Add decode to output and error

TonyLianLong:
This is to fix issue #453.


TonyLianLong:
Tested on dev ocfweb:
image

PR #473, opened 2019-02-17
add functionality for bouncing shorturls

abizer:
requires https://github.com/ocf/ocflib/pull/154

### DIFF
--- a/ocfweb/api/shorturls.py
+++ b/ocfweb/api/shorturls.py
@@ -1,0 +1,15 @@
+from django.http import HttpResponseNotFound
+from django.http import HttpResponseRedirect
+from ocflib.misc.shorturls import get_connection
+from ocflib.misc.shorturls import get_shorturl
+
+
+def bounce_shorturl(request, slug):
+    if slug:
+        with get_connection() as ctx:
+            target = get_shorturl(ctx, slug)
+
+        if target:
+            return HttpResponseRedirect(target)
+
+    return HttpResponseNotFound()
--- a/ocfweb/api/urls.py
+++ b/ocfweb/api/urls.py
@@ -3,10 +3,13 @@ from django.urls import path
 from ocfweb.api import hours
 from ocfweb.api import lab
 from ocfweb.api import session_tracking
+from ocfweb.api import shorturls
+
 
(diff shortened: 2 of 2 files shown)


dkess:
Do we ever expect this to be used?

abizer:
yes, this is how the shorturls will get redirected. Apache on death will turn `ocf.io/s/whatever` into `www.ocf.berkeley.edu/api/shorturl/whatever` which will return the `HttpResponseRedirect` to the final destination.

ocfweb/api/urls.py
```diff
@@ -3,10 +3,13 @@
...
 from ocfweb.api import session_tracking
+from ocfweb.api import shorturls
+
 
 urlpatterns = [
     path('hours', hours.get_hours_all, name='hours_all'),
     path('hours/today', hours.get_hours_today, name='hours_today'),
     path('lab/desktops', lab.desktop_usage, name='desktop_usage'),
     path('session/log', session_tracking.log_session, name='log_session'),
+    path('shorturl/<slug:slug>', shorturls.bounce_shorturl, name='bounce_shorturl'),
```
cg505: We don't actually have a slug path converter, do we? Shouldn't this just be `'shorturl/<slug>'`?

Reference: https://docs.djangoproject.com/en/2.1/topics/http/urls/#registering-custom-path-converters

abizer: it's built-in, I think. This seems to work fine when I run it in testing.

ocfweb/api/shorturls.py
```diff
@@ -0,0 +1,15 @@
+from django.http import HttpResponseNotFound
+from django.http import HttpResponseRedirect
+from ocflib.misc.shorturls import get_connection
+from ocflib.misc.shorturls import get_shorturl
+
+
+def bounce_shorturl(request, slug):
+    if slug:
+        with get_connection() as ctx:
+            target = get_shorturl(ctx, slug)
```
dkess: Does this cause an exception/internal server error if the slug doesn't exist?

abizer: no, the ocflib function returns `None`

Issue #437, opened 2018-12-29
add ssh host key fingerprints to server listing


abizer:
we list tsunami's host key fingerprint at ocf.io/ssh but it would be useful to have the other machine's fingerprints listed somewhere publically as well so we can verify them externally if we get a key mismatch or something like that

jvperrin:
This should be pretty easy to do if we use puppetdb for the server list, since there's a bunch of SSH key fingerprint facts:

```
sshdsakey
sshecdsakey
sshed25519key
sshfp_dsa
sshfp_ecdsa
sshfp_ed25519
sshfp_rsa
sshrsakey
```

(we probably don't want the `sshfp` ones though since those are for usage in DNS and we don't currently use them)

dkess:
With #406, ocfweb will have its own Puppet cert which will make this even easier. Perhaps we should even list them on the servers page?

cg505:
I'm currently just hosting the actual known_hosts file we use at https://failure.ocf.berkeley.edu/ssh_known_hosts so that I can pull it into the known hosts file on my machine. We should consider making such a file available on www.o.b.e, possibly only including non-staff hosts (segfault, werewolves, tsunami, etc).

PR #475, opened 2019-03-05
Added sshrsa and sshdsa keys to server listing

indianwolverine:
addresses issue #437


ocfweb/docs/views/servers.py
```diff
@@ -13,9 +13,9 @@
 PUPPET_CERT_DIR = '/etc/ocfweb/puppet-certs'
 
 
-class Host(namedtuple('Host', ['hostname', 'type', 'description', 'children'])):
+class Host(namedtuple('Host', ['hostname', 'type', 'description', 'children', 'ssh_keys'])):
```
dkess: Instead of having `ssh_keys` be a dictionary, we should have two properties, `rsa_key` and `dsa_key` which are just strings. This means we don't have to have four functions to get these keys.

ocfweb/docs/templates/docs/partials/host.html
```diff
@@ -24,6 +24,24 @@
...
+              <p>This machine does not have an SSH RSA Key</p>
+          {% else %}
+              <a href="#sshrsa{{host.hostname}}">Show SSH RSA Key: </a><p class="expand" id="sshrsa{{host.hostname}}">{{host.ssh_rsa_key}}</p>
+          {% endif %}
+        {% endif %}
+    </div>
+    <div class="col-sm-12 ssh-keys">
+            {% if host.has_ssh_dsa_key %}
+            {% if host.ssh_dsa_key == None %}
+                <p>This machine does not have an SSH DSA Key</p>
```
cg505: imo, If it doesn't have a key of one type, it should just show nothing.

indianwolverine: interesting, easy fix, but why do you think so?

indianwolverine: from #437 jvperrin mentioned that we don't want the `sshfp` facts since we don't use them

ocfweb/docs/views/servers.py
indianwolverine: servers_with_keys

abizer:
bump @indianwolverine

PR #483, opened 2019-03-15
staffdocs: add GitHub issues links to getinvolved page

dkess:
These should have been here the whole time but I recently learned that we could do these searches. Way easier then telling people to check all the repos manually.


dkess:
If no one has comments I'll merge this before next staff meeting on Monday.

PR #498, opened 2019-04-22
Update lab photo

MrMinos:
photo credit @edricx; ocfusername: exiang


dkess:
Wow this new lab looks great!

MrMinos:
> What does the page itself look like with the image? The image looks good though, I'd say go for it

https://i.fluffy.cc/mmMRxfztVwH0wFPCLp6q182mzN2df20h.png

kpengboy:
Why's it a PNG and not a JPG?

abizer:
also please make the image smaller

MrMinos:
> Why's it a PNG and not a JPG? I get that the current photo's not a JPG, but I think JPG is still a more suitable format for photos.

Actually, it's a jpg file having png extension. I assigned a wrong extension.

> Also might want to update the filename and the alt text where the image is included.

yes, that'd be good.

MrMinos:
> the outdated lab pic is part of the core essence of the open computing facility. please don't derive us of our cultural heritage.

thanos quote

MrMinos:
> Nice, thanks for showing what it looks like in the page. I think the image might be a bit tall, at least compared to the existing image, but I'm not sure if cropping the top and bottom would be better or not.

Yeah, I am not sure about that either. Maybe we can put more words to the right.

Actually, I'm thinking about adding more to this PR; like add actual [server image](https://i.fluffy.cc/lnrxmv2fQ51TVmWpBlkXjLx5hgQpQpMN.JPG) as well.

ocfweb/about/templates/about/staff.html
```diff
@@ -83,7 +83,7 @@ <h3>Computer Lab &amp; Printing</h3>
         <div class="container">
             <div class="row">
                 <div class="col-md-7">
-                    <img src="{% static 'img/about/server-lab.png' %}" alt="servers and lab" />
+                    <img src="{% static 'img/about/171-mlk-lab.jpg' %}" alt="ocf at MLK building, 2019 Spring" />
```
kpengboy: Nit: OCF

MrMinos: changed

MrMinos:
at the moment: http://supernova.ocf.berkeley.edu:8163/about/staff is running the dev env.

ocfweb dev image

kpengboy:
Nice, looking good. I wonder if there's a way to vertically center the lab photo with respect to the text and the server room photo. It might make the gap a little less obvious. Or maybe someone has an even better idea?

cg505:
Maybe we can get a cleaner server room photo (like without all the junk on top of corruption)

MrMinos:
Moved the lab photo horizontally and it looks like this.
a screenshot of dev ocfweb

MrMinos:
> Maybe we can get a cleaner server room photo (like without all the junk on top of corruption)

@edricx if you have time, could you take another shot of the server room? It was cleaned by staffers for IST meeting today, 04/24/2019.

ocfweb/about/templates/about/staff.html
```diff
@@ -82,13 +82,20 @@ <h3>Computer Lab &amp; Printing</h3>
...
+                <br/>
+                <br/>
+            </div>
+                    <img src="{% static 'img/about/171-mlk-lab.jpg' %}" alt="OCF at MLK building, 2019 Spring" />
                 </div>
                 <div class="col-md-5">
                     <h2>We have a spiffy computer lab on southside.</h2>
                     <h3><a href="{% url 'register' %}">Create an account</a> and drop by!</h3>
                     <p>Our computer lab, located in <a href="{% url 'doc' 'services/lab' %}">171 MLK Student Union</a>, features high-res scanners, free printing, and high-end desktops with mechanical keyboards and GTX graphics cards, all running Debian GNU/Linux.</p>
+                    <div class="col-md-12">
```
cg505: I don't think Bootstrap is intended to have nested columns. See if the `class="col-md-12"` is necessary.

MrMinos: https://getbootstrap.com/docs/4.0/layout/grid/#nesting

nesting is a feature, and I think the alignment broke without it.

cg505: You'll need to add a `<div class="row">` per those docs. Also, note that we are using Bootstrap 3, not 4.

MrMinos: removed from the code

ocfweb/about/templates/about/staff.html
```diff
@@ -82,13 +82,20 @@ <h3>Computer Lab &amp; Printing</h3>
     <div class="ocf-lab">
         <div class="container">
             <div class="row">
-                <div class="col-md-7">
-                    <img src="{% static 'img/about/server-lab.png' %}" alt="servers and lab" />
+                <div class="col-md-7 align-items-center">
+            <div class="col-md-7">
+                <br/>
+                <br/>
+            </div>
```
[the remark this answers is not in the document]
MrMinos: Yes, using _inline CSS_ instead, should we move this to main css file?

MrMinos:
mobile screenshot
mobile dev ocfweb screenshot

desktop screenshot
desktop dev ocfweb screenshot

ocfweb/about/templates/about/staff.html
```diff
@@ -82,13 +82,14 @@ <h3>Computer Lab &amp; Printing</h3>
     <div class="ocf-lab">
         <div class="container">
             <div class="row">
-                <div class="col-md-7">
-                    <img src="{% static 'img/about/server-lab.png' %}" alt="servers and lab" />
+                <div class="col-md-7" style="padding-top:3em;">
```
cg505: You could also consider `margin: auto 0;`, which might get closer to what you want (vertical centering). This is definitely an improvement though. If you want to merge as-is, that's fine with me.

cg505: Also, if you place it in a separate CSS file, you might be able to use media queries to remove the margin/padding on mobile.

MrMinos: > `margin: auto 0;`
unfortunately that does not help in this case as the `div` element of the column automatically is the size of the picture (the only element in that `div`), hence that `div` does not expand to the full height of the column next to it.

cg505: ah, makes sense. I'd go ahead and merge

MrMinos: moved inline css to the proper css file; the webpage looks the same as previous commit.

PR #500, opened 2019-04-23
Fix redis connection

ethanhs:
The newer redis requires using a new URL schema.


keur:
fwiw quick fixes like this can go straight into master

ethanhs:
@keur, yeah I just wanted to double check this passed CI, sadly it doesn't look like it actually fixed the issue so I've rolled back for now :/

PR #502, opened 2019-04-23
Update requirements

ethanhs:
This does two things:
- updates requirements
- work around redis not liking `redis://` URIs

I will merge this once CI is green


ethanhs:
Well so much for that idea. I think I'll just have settings.py replace redis:// with rediss://

PR #504, opened 2019-04-23
Update to Python 3.7 \o/

ethanhs:
Bless @asottile for pyupgrade.


.pre-commit-config.yaml
```diff
@@ -44,12 +43,12 @@ repos:
     sha: v1.11.0
```
asottile: might as well take this to 1.16.1 too -- there's some new py3 goodies in the later versions

PR #512, opened 2019-05-09
Update language in README

dkess:
I tried to make the README clearer to complete beginners to OCF infra, but more suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,20 +10,25 @@ ocfweb
 
 ## Working on `ocfweb`
 
+We recommend following all of these steps
+[on supernova](https://www.ocf.berkeley.edu/docs/staff/procedures/ssh-supernova/),
+the staff login server, because it is already configured to run ocfweb in
+development mode with minimal extra setup.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,20 +10,24 @@ ocfweb
 
 ## Working on `ocfweb`
 
+We recommend following all of these steps
+[on supernova](https://www.ocf.berkeley.edu/docs/staff/procedures/ssh-supernova/),
+the staff login server. This will let you preview your changes.
```
jvperrin: While this is true, the main reason we recommend running this on supernova is because it has the right packages, files, and network ports open through the firewall available to actually run ocfweb without extra setup, where other machines do not. I think the ability to preview changes is mostly incidental and this should be something like "We recommend following all of these steps on supernova, the staff login server, because it is already set up to run ocfweb with minimal extra setup".

dkess:
Thanks for the feedback! I've updated it, gonna merge now, however if anyone else has suggestions feel free to mention them.

PR #515, opened 2019-05-26
Change get_lab_status() to read from ocf's etc repo.

redplanks:
In the future we should probably move the banner's HTML and CSS to within ocfweb and have the yaml file specify the text.

### DIFF
--- a/ocfweb/component/lab_status.py
+++ b/ocfweb/component/lab_status.py
@@ -1,6 +1,5 @@
 from collections import namedtuple
 
-import requests
 import yaml
 
 from ocfweb.caching import periodic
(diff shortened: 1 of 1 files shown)


ocfweb/component/lab_status.py
```diff
@@ -17,15 +16,8 @@
...
-        with open('/home/s/st/staff/lab_status.yaml') as f:
-            tree = yaml.safe_load(f)
-    except IOError:
-        tree = yaml.safe_load(
-            requests.get(
-                'https://www.ocf.berkeley.edu/~staff/lab_status.yaml',
-            ).text,
-        )
+    with open('/etc/ocf/lab_status.yaml') as f:
+        tree = yaml.safe_load(f)
```
abizer: this should also probably fail with the IOError, in case there's a problem mounting the directory into the container

redplanks: I was thinking that we'd delete the lab_status.yaml file from ~staff once this is merged

abizer: keeping the old thing would actually be nice as a backup, but deleting it is fine too.

redplanks: Yeah, that is true. I was thinking it might lead to confusion in the future though, on whether you have to change the file in etc, the one in ~staff, or both.

dkess: We shouldn't have a "fallback" option unless there's a way of notifying us that the fallback path was taken, otherwise it "fails" silently. In that case, it should be part of ocflib (and used for all `etc` configs), not just this one.

PR #520, opened 2019-06-29
Add Staff Hours to API

BernardZhao:
Exposes our staff hours to the API. Uses the existing cached function.

### DIFF
--- a/ocfweb/api/staff_hours.py
+++ b/ocfweb/api/staff_hours.py
@@ -1,0 +1,13 @@
+from json import JSONEncoder
+
+from django.http import JsonResponse
+
+from ocfweb.main.staff_hours import get_staff_hours as real_get_staff_hours
+
+
+def get_staff_hours(request):
+    return JsonResponse(
+        list(map(lambda item: item._asdict(), real_get_staff_hours())),
+        encoder=JSONEncoder,
+        safe=False,
+    )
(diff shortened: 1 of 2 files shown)


ocfweb/api/staff_hours.py
```diff
@@ -0,0 +1,13 @@
...
+from django.http import JsonResponse
+
+from ocfweb.main.staff_hours import get_staff_hours as real_get_staff_hours
+
+
+def get_staff_hours(request):
+    return JsonResponse(
+        list(map(lambda item: item._asdict(), real_get_staff_hours())),
+        encoder=JSONEncoder,
+        safe=False,
```
BernardZhao: These two lines probably are unnecessary.

ocfweb/api/staff_hours.py
```diff
@@ -0,0 +1,13 @@
+from json import JSONEncoder
+
+from django.http import JsonResponse
+
+from ocfweb.main.staff_hours import get_staff_hours as real_get_staff_hours
+
+
+def get_staff_hours(request):
+    return JsonResponse(
+        list(map(lambda item: item._asdict(), real_get_staff_hours())),
```
ethanhs:
```suggestion
        [item._asdict() for item in real_get_staff_hours()],
```

ocfweb/api/staff_hours.py
ethanhs: Yes, I think it would be better to use the default Django encoder. Is there a reason you want to use the `json` encoder?

BernardZhao: Nope, this was my mistake

PR #523, opened 2019-07-07
Add Blog to API

BernardZhao:
Pretty much the same deal as #520. This could probably be improved by adding a limit parameter on the get request to get the few most recent blog posts, might do that later.


dkess:
This code is fine, but I don't think this is a good long-term way to handle displaying the blog. We have a project to [move the blog to a separate static site](https://ocf.io/d/229) since it can reasonably be distinct from the rest of what ocfweb does.

PR #527, opened 2019-08-18
don't add tornado and overheat twice to server listing

cg505:
ocf.io/servers currently has overheat and tornado listed twice.


ocfweb/docs/views/servers.py
```diff
@@ -116,7 +116,11 @@ def get_hosts():
     hypervisors_hostnames = dict(format_query_output(item) for item in query_puppet(PQL_IS_HYPERVISOR))
     all_children = dict(format_query_output(item) for item in query_puppet(PQL_GET_VMS))
 
-    hostnames_seen = set()
+    hostnames_seen = {
+        # These are manually added latter, with the correct type
```
kpengboy: Nit: typo

cg505: f

PR #537, opened 2019-08-28
make updates to eligibility

pxhanus:
Updated instructions for letter to email in eligibility requirements where things like departmental letterhead were not needed.


abizer:
in keeping with previous practice, this should still come up at BoD as it relates to a policy change.

nikhiljha (approved):

PR #542, opened 2019-09-02
Moving docs to submodule

BernardZhao:
I want to move docs to its own repo so that they can be operated on and used [here](https://github.com/BernardZhao/ocfweb) separately.

The next steps I want to make include making sure that changing doc structure does not affect ocfweb, so that these two repos can be completely separate. Then I can begin to make improvements like changing our `[[!meta title="Old title syntax"]]` to YAML definitions that all other markdown parsers use.


cg505:
I'm a bit worried that this will make the barrier to entry higher, because to update docs you will have to make 2 PRs: one to docs and one to ocfweb to update the submodule. I wonder if we could figure out a way to split docs out to another repo without using a submodule (or make the update process easier). For example, Jenkins on the docs repo maybe could push a commit to this repo updating the submodule.

BernardZhao:
> I'm a bit worried that this will make the barrier to entry higher, because to update docs you will have to make 2 PRs: one to docs and one to ocfweb to update the submodule. I wonder if we could figure out a way to split docs out to another repo without using a submodule (or make the update process easier). For example, Jenkins on the docs repo maybe could push a commit to this repo updating the submodule.

The goal would actually be to make the barrier to entry lower for editing docs, because they would no longer need to worry about working with the website to make changes to the docs. I would also probably need to make changes so that people working on ocfweb don't have to worry about doc changes either. Submodules can follow branches now with `git submodule add -b <branch> <repository> [<path>]`, so another PR to ocfweb wouldn't be needed to update the docs.

cg505:
> Submodules can follow branches now with git submodule add -b <branch> <repository> [<path>], so another PR to ocfweb wouldn't be needed to update the docs.

I believe that in this case you still need to do `git submodule update --remote` to update the submodule. I guess maybe we can just do this as part of the deploy process, but it seems a bit funky.

kpengboy:
Some history: we used to actually have docs in a separate repo (which you can still find under the moniker of "wiki-old"). When ocfweb was created, ckuehl migrated the commit history of wiki-old into the ocfweb repo. Do you think it would be beneficial to do something similar for the new docs repository?

BernardZhao:
After giving this some more thought, this probably isn't necessary for what I am trying to do. Or rather, it really isn't worth the effort. I guess if I wanted to push the new frontend through, I might as well maintain another version of the docs anyways, which would be easier for me to change and modify. I really just wanted the docs to be the exact same on the current ocfweb and what I'm trying to build.

PR #543, opened 2019-09-10
Update slack documentation

fydai:

PR #544, opened 2019-09-10
Fix typo and respond to comments from #543

ethanhs:


ethanhs:
Thanks for the quick review @dkess :P

PR #551, opened 2019-10-15
Fix 500 Error on stats/printing

64bitpandas:
Due to the power outage, no data was collected on 2019/10/10, so ocf.io/stats/printing would throw a KeyError when trying to display the missing data.

I added this to fill in the blanks for the last month's worth of data in case another power outage / server issue happens again.
```python
    # Fix missing rows in table (if data collection failed)
    for day in (date.today() - timedelta(n) for n in range(30)):
        if day not in pages_printed.keys():
            pages_printed[day] = defaultdict(int)
```

Now all the tests pass again :D

### DIFF
--- a/ocfweb/stats/printing.py
+++ b/ocfweb/stats/printing.py
@@ -162,11 +162,6 @@ def _pages_per_day():
                 )
             last_seen[row['printer']] = row['value']
 
-    # # Fix missing rows in table (if data collection failed)
-    # for day in (date.today() - timedelta(n) for n in range(30)):
-    #     if day not in pages_printed.keys():
-    #     pages_printed[day] = defaultdict(int)
-
     return pages_printed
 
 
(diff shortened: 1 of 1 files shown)


64bitpandas:
Great, thanks for the suggestion! Commit incoming

ocfweb/stats/printing.py
```diff
@@ -152,8 +153,7 @@ def _pages_per_day():
         ''')
 
         last_seen = {}
-        pages_printed = {}
-
+        pages_printed = pages_printed = defaultdict(partial(defaultdict, int))
```
ethanhs: It looks like you accidentally assigned here twice, unless I am mistaken.

Also if you don't mind, a comment explaining why we have a `defaultdict(partial(defaultdict, int))` would be appreciated. Thanks!

jvperrin: Ah, good catch on the double assignment, I was clearly looking too fast and didn't see that

Agreed on a comment

64bitpandas: Oops! thanks for the reviews, double assignment should be fixed. I need sleep :p

ethanhs: Happens to everyone :P

PR #557, opened 2019-11-04
Implement mypy annotations (Resolves #549)

### LINKED ISSUE #549: Add mypy annotations for Python gradual typing
(I'm currently working on this but help is appreciated)

It would be nice to have some static type checking using mypy (like what has already been implemented in slackbridge by @ethanhs ). This will involve enabling more flags in `mypy.ini` and making sure all of the Python code in the repo is annotated properly.

64bitpandas:
This should make improper type usages much easier to catch and encourage better self-documentation in the future :)

Some notes:
 - ~~Typing annotations are very much WIP and I used `Any` too liberally, especially for `request: Any` since I was unsure what type `request` is.~~ `request: Any` has been fixed and replaced with `request: HttpRequest`
 Some of the `Any` keywords are used purposefully, but many can definitely be improved.
 - ~~`new_semantic_analyzer` hasn't been enabled yet- an internal error occurs (see bottom of comment for stacktrace) and I figured that fixing it would be a whole other unrelated problem.~~ Fixed
 - Perhaps some mypy-related documentation is in order?

### DIFF
--- a/ocfweb/account/chpass.py
+++ b/ocfweb/account/chpass.py
@@ -137,9 +137,6 @@ class ChpassForm(Form):
             label='OCF account',
         )
 
-        # mypy expects fields to be a dict, but it isn't. This is defined in django so it can't be fixed
-        # self.fields.keyOrder = field_order
-
     new_password = forms.CharField(
         widget=forms.PasswordInput,
         label='New password',
--- a/ocfweb/account/chpass.py
+++ b/ocfweb/account/chpass.py
@@ -122,6 +122,12 @@ def change_password(request: HttpRequest) -> HttpResponse:
 
 
 class ChpassForm(Form):
+    # fix self.fields.keyOrder type error in mypy
+    field_order = [
+        'ocf_account',
+        'new_password',
+        'confirm_password',
+    ]
 
     def __init__(self, ocf_accounts: List[str], calnet_uid: str, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
@@ -132,11 +138,7 @@ class ChpassForm(Form):
         )
 
         # mypy expects fields to be a dict, but it isn't. This is defined in django so it can't be fixed
-        self.fields.keyOrder = [  # type: ignore
-            'ocf_account',
-            'new_password',
-            'confirm_password',
-        ]
+        # self.fields.keyOrder = field_order
 
     new_password = forms.CharField(
         widget=forms.PasswordInput,
--- a/ocfweb/account/chpass.py
+++ b/ocfweb/account/chpass.py
@@ -3,6 +3,7 @@ from typing import Iterator
 from typing import List
 
 from django import forms
+from django.http import HttpRequest
 from django.http import HttpResponse
 from django.shortcuts import render
 from django.template.loader import render_to_string
@@ -55,7 +56,7 @@ def get_accounts_for(calnet_uid: str) -> List[Any]:
 
 
 @calnet_required
-def change_password(request: Any) -> HttpResponse:
+def change_password(request: HttpRequest) -> HttpResponse:
     calnet_uid = request.session['calnet_uid']
     error = None
     accounts = get_accounts_for(calnet_uid)
--- a/ocfweb/account/chpass.py
+++ b/ocfweb/account/chpass.py
@@ -1,4 +1,9 @@
+from typing import Any
+from typing import Iterator
+from typing import List
+
 from django import forms
+from django.http import HttpResponse
 from django.shortcuts import render
 from django.template.loader import render_to_string
 from django.utils.safestring import mark_safe
@@ -15,15 +20,14 @@ from ocfweb.auth import calnet_required
 from ocfweb.component.celery import change_password as change_password_task
 from ocfweb.component.forms import Form
 
-
 CALLINK_ERROR_MSG = (
     "Couldn't connect to CalLink API. Resetting group "
     'account passwords online is unavailable.'
 )
 
 
-def get_accounts_signatory_for(calnet_uid):
-    def flatten(lst):
+def get_accounts_signatory_for(calnet_uid: str) -> List[Any]:
+    def flatten(lst: Iterator[Any]) -> List[Any]:
         return [item for sublist in lst for item in sublist]
 
     group_accounts = flatten(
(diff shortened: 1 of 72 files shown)


ethanhs:
Ah, yeah we need to update the version of the django plugin to work with the new semantic analyzer. If you could update that and mypy itself, that would be great! I'll start reviewing this a bit later.

64bitpandas:
Alright sounds good! I've fixed the first 2 notes.

ocfweb/account/chpass.py
```diff
@@ -118,14 +123,16 @@ def change_password(request):
...
         super().__init__(*args, **kwargs)
         self.calnet_uid = calnet_uid
         self.fields['ocf_account'] = forms.ChoiceField(
             choices=[(x, x) for x in ocf_accounts],
             label='OCF account',
         )
-        self.fields.keyOrder = [
+
+        # mypy expects fields to be a dict, but it isn't. This is defined in django so it can't be fixed
+        self.fields.keyOrder = [  # type: ignore
```
ethanhs: If you move this list into the class declaration like:
```
class ChpassForm(Form):
    field_order = [...]
```

you shouldn't need this type ignore.

c.f. https://stackoverflow.com/a/34502078

ocfweb/account/register.py
```diff
@@ -61,7 +66,7 @@ def request_account(request):
     real_name = directory.name_by_calnet_uid(calnet_uid)
 
     if request.method == 'POST':
-        form = ApproveForm(request.POST)
+        form: Any = ApproveForm(request.POST)
```
ethanhs: Why do you need to annoate this as `Any`?

ocfweb/account/vhost_mail.py
```diff
@@ -69,12 +76,13 @@ def vhost_mail(request):
...
 @group_account_required
 @require_POST
-def vhost_mail_update(request):
+def vhost_mail_update(request: HttpRequest) -> HttpResponseRedirect:
     user = logged_in_user(request)
 
     # All requests are required to have these
     action = _get_action(request)
-    addr_name, addr_domain, addr_vhost = _get_addr(request, user, 'addr', required=True)
+    # _get_addr should either return a valid tuple or error so it's not necessary to verify the Optional type
```
ethanhs: You can annotate `_error` as `def _error(...) -> NoReturn`, and then remove the `Optional` from `_get_addr` (also you should probably do the same for anything that calls `_error`). 

In general if you want to unwrap an `Optional` that you know will never be `None`, you can do something like:

```
addr = _get_addr(...)
assert addr is not None
addr_name, addr_domain, addr_vhost = addr
```

ocfweb/component/markdown.py
```diff
@@ -14,24 +20,37 @@
 # tags of a format like: [[!meta title="Backups"]]
 META_REGEX = re.compile(r'\[\[!meta ([a-z]+)="([^"]*)"\]\]')
 
+# Make mypy play nicely with mixins https://github.com/python/mypy/issues/5837
```
ethanhs: A fix for this was actually just merged (https://github.com/python/mypy/pull/7860) so if you could leave a todo, to remove the work around once we update mypy that'd be great!

ocfweb/account/register.py
64bitpandas: Looking into this more, I actually found a preexisting issue.

The original reason for putting `Any` was because mypy complained that ApproveForm doesn't have an attribute `_errors`.

Turns out that assigning to the `_errors` dict directly is deprecated and `add_error()` should be used instead. 

https://stackoverflow.com/questions/188886/inject-errors-into-already-validated-form

ocfweb/account/vhost_mail.py
64bitpandas: `Optional` can't be removed from `_get_addr` because it could return `None`, but just not for that particular call. The `assert` statement works well, though. Thanks for the suggestions 😃

ocfweb/account/register.py
ethanhs: Awesome!

ocfweb/account/vhost_mail.py
ethanhs: It cannot actually, all of the cases where it doesn't return are calls to `_error`, so it is really `Union[NoReturn, ...]`, but I think it is fine to leave this as is.

ethanhs:
Thank you so much for putting in the time on this! It is greatly appreciated.

ocfweb/account/vhost_mail.py
64bitpandas: It does return `None` sometimes- I tested it by putting `assert False` at the bottom of the function instead of `return None` and it does raise an AssertionError.. So maybe `Optional[Union[NoReturn, Tuple...]]` might be preferred?

ethanhs: Oh yeah sorry if `original` is None and required is false then it returns None. I think it is fine as is.

ocfweb/component/markdown.py
BernardZhao: @64bitpandas @ethanhs can this be removed now?

ethanhs: Yup! Thank you for following up on this, mypy 0.750 just came out so it should be possible to remove this.

64bitpandas: I tried to upgrade to .750 but this seems to break mypy_django_plugin:

```
Traceback (most recent call last):
  File "venv/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "/home/b/be/bencuan/ocfweb/venv/lib/python3.7/site-packages/mypy/__main__.py", line 8, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "mypy/main.py", line 89, in main
  File "mypy/build.py", line 166, in build
  File "mypy/build.py", line 235, in _build
  File "mypy/build.py", line 2567, in dispatch
  File "mypy/build.py", line 2725, in load_graph
  File "mypy/build.py", line 1826, in __init__
  File "mypy/build.py", line 2101, in compute_dependencies
  File "mypy/plugin.py", line 785, in get_additional_deps
  File "mypy/plugin.py", line 711, in get_additional_deps
  File "/home/b/be/bencuan/ocfweb/venv/lib/python3.7/site-packages/mypy_django_plugin/main.py", line 124, in get_additional_deps
    if file.fullname() == 'django.conf' and self.django_context.django_settings_module:
TypeError: 'str' object is not callable
Makefile:26: recipe for target 'mypy' failed
make: *** [mypy] Error 1
```
I tried upgrading django and django-stubs to latest and the error didn't change. I'm not sure which codebase the error originates from?

ethanhs: Ah, yeah that is an issue. I'll see if I can make a PR to fix the plugin with the changes made in this PR to mypy https://github.com/python/mypy/pull/7829

PR #564, opened 2019-11-22
Convert as many HTTP links to HTTPS as possible

chriskuehl:
This branch is pretty much just me looking at all the `http:` links and seeing if they can be turned into `https:` links.

When I found a dead link I tried to replace it with a working link; if I couldn't find one (particularly on the "where did alumni go" page), I removed the link rather than have a broken link or a link pointing to a domain parking page. (My thinking is that these domains are probably no longer owned by the original organization and could point anywhere in the future.)

There are about 10 remaining actual HTTP links that I wasn't able to update (not counting things like XML schema references, vendored code, intentional uses of an HTTP link, etc.).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -41,7 +41,7 @@ Also available at: https://www.apache.org/licenses/LICENSE-2.0
 
                               Apache License
                         Version 2.0, January 2004
-                     http://www.apache.org/licenses/
+                     https://www.apache.org/licenses/
 
 TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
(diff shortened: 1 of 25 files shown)


LICENSE
```diff
@@ -41,7 +41,7 @@ Also available at: https://www.apache.org/licenses/LICENSE-2.0
 
                               Apache License
                         Version 2.0, January 2004
-                     http://www.apache.org/licenses/
+                     https://www.apache.org/licenses/
```
cg505: I am not sure if this change and the similar change to the GPL header constitute changes to the licenses in question: if so, that might be problematic.

PR #567, opened 2019-12-09
add STF survey page, to be set as browser homepage

cg505:


cg505:
imo it's still useful to have links to other OCF things in the same page so people know what's going on.

PR #568, opened 2019-12-12
update officers page and add committee heads

cg505:
This updates the officers page to include recent elections. It also adds committee heads to the officers page. I didn't add committee heads to the "previous officers" section, because I wasn't sure how to do it (and we don't have any official historical committee heads). We should think about if we want to do this and if so how.

Ideally this can be merged when the new terms take effect on December 21.

### DIFF
--- a/ocfweb/docs/templates/docs/officers.html
+++ b/ocfweb/docs/templates/docs/officers.html
@@ -33,7 +33,7 @@
             {% endif %}
           {% endfor %}
         </p>
-        {% if current_term.dgms|length > 0 %}
+        {% if current_term.dgms %}
           <h4>
             Deputy General Managers
           </h4>
@@ -62,7 +62,7 @@
             {% endif %}
           {% endfor %}
         </p>
-        {% if current_term.dsms|length > 0 %}
+        {% if current_term.dsms %}
           <h4>
             Deputy Site Managers
           </h4>
@@ -80,7 +80,7 @@
           the OCF lab and other services. They are elected by the Board of
           Directors at the end of each semester.
         </p>
-      {% if current_term.heads|length > 0 %}
+      {% if current_term.heads %}
         <h3>
           Committee Heads
         </h3>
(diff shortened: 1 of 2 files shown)


cg505:
This is how it looks:
Screenshot_2019-12-12 Officers – Open Computing Facility

64bitpandas:
There's a typo in the date for ethanhs- should be /19 for the year right?

ethanhs:
>There's a typo in the date for ethanhs- should be /19 for the year right?

Ha. I'll never reveal my time traveling secrets!

cg505:
@64bitpandas good catch, fixed

ocfweb/docs/templates/docs/officers.html
```diff
@@ -80,6 +80,26 @@ <h4>
           the OCF lab and other services. They are elected by the Board of
           Directors at the end of each semester.
         </p>
+      {% if current_term.heads|length > 0 %}
```
kpengboy: Is it possible to just say `{% if current_term.heads %}`?

ocfweb/docs/views/officers.py
```diff
@@ -24,12 +25,17 @@ def Term(
     sms: List[Any],
     dgms: Optional[List[Any]] = None,
     dsms: Optional[List[Any]] = None,
+    heads: Optional[List[Tuple[str, List[Any]]]] = None,
```
kpengboy: Ideally we'd make the `Any`s more specific. Perhaps this can wait until we can use Python 3.7 data classes.

dkess: This will get refactored anyways, assuming someone works on https://github.com/ocf/etc/issues/3

ocfweb/docs/templates/docs/officers.html
cg505: Not totally sure, but this is following the pattern of lines 36 and 65. I'll try to test it.

PR #570, opened 2020-01-29
update SSH host keys for tsunami

cg505:
shows how much these get used

### DIFF
--- a/ocfweb/docs/docs/services/shell.md
+++ b/ocfweb/docs/docs/services/shell.md
@@ -15,10 +15,14 @@ Most SSH/SFTP clients will prompt you to accept an unknown key when you first
 connect. Our SSH fingerprint can be used to verify that you're connecting to
 the correct server:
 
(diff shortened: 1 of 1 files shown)


cg505:
I think they were wrong since we upgraded to buster lol.

jvperrin:
They look the same to me. We place these [from the puppet private share onto disk](https://github.com/ocf/puppet/blob/master/modules/ocf/manifests/hostkeys.pp), so they shouldn't change if tsunami is reprovisioned or anything. I think this change is good though since it adds more hash formats and makes then more consistent across types of keys, but I don't think they were actually _wrong_ before, if that makes sense.

cg505:
Ah, you're right. We didn't have the ed25519 key before and ssh uses sha256 by default now so this is still good to have.

cg505:
Also, I meant to say stretch instead of buster. edited.

PR #573, opened 2020-02-02
Upgrade pygments, don't try to guess a lexer

jvperrin:
With newer versions of pygments, trying to guess a lexer was giving some lexers like `MIMELexer`, `TeraTermLexer`, or `TransactSqlLexer` that are definitely not what we want. Instead of trying to be too smart and guess, just default to `text` (the `TextLexer`) unless a specific language is passed in. If a specific language is used and there is no lexer that matches, don't catch the error so it bubbles up and is clear to the person writing the page that they have chosen an invalid lexer option.

In doing this, I also discovered that we were using invalid pygments languages in a few places. `htaccess` should be `apache`/`apacheconf`, and `haproxy` doesn't exist. I also removed occurrences of `text` as a choice since that is now the default. See https://pygments.org/docs/lexers/ for more details on available lexers.


ja5087:
oops, it seems I forgot that 'c' comes before 'g'

PR #550, opened 2019-10-11
Delete section on giving people Comodo SSL

fydai:
Letsencrypt exists now!

PR #586, opened 2020-03-16
Remove InCommon access from staff privileges doc

jvperrin:
We don't use InCommon any more since we're now using entirely Let's Encrypt (I guess we missed this in both #572 and #550?)


jvperrin:
Ah yeah we shouldn't need either of those any more. The rt ones could even change too, that doc is pretty outdated now anyway (although semi-frequently found by people online setting up rt).

PR #587, opened 2020-03-22
update info on ldap-add-host

cg505:
Also, add a link to the IP allocation spreadsheet.

See ocf/utils#142.

### DIFF
--- a/ocfweb/docs/docs/staff/procedures/new-host.md
+++ b/ocfweb/docs/docs/staff/procedures/new-host.md
@@ -13,11 +13,14 @@ running `virsh undefine hozer-{num}` to remove the VM and `lvremove
 ## Step 0. Pick a hostname and IP
 
 If you are creating a brand-new host, you can find a list of IP addresses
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/staff/procedures/new-host.md
```diff
@@ -29,9 +32,11 @@ Unfortunately, these steps tend to change a lot as our infrastructure evolves.
 
 ### Step 1.1. Add the LDAP entry
 
-On supernova, `kinit $USER/admin ldap-add-host <hostname> <ip>`. If setting up
-a desktop, also do `kinit $USER/admin ldapvi cn=<hostname>` and set the `type`
-attribute to `desktop`. If doing a staff VM, set it to `staffvm` instead.
+On supernova, `kinit $USER/admin ldap-add-host <hostname> <ip-last-octet>`.
```
ja5087: This is not really an issue with this PR but somewhere we should explain our IP scheme (using the same last octet), which explains the arguments to ldap add host

PR #591, opened 2020-04-21
Allow groups to sign up via web interface

ja5087:
Sure there's only a few weeks left in the online semester but this would be useful over breaks and during unexpected extended lab closures.

I've mostly tested this by hand up except the part about creating the actual group account since that would need actual credentials, but will make sure to test before/if we merge.

### DIFF
--- a/ocfweb/account/constants.py
+++ b/ocfweb/account/constants.py
@@ -14,6 +14,7 @@ TESTER_CALNET_UIDS = (
     1101587,  # jvperrin
     1511731,  # ethanhs
     1623751,  # cooperc
+    1619256,  # jaw
 )
 
 # comma separated tuples of CalLink OIDs and student group names
@@ -21,4 +22,5 @@ TEST_GROUP_ACCOUNTS = (
     (91740, 'The Testing Group'),  # needs to have a real OID, so boo
     (46187, 'Open Computing Facility'),  # good old ocf
     (46692, 'Awesome Group of Awesome'),  # boo another real OID
+    (92029, 'eXperimental Computing Facility'),  # our sister org
 )
--- a/ocfweb/account/register.py
+++ b/ocfweb/account/register.py
@@ -1,4 +1,3 @@
-import operator
 from typing import Any
 from typing import Union
 
@@ -42,14 +41,15 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
     status = 'new_request'
 
     existing_accounts = search.users_by_calnet_uid(calnet_uid)
-    eligible_group_signatories = groups_by_student_signat(calnet_uid)
+    eligible_signatory_groups = groups_by_student_signat(calnet_uid)
 
+    # disallow creation forr groups already have accounts
     existing_group_accounts = {}
-    for group_oid in list(eligible_group_signatories.keys()):
-        if group_by_oid(group_oid) and group_oid not in map(operator.itemgetter(0), TEST_GROUP_ACCOUNTS):
-            existing_group_accounts[group_oid] = eligible_group_signatories.pop(group_oid)
+    for group_oid in list(eligible_signatory_groups.keys()):
+        if len(group_by_oid(group_oid)['accounts']) and group_oid not in [group[0] for group in TEST_GROUP_ACCOUNTS]:
+            existing_group_accounts[group_oid] = eligible_signatory_groups.pop(group_oid)
 
-    if existing_accounts and not eligible_group_signatories and calnet_uid not in TESTER_CALNET_UIDS:
+    if existing_accounts and not eligible_signatory_groups and calnet_uid not in TESTER_CALNET_UIDS:
         return render(
             request,
             'account/register/already-has-account.html',
@@ -77,7 +77,7 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
     association_choices = []
     if not existing_accounts or calnet_uid in TESTER_CALNET_UIDS:
         association_choices.append((calnet_uid, real_name))
-    for group_oid, group in eligible_group_signatories.items():
+    for group_oid, group in eligible_signatory_groups.items():
         association_choices.append((group_oid, group['name']))
 
     if request.method == 'POST':
@@ -87,7 +87,7 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
             if is_group_account:
                 req = NewAccountRequest(
                     user_name=form.cleaned_data['ocf_login_name'],
-                    real_name=real_name,
+                    real_name=eligible_signatory_groups[form.cleaned_data['account_association']],
                     is_group=True,
                     calnet_uid=None,
                     callink_oid=form.cleaned_data['account_association'],
--- a/ocfweb/account/register.py
+++ b/ocfweb/account/register.py
@@ -41,15 +41,16 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
     status = 'new_request'
 
     existing_accounts = search.users_by_calnet_uid(calnet_uid)
-    eligible_signatory_groups = groups_by_student_signat(calnet_uid)
+    groups_for_user = groups_by_student_signat(calnet_uid)
 
-    # disallow creation forr groups already have accounts
-    existing_group_accounts = {}
-    for group_oid in list(eligible_signatory_groups.keys()):
-        if len(group_by_oid(group_oid)['accounts']) and group_oid not in [group[0] for group in TEST_GROUP_ACCOUNTS]:
-            existing_group_accounts[group_oid] = eligible_signatory_groups.pop(group_oid)
+    eligible_new_group_accounts, existing_group_accounts = {}, {}
+    for group_oid in groups_for_user:
+        if not group_by_oid(group_oid)['accounts'] or group_oid in [group[0] for group in TEST_GROUP_ACCOUNTS]:
+            eligible_new_group_accounts[group_oid] = groups_for_user[group_oid]
+        else:
+            existing_group_accounts[group_oid] = groups_for_user[group_oid]
 
-    if existing_accounts and not eligible_signatory_groups and calnet_uid not in TESTER_CALNET_UIDS:
+    if existing_accounts and not eligible_new_group_accounts and calnet_uid not in TESTER_CALNET_UIDS:
         return render(
             request,
             'account/register/already-has-account.html',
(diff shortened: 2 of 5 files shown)


encadyma:
Wow, this is pog

ocfweb/account/register.py
```diff
@@ -37,8 +42,14 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
     status = 'new_request'
 
     existing_accounts = search.users_by_calnet_uid(calnet_uid)
+    eligible_group_signatories = groups_by_student_signat(calnet_uid)
 
-    if existing_accounts and calnet_uid not in TESTER_CALNET_UIDS:
+    existing_group_accounts = {}
+    for group_oid in list(eligible_group_signatories.keys()):
```
cg505, line 48: why do we need `list` here? because we modify the dict later in the loop?

ocfweb/account/register.py
```diff
@@ -37,8 +42,14 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
+        if group_by_oid(group_oid) and group_oid not in map(operator.itemgetter(0), TEST_GROUP_ACCOUNTS):
```
cg505, line 49: why do we need to do `group_by_oid(group_oid)` here? isn't the group already guaranteed to exist from line 45?

also ik its less functional but can we just use a list comprehension? my brain hurt

ocfweb/account/register.py
```diff
@@ -37,8 +42,14 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
+            existing_group_accounts[group_oid] = eligible_group_signatories.pop(group_oid)
```
cg505, line 50: ~can't we just `[group_oid]` normally instead of using `pop`~

oh i get it, we are removing groups with accounts that already exist. this code was rly confusing imo but I'm not sure if there's anything to be done about it

ocfweb/account/register.py
```diff
@@ -37,8 +42,14 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
     status = 'new_request'
 
     existing_accounts = search.users_by_calnet_uid(calnet_uid)
+    eligible_group_signatories = groups_by_student_signat(calnet_uid)
```
cg505, line 45: variable name makes it sound like a list of signatories, but it's really a list of groups. maybe `eligible_signatory_groups`?

ocfweb/account/templates/account/register/index.html
```diff
@@ -41,21 +41,36 @@
...
+                {% if existing_group_accounts %}
+                <p><strong>Existing OCF group accounts</strong>
+                    {% for account in existing_group_accounts.values %}
+                        <p>{{account.name}}: {{account.accounts|join:", "}}</p>
+                    {% endfor %}
+                {% endif %}
+                <hr />
+
+                <p>You should select the identity that your new account will be associated with. This is either yourself or an RSO that you are a signatory for. Note that identities already associated to an OCF account will not show up here.</p>
+                <p id="account-association">{{form.account_association|bootstrap}}</p>
```
cg505, line 67: This should be hidden for people that are not signatories.

ocfweb/account/templates/account/register/index.html
```diff
@@ -41,21 +41,36 @@
...
                 <hr />
 
                 <div class="container-fluid">
                     <div class="row">
                         <div class="col-sm-7">
-                            <p>Your account name must consist of between three and 16 lowercase letters (no spaces, numbers, underscores, or other symbols), and it <strong><strong>must</strong> be based on your real name or initials.</strong></p>
+                            <p>Your account name must consist of between three and 16 lowercase letters (no spaces, numbers, underscores, or other symbols), and it <strong>must be based on your real/group name or initials.</strong></p>
```
cg505, line 73: wording is confusing. maybe:

> must be based on your real name or initials, or on your group's registered name

ocfweb/account/register.py
```diff
@@ -63,22 +74,44 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
-                    form.cleaned_data['password'],
-                    RSA.importKey(CREATE_PUBLIC_KEY),
-                ),
-                handle_warnings=NewAccountRequest.WARNINGS_WARN,
-            )
+            is_group_account = form.cleaned_data['account_association'] != calnet_uid
+            if is_group_account:
+                req = NewAccountRequest(
+                    user_name=form.cleaned_data['ocf_login_name'],
+                    real_name=real_name,
```
cg505, line 90: seems... wrong

ocfweb/account/register.py
```diff
@@ -63,22 +74,44 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
+                    is_group=True,
+                    calnet_uid=None,
+                    callink_oid=form.cleaned_data['account_association'],
```
cg505, line 93: does django validate server-side that this must be one of the choices? (in other words, can I manually craft a request to register an org as an arbitrary oid?

ocfweb/account/constants.py
```diff
@@ -14,11 +14,13 @@
...
     1623751,  # cooperc
+    1619256,  # jaw
 )
 
 # comma separated tuples of CalLink OIDs and student group names
 TEST_GROUP_ACCOUNTS = (
     (91740, 'The Testing Group'),  # needs to have a real OID, so boo
     (46187, 'Open Computing Facility'),  # good old ocf
     (46692, 'Awesome Group of Awesome'),  # boo another real OID
+    (92029, 'eXperimental Computing Facility'),  # our sister org
```
ethanhs, line 25: ♥

kpengboy:
A question about the plan here: is this intended to be disabled when normal times return?

ja5087:
I would say since campus will still be optionally remote next semester, we should offer it at least until then. After that, I'm not sure yet.

ocfweb/account/register.py
```diff
@@ -42,14 +41,15 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
-    eligible_group_signatories = groups_by_student_signat(calnet_uid)
+    eligible_signatory_groups = groups_by_student_signat(calnet_uid)
 
+    # disallow creation forr groups already have accounts
     existing_group_accounts = {}
-    for group_oid in list(eligible_group_signatories.keys()):
-        if group_by_oid(group_oid) and group_oid not in map(operator.itemgetter(0), TEST_GROUP_ACCOUNTS):
-            existing_group_accounts[group_oid] = eligible_group_signatories.pop(group_oid)
+    for group_oid in list(eligible_signatory_groups.keys()):
+        if len(group_by_oid(group_oid)['accounts']) and group_oid not in [group[0] for group in TEST_GROUP_ACCOUNTS]:
```
cg505, line 49: empty list is already falsy (no need for len). thanks for the clarification on this tho

ocfweb/account/templates/account/register/index.html
ja5087: I actually think we should just leave it in. It lets users know that if they come back to this form they will have the option to select, and there will only be one option if they aren't a signatory. I'd like to hear your reasoning though.

ocfweb/account/templates/account/register/index.html
```diff
@@ -72,8 +91,8 @@
                 <p>
                     Your password should be unique to your OCF account and not
                     shared with any other accounts, including your Berkeley
-                    CalNet account. <em>Never share your password with anyone,
-                    including OCF staff members.</em>
+                    CalNet account. <em>Never share your user account password
+                    with anyone, including OCF staff members.</em>
                 </p>
                 {{form.password|bootstrap}}
                 {{form.verify_password|bootstrap}}
```
MrMinos, line 98: For group accounts, it would be good to add a wording about using a contact email address that will last between club member generations

MrMinos: Sorry, nvm - should have read the entire thread first before code

ja5087:
BoD voted to allow this on 2020-05-04. The length of time is yet undetermined but will be at least for the duration of the fall semester.

ocfweb/account/register.py
```diff
@@ -41,15 +41,16 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
-    existing_group_accounts = {}
-    for group_oid in list(eligible_signatory_groups.keys()):
-        if len(group_by_oid(group_oid)['accounts']) and group_oid not in [group[0] for group in TEST_GROUP_ACCOUNTS]:
-            existing_group_accounts[group_oid] = eligible_signatory_groups.pop(group_oid)
+    eligible_new_group_accounts, existing_group_accounts = {}, {}
+    for group_oid in groups_for_user:
+        if not group_by_oid(group_oid)['accounts'] or group_oid in [group[0] for group in TEST_GROUP_ACCOUNTS]:
+            eligible_new_group_accounts[group_oid] = groups_for_user[group_oid]
+        else:
+            existing_group_accounts[group_oid] = groups_for_user[group_oid]
```
cg505, line 51: thanks, i like this a lot more

ocfweb/account/register.py
```diff
@@ -77,20 +78,21 @@ def request_account(request: HttpRequest) -> Union[HttpResponseRedirect, HttpRes
...
-    for group_oid, group in eligible_signatory_groups.items():
+    for group_oid, group in eligible_new_group_accounts.items():
         association_choices.append((group_oid, group['name']))
 
     if request.method == 'POST':
         form = ApproveForm(request.POST, association_choices=association_choices)
         if form.is_valid():
-            is_group_account = form.cleaned_data['account_association'] != calnet_uid
+            assoc_id = form.cleaned_data['account_association']
+            is_group_account = assoc_id != calnet_uid
```
cg505, line 88: we pray that callink oid and calnet uid never overlap... fair enough

ocfweb/account/templates/account/register/index.html
cg505: If it automatically selects your user when it's the only option, it's probably okay. We just need to keep in mind that normal user account creation is like two order of magnitudes more common than group account creation.

ocfweb/account/templates/account/register/index.html
```diff
@@ -65,6 +68,7 @@
 
                 <p>You should select the identity that your new account will be associated with. This is either yourself or an RSO that you are a signatory for. Note that identities already associated to an OCF account will not show up here.</p>
                 <p id="account-association">{{form.account_association|bootstrap}}</p>
+                <div class="alert association-feedback" role="alert" id="association-feedback"></div>
```
cg505, line 71: Thanks, I like this approach.

ocfweb/account/templates/account/register/index.html
```diff
@@ -41,21 +41,40 @@
...
+                <p>You should select the identity that your new account will be associated with. This is either yourself or an RSO that you are a signatory for. Note that identities already associated to an OCF account will not show up here.</p>
+                <p id="account-association">{{form.account_association|bootstrap}}</p>
+                <div class="alert association-feedback" role="alert" id="association-feedback"></div>
                 <hr />
 
                 <div class="container-fluid">
                     <div class="row">
                         <div class="col-sm-7">
-                            <p>Your account name must consist of between three and 16 lowercase letters (no spaces, numbers, underscores, or other symbols), and it <strong><strong>must</strong> be based on your real name or initials.</strong></p>
+                            <p>Your account name must consist of between three and 16 lowercase letters (no spaces, numbers, underscores, or other symbols), and it <strong>must be based on your real name or initials, or on your group's registered name.</strong></p>
```
cg505, line 77: Bonus points, hide `, or on your group's registered name` if there are no new eligible group accounts.

Issue #593, opened 2020-04-28
Remove Google Analytics (w/ or w/o replacement)


nikhiljha:
Discussion on IRC, starting here https://irclogs.ocf.berkeley.edu/rebuild/2020-04-28#807157

Summary
- We aren't actually looking at the data.
- It might not be the best (privacy-wise) to have google analytics.

PotatoParser:
Further discussion of removing Google Analytics is needed

dkess:
I would say: bring it up at BoD, if no one defends analytics, remove it and don't bother looking into open-source things to replace it with.

BernardZhao:
Noone defended it in BoD, and we voted to have it removed.

PR #594, opened 2020-04-28
Remove Google Analytics

PotatoParser:
Resolves @ #593


PotatoParser:
Yeah, I'm just leaving this here until it's decided

ja5087:
As a side note, I talked to Elle and it seems opstaff aren't actively using this either. Let's bring this up at bod

PotatoParser:
Approved by BOD

PR #595, opened 2020-05-03
Update vhost config instructions ot use ocf/etc

jvperrin:
Since ocf/etc#108 these configs are in the ocf/etc and we're switching that to be the source of truth. I also updated some of the language here around how quickly updates for new vhosts/apphosts/mailhosts go through the pipeline and tested viewing the page through a local copy of ocfweb to make sure it looks reasonable and the new links load.

### DIFF
--- a/ocfweb/docs/docs/staff/procedures/vhost.md
+++ b/ocfweb/docs/docs/staff/procedures/vhost.md
@@ -30,12 +30,13 @@
 
 ### Web
 
-Edit the file `~staff/vhost/vhost.conf`, adding a new line. The format is
-documented at the top.
+Edit the file `configs/vhost.conf` in the [`ocf/etc` repo][ocf-etc], adding new
+entries at the top. The format is documented at the top of that file.
 
-This takes effect at the top of every hour when a cronjob runs. HTTPS takes
-about an additional hour to take effect (for the first hour, it will be
-HTTP-only).
+This takes effect at the top of every hour when a cronjob runs. HTTPS should
+available shortly afterwards (within 5-10 minutes). Keep in mind that vhosts
+are not available without HTTPS, so there may be a short period of time where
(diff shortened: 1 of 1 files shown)


jvperrin:
Oh, one thing I ran into with this is that updates to mail hosts are going to be much slower (not instant, around 30 minutes now) because `ocf/etc` syncing is locked to puppet updates. I'm quite tempted to add a rsync layer or something on top of it to minimize the time there are inconsistencies with updates only on one host running ocfweb but not on others for instance.

For web vhosts I thought 2 hours was pretty slow, so I opened https://github.com/ocf/puppet/pull/958 to speed that up a bit (2 hours => 1 hour 5 mins or so). I've updated the language here to say that's the case but that review isn't actually merged yet.

cg505:
we can also update build-vhosts to run puppet-trigger

jvperrin:
I assume you mean `build-vhosts` to `lets-encrypt-update`? Anyway, yeah that'd probably be better than just moving them closer together. Or we could have updates to `ocf/etc` trigger puppet on some key hosts like kubernetes workers, but that might be a bit too complicated.

cg505:
no, I mean have build-vhosts run puppet-trigger at the start to update etc

jvperrin:
Ohh, I see... Hmmm, that'd be interesting but I don't think it fully fixes the problems with vhost config updates. The web ones don't really change much since there should be a couple puppet updates between each vhost rebuild anyway, and that is still an hour. But the apphosting ones (previously every 10 minutes) or the mail ones (previously instant since they just gate off the ocfweb page) are affected much more, so those are the ones I'm more concerned about.

Making the web ones more frequent also sounds like a good move, but that's a change in itself that's not really related to the move of the configs into ocf/etc if that makes sense.

PR #598, opened 2020-05-11
add ~njha to pr completion

nikhiljha:
self referential pr


cg505:
wow, did you just imply the PR number by looking at the next open number? massive brain

PR #600, opened 2020-05-12
Finally move my live disk resizing guide to the docs

jvperrin:
Finally! Over 3 years as a gist is probably enough time :P

This should replace the usage of https://gist.github.com/jvperrin/776a45dc68bf9fd91b3dae47b4aee32a and I'll try and edit that guide to point to the official docs once this is merged.

I did make a decent number of changes in content (displaying more sample command output, adding an assumptions section, some organization, etc.), so I'd appreciate any read throughs of it all!

### DIFF
--- a/ocfweb/docs/docs/staff/backend/libvirt.md
+++ b/ocfweb/docs/docs/staff/backend/libvirt.md
@@ -137,21 +137,8 @@ and for vCPUs:
 
 ### How do I edit my VM's disk size?
 
-Find the lvm vg
-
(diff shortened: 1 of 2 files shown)


jvperrin:
@ja5087 Ah, very good point, thanks for noticing that! I didn't realize that page had a section on disk size editing so I'll include a reference there to this doc and include it in this review. Better now than later when I've forgotten or something.

jvperrin:
Ah thanks, meant to merge this before but didn't do so yet.

PR #601, opened 2020-05-13
Added num users and staff to api

BernardZhao:
Some small additions to the API.

### DIFF
--- a/ocfweb/api/urls.py
+++ b/ocfweb/api/urls.py
@@ -14,8 +14,8 @@ urlpatterns = [
     path('hours/staff', staff_hours.get_staff_hours, name='staff_hours'),
     path('hours/today', hours.get_hours_today, name='hours_today'),
     path('lab/desktops', lab.desktop_usage, name='desktop_usage'),
+    path('lab/num_users', stats.get_num_users_in_lab, name='get_num_users_in_lab'),
+    path('lab/staff', stats.get_staff_in_lab, name='get_staff_in_lab'),
     path('session/log', session_tracking.log_session, name='log_session'),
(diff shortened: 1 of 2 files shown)


nikhiljha:
Wait... weren't we planning to separate the API out from ocfweb?

ocfweb/api/urls.py
```diff
@@ -15,4 +16,6 @@
     path('lab/desktops', lab.desktop_usage, name='desktop_usage'),
     path('session/log', session_tracking.log_session, name='log_session'),
     path('shorturl/<path:slug>', shorturls.bounce_shorturl, name='bounce_shorturl'),
+    path('stats/num_users', stats.get_num_users_in_lab, name='get_num_users_in_lab'),
+    path('stats/staff', stats.get_staff_in_lab, name='get_staff_in_lab'),
```
cg505, line 20: imo lets make these make more sense, like `stats/lab/num_users` and `stats/lab/current_staff`. `stats/num_users` sounds like the total number of accounts or something

jvperrin: Agreed, they'd fit well under the `lab` path that `lab/desktops` is already using, maybe `lab/stats/num_users`/`lab/stats/current_staff` instead of `stats/lab`?

BernardZhao: I actually originally had it as `lab/num_users`, I'll just revert it to that.

BernardZhao:
> Wait... weren't we planning to separate the API out from ocfweb?

Yes, and I am still planning to. However these endpoints didn't previously exist in the first place. Once I have everything I need then hopefully it will be easier to translate into something new.

PR #602, opened 2020-05-17
add Fall 2020 officers

cg505:
Also, update the Officers page to support historical committee heads.

### DIFF
--- a/ocfweb/docs/templates/docs/officers.html
+++ b/ocfweb/docs/templates/docs/officers.html
@@ -85,7 +85,7 @@
           Committee Heads
         </h3>
           {% for committee in current_term.heads %}
-            <h4>{{ committee.name }} Committee Heads</h4>
+            <h4>{{ committee.name }} Heads</h4>
               <p>
                 {% for head in committee.heads %}
                   {% if head.end is None %}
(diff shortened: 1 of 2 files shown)


ocfweb/docs/templates/docs/officers.html
```diff
@@ -142,6 +143,24 @@ <h5>Deputies:</h5>
...
+          {% if term.heads %}
+            <tr>
+              <td style="border: none;" />
+              <td colspan="2">
+                <details>
+                  <summary style="display: list-item;">
+                    Committee Heads
+                  </summary>
+                  {% for committee in term.heads %}
+                    <h5>{{ committee.name }} Heads</h5>
```
dkess, line 155: Probably don't need the word "Heads" on this line, the word head is used a lot

cg505: I am worried if we do not put "Heads" here then people will think the heads are the whole committee.

dkess: The word is still used in the dropdrown menu so I don't think it'll have that effect

cg505:
sorry im so slow

ocfweb/docs/templates/docs/officers.html
cg505: ok, done

PR #603, opened 2020-05-18
Use dataclasses and stricter typing in officers module

kpengboy:
This PR builds off #602. Please comment if I have done something unadvisable or unidiomatic (I am not that experienced in Python typing).

### DIFF
--- a/ocfweb/docs/views/officers.py
+++ b/ocfweb/docs/views/officers.py
@@ -1,8 +1,9 @@
 import math
-from collections import namedtuple
+from dataclasses import dataclass
 from datetime import date
 from typing import Any
 from typing import Callable
+from typing import cast
 from typing import List
 from typing import Optional
 from typing import Tuple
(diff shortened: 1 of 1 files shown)


ocfweb/docs/views/officers.py
```diff
@@ -73,7 +98,7 @@ def __str__(self) -> str:
                 s += ' (acting)'
         if not self.full_term:
             s += ' ({}–{})'.format(
-                self.start.strftime('%m/%d/%y'),
+                cast(date, self.start).strftime('%m/%d/%y'),
```
ethanhs, line 101: Generally `cast`s are used a tool of last resort. For something like this, I'd recommend doing something like
```
assert self.start is not None
... # self.start should be a date from here on
since it seems that `self.start` could be `None`?

File: ocfweb/docs/views/officers.py [hunk 2]
```diff
@@ -15,39 +16,63 @@
...
     @classmethod
-    def from_uid_or_info(cls: Callable[..., Any], uid_or_info: Union[Tuple[Any, ...], str]) -> Any:
+    def from_uid_or_info(cls: 'Callable[..., Officer]', uid_or_info: OfficerUidOrInfo) -> 'Officer':
+        start: Optional[date]
+        end: Optional[date]
+
         if isinstance(uid_or_info, tuple):
             if len(uid_or_info) == 3:
-                uid, start, end = uid_or_info
+                uid, start, end = cast(Tuple[str, date, date], uid_or_info)
```
ethanhs, line 72: Perhaps this could be made a `typing.NamedTuple`?

ocfweb/docs/views/officers.py
kpengboy: Thanks, done.

ocfweb/docs/views/officers.py
kpengboy: It seems that if I were to do that, I would have to add more boilerplate to the actual list of officers, right? I'm not sure I would like to do so at this time...

ethanhs: Well, you could do something like `OfficerInfo(*uid_or_tupl)`, but thinking about it more, its probably not worth it.

PR #607, opened 2020-06-08
optimize svg

nikhiljha:
micro-optimization at its finest, saved a few KB off everything


nikhiljha:
Screen Shot 2020-06-08 at 3 51 42 PM

Nooo we must save the extra two bytes! Down with `\n` in svg!

dkess:
lmao `python.png` got bigger

nikhiljha:
that one should be an svg anyway, I should look at replacing them

ja5087:
If you run `make install-hooks` it'll run pre-commit before you, well, commit so Jenkins won't get angry.

PR #609, opened 2020-06-19
blm changes

bjb568:


bjb568:
I broke something idk.

ja5087:
Yep, see https://jenkins.ocf.berkeley.edu/blue/organizations/jenkins/ocfweb/detail/blm/3/pipeline/#step-47-log-945

ocfweb/templates/base.html
```diff
@@ -100,24 +100,11 @@
...
-                            {% endif %}
-                            {{current_lab_hours|lab_hours_time}}
-                            on {% now "l" %} {% lab_hours_holiday holidays %}
-                            <a class="subtle" href="{% url 'doc' 'services/lab' %}">more &raquo;</a>
-                        </span>
-                    {% else %}
-                        <strong>Lab Temporarily Closed</strong>
-                        <a class="subtle" href="{% url 'home' %}">more &raquo;</a>
-                    {% endif %}
+                    <strong>Defund the police.</strong>
```
ja5087, line 103: I would focus on the movement and steer away from potentially political statements.

ja5087: My personal opinions notwithstanding

bjb568: The movement is political tho, and a common criticism of these kinds of messages is that they want change but aren't clear about what change. "Defund the police" is a bold statement, but one that is apt given current events, clear, and actionable. "Black lives matter", while a more common slogan for tech websites to put up, is much more vague and uncontroversial by virtue of being vague.

Baisang: If you are speaking on behalf of an organization, you should make sure everyone is on board with the message you are presenting

pxhanus: While I disagree with others' statements about avoiding politics... We live in a society where the topic of BLM tends is unfortunately inherently political... I think instead my issue with it is that while I would happily approve that it could go into the body I see this like a paper in the sense that this header should be like a topic sentence or title and encapsulate everything we want to say.

bjb568:
yeet

PR #618, opened 2020-07-02
Allocate ptty

ja5087:
Not sure why this is but at some point `makemysql` broke with the following error:

`tee: /dev/tty: No such device or address`

I've not root caused this or figured out when it broke, but a workaround for now is to force paramiko to allocate a pseudo tty using get_pty. This actually also combines stdout and stderr but thats ok for now.


ja5087:
Paramiko is a pretty lightweight wrapper so it could be OpenSSH behavior that changed. It probably worked at some point.

PR #619, opened 2020-07-04
Sort the columns in the Today's Lab Utilization table when clicking on table heads

Hexhu:
When a table head is clicked (like "% Busy" or "Desktop"), the column will be sorted in ascending order. Click again to sort in descending order. Hold the "Control" key while clicking to reset the column order to default.

A demo is available at [trojan.ocf.berkeley.edu:8794/stats/](http://trojan.ocf.berkeley.edu:8794/stats/). Scroll down and click on the "% Busy" header to see the effect.

### DIFF
--- a/ocfweb/stats/templates/stats/summary.html
+++ b/ocfweb/stats/templates/stats/summary.html
@@ -75,7 +75,7 @@
     </div>
 {% endblock %}
 
-{% block sortable_lab_utilization %}
+{% block inline_js %}
 <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.tablesorter/2.31.3/js/jquery.tablesorter.min.js"></script>
 <script>
 $(document).ready(function() {
--- a/ocfweb/stats/templates/stats/summary.html
+++ b/ocfweb/stats/templates/stats/summary.html
@@ -49,24 +49,28 @@
 
         <h3>Today's Lab Utilization</h3>
         <table id="todays-lab-utilization" class="table table-striped table-hover">
-            <tr>
-                <th>Desktop</th>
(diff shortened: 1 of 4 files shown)


ja5087:
Thanks for the change!

Frontend for reference:

image

Can we add some UI elements to indicate that this is sortable? Right now it's not clear that it is, maybe something like 
image would work.

ocfweb/stats/templates/stats/summary.html
```diff
@@ -48,25 +48,38 @@ <h4>{{printer}}</h4>
...
+                {% endfor %}
+            </tbody>
         </table>
     </div>
 {% endblock %}
+
+{% block inline_js %}
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.tablesorter/2.31.3/js/jquery.tablesorter.min.js"></script>
+<script>
+$(document).ready(function() {
```
dkess, line 81: This line can be replace with just `$(function() {`

Hexhu: Shortened it. They are the same though. Plz review :3

Hexhu:
TODO: add arrow images to table heads to indicate that the columns are sortable
- pick relevant `tablesorter-header` section from https://github.com/Mottie/tablesorter/blob/master/css/theme.default.css
- use arrows to indicate `headerSortUp`, `headerSortDown` and `sorter-false` states
- save style at `ocfweb/static/scss/pages/stats/_summary.scss`

Hexhu:
Visual indicator added @ja5087 
Demo at http://trojan.ocf.berkeley.edu:8794/stats/

PR #627, opened 2020-08-01
change map to static

ja5087:
image

image

frontend is my passion


chriskuehl:
Is it meant to be that narrow?

jvperrin:
I think to get a replacement for the 100% width (since the width needs to be provided to the staticmap API call), you'd need to add some JS to get the element width and set it from that? At least that's definitely much simpler than the non-static version of the map and all it does.

ja5087:
so originally I made it more narrow since it was going to run into issues when screens were more narrow (kinda lazy I know). This should look a lot better.

PR #508, opened 2019-04-30
docs: Add Kubernetes deploy HOWTO

keur:

PR #631, opened 2020-08-10
Reviving service deployment docs

BernardZhao:
Decided to put some work on trying to update and bring back the work @keur made in #508, especially since @pxhanus could use some documentation as to how to deploy the discordbridge. These are some preliminary changes, and I still need a lot of help describing parts of the process I don't understand as well.
* How should I go about formatting the Markdown? I found the existing file to be well spaced, but `make lint` didn't seem to do the trick.
* Are there some parts of Keur's original documentation that aren't true anymore? I tried to not cut stuff that I didn't 100% know was not relevant anymore, but I really ended up cutting nothing.

### DIFF
--- a/ocfweb/docs/docs/staff/procedures/deploying-services.md
+++ b/ocfweb/docs/docs/staff/procedures/deploying-services.md
@@ -1,0 +1,373 @@
+[[!meta title="Deploying Services"]]
+
+## Overview
+Are you looking to deploy a new service to the OCF Kubernetes cluster? This
+document will cover all of the steps required to do so. Do note that this is **not** a
+substitute for a Kubernetes tutorial or a Docker tutorial _(there are many
+resources online for that)_ but a guide for getting your service running on the
+OCF. There are a lot of pieces of OCF infrastructure at play here, but this
+should serve as a gentle introduction into how things are done.
+
+This `HOWTO` uses one of the OCF's simplest services as an example:
+[Templates][templates]. Templates is service used internally by OCF staff
+serving 'copy-pasteable' email templates. You can use [[git|doc
+staff/backend/git]] to `clone` this [repo](https://github.com/ocf/templates/blob/master/kubernetes/templates.yml.erb)
+, or just follow along looking at the files. If you want to blaze ahead,
+this repo should serve as a good example of the required components in the
+repo for deployment.
+
+## Prerequisites
+If you're looking to deploy your service, you should first have a service to
+deploy. This is normally done with a repository underneath the
+[OCF's Github organization](https://github.com/ocf). If you have the correct
+permissions, you should be able to transfer, fork, or just create a
+repository in the org.
+
+### Docker & Jenkins
+
+For your service to be deployed onto the Kubernetes cluster, it must have
+a `Dockerfile` so that we can containerize the service. Nothing OCF specific
+here, just include whatever your service needs.
+
+[[Jenkins|doc staff/backend/jenkins]] will try to build that container,
+and then send the image to the OCF Docker server. This has to be specified in
+the `Makefile` at the root of the repository. Here's part of what templates has:
+
+```
+BIN := venv/bin
+
+DOCKER_REVISION ?= testing-$(USER)
+DOCKER_TAG = docker-push.ocf.berkeley.edu/templates:$(DOCKER_REVISION)
+RANDOM_PORT := $(shell expr $$(( 8000 + (`id -u` % 1000) + 1 )))
+
+
+.PHONY: dev
+dev: cook-image
+    @echo "Will be accessible at http://$(shell hostname -f ):$(RANDOM_PORT)/"
+    docker run --rm -p "$(RANDOM_PORT):8000" "$(DOCKER_TAG)"
+
+.PHONY: cook-image
+cook-image:
+    docker build --pull -t $(DOCKER_TAG) .
+
+.PHONY: push-image
+push-image:
+    docker push $(DOCKER_TAG)
+```
+
+The important targets to pay attention here are `cook-image` and `push-image`,
+Jenkins will try to run these as part of the deployment pipeline that continuously
+deploys to Kubernetes.
+
+Finally, we have to include a `Jenkinsfile` to the repository, so that Jenkins
+knows how we want it to go about deploying the service. In this case, it's just
+specifying that we want it to go through the pipeline.
+
+```
+servicePipeline(
+    upstreamProjects: ['ocf/dockers/master'],
+)
+```
+
+## Kubernetes
+
+In the root of your project create a `kubernetes` folder. This is where all
+your Kubernetes configuration files will live. Templates, a relatively simple
+service, is a single `nginx` server serving static content. Because this
+application is self-contained we need to create one file,
+`kubernetes/templates.yaml`.
+
+## Service
+Since templates is a web service we will first create a `Service` object. The
+first step to make your Kubernetes service internet-facing is to make your
+application accessible within the Kubernetes cluster. In most cases you can
+simply fill in this template.
+
+```
+apiVersion: v1
+kind: Service
+metadata:
+  name: <myapp>-service
+spec:
+  selector:
+    app: <myapp>
+  ports:
+    - port: 80
+      targetPort: <docker-port>
+```
+
+The `name` field under `metadata` resource is the name Kubernetes uses to
+identify your `Service` object when, for example, you run `kubectl get
+services`. The `selector` resource is the name you will use to bind `Pods` to
+this `Service` object. Fill in the `targetPort` with the port that your
+application uses _inside_ of the Docker container. In the case of templates we
+bind to port `8000`. Here is the `Service` configuration for templates with all
+the fields filled in.
+
+```
+apiVersion: v1
+kind: Service
+metadata:
+  name: templates-service
+spec:
+  selector:
+    app: templates
+  ports:
+    - port: 80
+      targetPort: 8000
+```
+
+## Creating the deployment
+
+Great! Now let's move onto creating our pods! To do this we'll create a
+`Deployment` object. Deployments can get become complicated with application
+specific configuration, but the simplicity of Templates elucidates the
+bare-bones requirements for any `Deployment`.
+
+```
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: <myapp>-deployment
+  labels:
+    app: <myapp>
+spec:
+  replicas: <#pods>
+  selector:
+    matchLabels:
+      app: <myapp>
+  template:
+    metadata:
+      labels:
+        app: <myapp>
+    spec:
+      containers:
+        - name: <container-name>
+          image: "docker.ocf.berkeley.edu/<your-repo-name>:<%= version %>"
+          resources:
+            limits:
+              memory: <#Mi>
+              cpu: <cpus-in-millicores>m
+          ports:
+            - containerPort: <docker-port>
+```
+
+This section can be a bit daunting, but we'll go through it step-by-step. Fill
+in `<app-name>` and `<docker-port>` with the same name you used in your
+`Service`. This will ensure your Pods are bound to the `Service` we previously
+created. `replicas` is the number of instances we want. Because Templates is
+used internally by OCF staff, we aren't super concerned with uptime and create
+only 1 instance. For a service like `ocfweb`, where uptime is crucial, we would
+opt for 3 instances to handle failover.
+
+The `containers` resource is where Kubernetes looks to obtain `docker` images
+to deploy. For production services this will _always_ be the OCF docker server:
+`docker.ocf.berkeley.edu`. The field `<your-repo-name>` is the name of the
+repository on the OCF GitHub, and version will be filled in automatically by
+[[Jenkins|doc staff/backend/jenkins]]. For testing, it is recommended you push
+your image to [DockerHub][dockerhub] or to `docker.ocf.berkeley.edu` (talk to
+a root staffer in the latter case) and use a hardcoded image name.
+
+Lastly, we set our resource limits. Templates is a low-resource service so
+we'll give it 1 megabyte of memory and `50/1000` of a CPU core (Kubernetes uses
+millicores for CPU units, so 1 core = 1000m). Do note that every instance of
+the application gets these resources, so with _N_ instances you are using _N *
+limits_.
+
+**WARNING**: On low-resource development cluster, asking for too much CPU or RAM
+can put your application in an infinite `Pending` loop since the cluster will
+never have enough resources to schedule your service (yes, this has happened to
+us).
+
+With all the fields filled in we have this Deployment object for Templates.
+
+```
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: templates-deployment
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,373 @@
...
+
+Finally, we have to include a `Jenkinsfile` to the repository, so that Jenkins
+knows how we want it to go about deploying the service. In this case, it's just
+specifying that we want it to go through the pipeline.
+
+```
+servicePipeline(
+    upstreamProjects: ['ocf/dockers/master'],
+)
+```
```
BernardZhao, line 70: This part is a little confusing to me, as in I described my understanding of it, but in other projects an empty `upstreamProjects` array also seemed to work fine. Perhaps @jvperrin could chime in on this?

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,373 @@
...
+run `kinit`. This will obtain a [[kerberos|doc staff/backend/kerberos]] ticket
+giving us access to the Kubernetes cluster. Now run
+
+```
+kubectl create namespace <myapp>
+kubectl apply -n <myapp> -f <myapp>.yaml
+```
+
+You can run `kubectl -n <myapp> get all` to Kubernetes create your `Service`
+and `Deployment` objects.
```
BernardZhao, line 332: This isn't relevant with Jenkins right? I thought this might be nice to include as example commands if someone was planning to deploy to dev-kubernetes.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,373 @@
...
+
+[kafka][kafka-deploy] (Advanced): Runs a `kafka` cluster inside of Kubernetes.
+
+[templates]: https://templates.ocf.berkeley.edu
+[dockerhub]: https://hub.docker.com
+[puppet]: https://github.com/ocf/puppet/tree/master/modules/ocf_kubernetes/files/persistent-volume-nfs
+[templates-deploy]: https://github.com/ocf/templates/tree/master/kubernetes
+[kanboard-deploy]: https://github.com/ocf/kanboard/tree/master/kubernetes
+[mastodon-deploy]: https://github.com/ocf/mastodon/tree/master/kubernetes
+[kafka-deploy]: https://github.com/ocf/kafka/tree/master/kubernetes
```
BernardZhao, line 373: We have a lot more services now - this many not be that accurate. What are the best examples at the OCF today?

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+specifying that we want it to go through the pipeline.
+
+```
+servicePipeline(
+    upstreamProjects: ['ocf/dockers/master'],
+)
+```
+
+**NOTE**: While Jenkins is supposed to scan all of the repositories in the organization,
+it won't necessarily be triggered if you transfter a repository over. Trigger the job
```
Baisang, line 73: transfer

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+repo for deployment.
+
+## Prerequisites
+If you're looking to deploy your service, you should first have a service to
+deploy. This is normally done with a repository underneath the
+[OCF's Github organization](https://github.com/ocf). If you have the correct
+permissions, you should be able to transfer, fork, or just create a
+repository in the org.
+
+### Docker & Jenkins
```
Baisang, line 26: It would be better for people who want to deploy services to just clone or copy these files somehow. They're just going to copy paste from this doc anyway. I would wager that the vast majority of people who want to deploy services on kubernetes don't need to know how this works, it just has to "work." (of course, we should provide docs somewhere that explain how it works anyway)

You could have them run some script that takes in their proposed service name and it could even generate a basic template for those kubernetes yaml files too.

Probably out of scope for this PR, but something to consider for the future.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+This section can be a bit daunting, but we'll go through it step-by-step. Fill
+in `<app-name>` and `<docker-port>` with the same name you used in your
+`Service`. This will ensure your Pods are bound to the `Service` we previously
+created. `replicas` is the number of instances we want. Because Templates is
+used internally by OCF staff, we aren't super concerned with uptime and create
+only 1 instance. For a service like `ocfweb`, where uptime is crucial, we would
+opt for 3 instances to handle failover.
+
+The `containers` resource is where Kubernetes looks to obtain `docker` images
+to deploy. For production services this will _always_ be the OCF docker server:
```
Baisang, line 168: Docker repository

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+as a much simpler option before making use of `NFS`.
+
+For Kubernetes to access the file system we need two objects: a
+`PersistentVolume` and a `PersistentVolumeClaim`. The former maps a filesystem
+to the cluster, and the latter is how a service asks to access that filesystem.
+You will need to create the `PersistentVolume` in [Puppet][puppet] as
+<app-nfs-pv.yaml>. In this example we'll create 30 gigabytes of readable and
+writeable storage.
+
+```
```
Baisang, line 269: Please have people try to use the https://github.com/ocf/nfs-provisioner instead. We set it up so you no longer need to make puppet changes to create PVs.

Baisang: It's a lot easier. All you need to do is make a PVC like in the README with the correct storageClassName.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+
+Now we have all the necessary configuration to deploy our service. At this point
+everything should already be deployed automatically by Jenkins. If deploying on
+`dev-kubernetes` we can run these commands to deploy the service manually.
+On `supernova`, first run `kinit`. This will obtain a
+[[kerberos|doc staff/backend/kerberos]] ticket giving us access to the Kubernetes cluster.
+Now run
+
+```
+kubectl create namespace <myapp>
```
Baisang, line 333: Technically you can include the yaml to create this within the <myapp>.yaml file

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+Now we've set up the volume claim. Finally, we need to tell Kubernetes to mount
+this `PVC` into our docker container. Under the `container` resource add:
+
+```
+          volumeMounts:
+            - mountPath: /target/path/in/my/container
+              name: <myapp-data>
+```
+
+
```
Baisang, line 322: Eventually you'll want to add a section for adding keycloak auth and handling secrets, but that's out of scope.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+          image: "docker.ocf.berkeley.edu/templates:<%= version %>"
+          resources:
+            limits:
+              memory: 128Mi
+              cpu: 50m
+          ports:
+            - containerPort: 8000
+```
+
+The last object we need to create for the Templates service is `Ingress`. We
```
Baisang, line 216: Does adding the name to https://github.com/ocf/puppet/blob/master/modules/ocf_kubernetes/manifests/master/loadbalancer.pp still apply? If so we need to include that step.

ocfweb/docs/docs/staff/procedures/deploying-services.md
jvperrin: Yeah, that's a great question! The only thing this does is mean that a service build pipeline gets kicked off if the upstream one listed here succeeds. The reason why you see `ocf/dockers/master` in a number of services is when they are both (1) built off the OCF base docker images and (2) redeploying them can be done smoothly. These redeploys are mainly done so that security updates are pulled into service containers in a timely manner, so it doesn't have to be a daily thing but could probably be weekly or something instead.

For instance, [ocfweb has both ocflib and dockers](https://github.com/ocf/ocfweb/blob/303e0671ace54ce6e70114d7f36908c329d41a31/Jenkinsfile#L2) as upstreams because it depends on the latest versions of both of those repos, and because redeploying after either one finishes building successfully can be done smoothly. [slackbridge](https://github.com/ocf/slackbridge/blob/9dd3ac36b9e62ebb9b886435f6a2fe6458406431/Jenkinsfile#L2) on the other hand doesn't have any upstream projects listed despite being built off the OCF base image because every time it rebuilt it causes large join/quit spam on IRC that wasn't a pleasant experience.

Essentially, to figure out what's listed here for a new service, you only need to know what base image the service is using, and how disruptive a deploy would be.

ocfweb/docs/docs/staff/procedures/deploying-services.md
jvperrin: I don't believe it's necessary because Jenkins will create the app namespace and all that when deploying, and I think it also does the service/deployment creation parts too but I'm not too familiar with that side of things.

Maybe just mention before this that it's something you can do with your dev namespace to test out running the service somewhere?

ocfweb/docs/docs/staff/procedures/deploying-services.md
jvperrin: `templates` is still a really good simple example, and I think `mastodon` is a good complex example, but I'm not sure about the others. I do know kanboard isn't running any more (gives a 503), and I don't really know the state of `kafka`.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+[[Jenkins|doc staff/backend/jenkins]] will try to build that container,
+and then send the image to the OCF Docker server. This has to be specified in
+the `Makefile` at the root of the repository. Here's part of what templates has:
+
+```
+BIN := venv/bin
+
+DOCKER_REVISION ?= testing-$(USER)
+DOCKER_TAG = docker-push.ocf.berkeley.edu/templates:$(DOCKER_REVISION)
+RANDOM_PORT := $(shell expr $$(( 8000 + (`id -u` % 1000) + 1 )))
```
jvperrin, line 41: For what it's worth, this `+ 1` is to space out ports so you can run multiple services under one user without them colliding. I think we should probably come up with a better port allocation strategy, or just use the same port for the same user each time and forego this method as it's pretty confusing if you don't understand it.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+    docker build --pull -t $(DOCKER_TAG) .
+
+.PHONY: push-image
+push-image:
+    docker push $(DOCKER_TAG)
+```
+
+The important targets to pay attention here are `cook-image` and `push-image`,
+Jenkins will try to run these as part of the deployment pipeline that continuously
+deploys to Kubernetes.
```
jvperrin, line 60: `make test` is also a good one to mention, it's automatically run if it exists as part of the service pipeline so that's a great place to put any tests that a service may have.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+Note that `serviceName` _must_ be the same as that used in the `Service`
+object. Now that we have ingress, all requests with the `Host` header
+`templates.ocf.berkeley.edu` will be directed to a Templates Pod!
+
+
+## Deployment extras
+
+### OCF DNS
+
+If your application at any point uses OCF-specific DNS, like using the hostname
```
jvperrin, line 245: I'd probably mention the term "search domain" somewhere in here as that's the keyword that can help to find more information on the internet about this behavior.

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+
+[kanboard][kanboard-deploy]: Project management software that makes use of `ldap` and mounts `nfs`.
+
+[mastodon][mastodon-deploy] (Advanced): Applies custom patches, uses `ldap`, mounts `nfs`, has pods for `redis`, `sidekiq`, and `http-streaming`.
+
+[kafka][kafka-deploy] (Advanced): Runs a `kafka` cluster inside of Kubernetes.
+
+[templates]: https://templates.ocf.berkeley.edu
+[dockerhub]: https://hub.docker.com
+[puppet]: https://github.com/ocf/puppet/tree/master/modules/ocf_kubernetes/files/persistent-volume-nfs
```
jvperrin, line 375: nit: For these 3 links that are referred to earlier on in the doc, I'd put them closer to their respective paragraphs just so that it's easier to see what is actually being linked to closer to the text that's doing the linking. (That or using the `[]()` link syntax instead)

ocfweb/docs/docs/staff/procedures/deploying-services.md
```diff
@@ -0,0 +1,379 @@
...
+follow the instructions below.
+
+The final step to make your service live is to create a DNS entry for your
+Kubernetes service.  You will need to clone the OCF dns repo.
+
+```
+git clone git@github.com:ocf/dns.git
+```
+
+Since we are adding DNS for a Kubernetes service, we run `ldapvi
```
jvperrin, line 354: This'll also need a `kinit $USER/admin` before the `ldapvi` (either as a separate command or just put as a prefix to the command) to actually be able to edit and submit changes to ldap.

I think these editing DNS instructions should actually almost be their own docs. I looked and there's https://www.ocf.berkeley.edu/docs/staff/procedures/new-host/#h3_step-11-add-the-ldap-entry in the docs for adding a new host and DNS, but this is a bit different as it's just editing the existing `lb-kubernetes` pseudo-host instead. Anyway, something to think about, but this might be useful as a separate docs page that multiple places can link to.

jvperrin: Actually, `lb-kubernetes` no longer exists since there's no non-kubernetes version, it's just a `cn=lb` now (this is something that changed since the previous pull request).

ocfweb/docs/docs/staff/procedures/deploying-services.md
encadyma: Github has template repositories that can help solve the issue of copying files, so we can create one for Kubernetes deployments.

nikhiljha:
The funny thing is we might actually obsolete these docs before they're merged.

BernardZhao:
> 
> 
> The funny thing is we might actually obsolete these docs before they're merged.

Thats why I paused on updating this

nikhiljha:
https://notes.ocf.berkeley.edu/jCl7t_M7RYGzcqQpbzWrOQ#

PR #642, opened 2020-09-16
Remove status bar

bjb568:
lab hours are not that useful when the lab is closed all semester. revert this when MLK opens.


Vio-Eli (approved):
lgtm for now, we still need to fix the link tho

PR #649, opened 2020-10-02
Switch over to CSS sticky footer

PotatoParser:


ja5087 (approved):
I assume you tested this (hopefully on mobile too) and it doesn't behave weirdly. Thanks!

PR #650, opened 2020-10-12
Update shell.md

### LINKED ISSUE #633: Windows SSH instructions
Windows has shipped with OpenSSH enabled by default since [May](https://virtualizationreview.com/articles/2020/05/18/ssh-on-windows.aspx). Does it still make sense to have [separate instructions](https://www.ocf.berkeley.edu/docs/services/shell/#h4_windows) for Windows users to install PuTTY?

PotatoParser:
Fixes #633

### DIFF
--- a/ocfweb/docs/docs/services/shell.md
+++ b/ocfweb/docs/docs/services/shell.md
@@ -35,7 +35,7 @@ Your shell account can be controlled remotely using
 If you just need to access SSH quickly, you can use our [web-based SSH
 interface](https://ssh.ocf.berkeley.edu/) from your web browser.
 
-#### Mac OS X or Linux
+#### Mac OS X, Linux, or Windows 10 (Version 1803 and above)
 
 On Mac OS X or Linux, enter in the terminal:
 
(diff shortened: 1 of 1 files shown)


ocfweb/docs/docs/services/shell.md
```diff
@@ -35,7 +35,7 @@ Your shell account can be controlled remotely using
 If you just need to access SSH quickly, you can use our [web-based SSH
 interface](https://ssh.ocf.berkeley.edu/) from your web browser.
 
-#### Mac OS X or Linux
+#### Mac OS X, Linux, or Windows 10 (Version 1803 and above)
 
 On Mac OS X or Linux, enter in the terminal:
```
jvperrin, line 40: You'll also need to update this line

PotatoParser: Whoops! Thanks for the catch!

jvperrin (approved):
Probably a bit repetitive with specifying the OS versions twice, but definitely better than what was here before. Thanks for the change!

PR #662, opened 2021-01-21
docs/contact/irc: three options => four options

chriskuehl:
Unless this was some kind of snub against XMPP being simple, we probably
shouldn't say there are "three simple options" and then list four options.


dkess:
It's more likely that we (I?) added the option and forgot to update that counter. Might be better to not include the number at all?

ja5087:
Sorry to hijack your PR but hopefully this also fixes it more permanently

nikhiljha (approved):

ethanhs (approved):

PR #681, opened 2021-04-07
Fix STF banner wrapping on mobile

singingtelegram:


ocfweb/templates/base.html
```diff
@@ -101,7 +101,8 @@
             <div class="container">
                 <p>
                     <strong>Vote for the Student Tech Fund's renewal during the ASUC Elections @ April 5-7!</strong>
-                    <span class="nowrap">
+                    <span>
+                        <br>
```
snarain01, line 105: Looks good, seems like a pretty quick fix that'll make the banner look nice on mobile. Thanks for doing this!

snarain01 (approved):
Going to approve the merge since its a pretty simple fix

PR #683, opened 2021-04-11
switching banner to announce STF passing

snarain01:
Since the STF renewal has passed, we should change the banner


ocfweb/templates/base.html
```diff
@@ -100,11 +100,11 @@
...
             <div class="container">
                 <p>
-                    <strong>Vote for the Student Tech Fund's renewal during the ASUC Elections @ April 5-7!</strong>
+                    <strong>The Student Tech Fund Renewal has Passed!</strong>
                     <span>
                         <br>
-                        The OCF is funded by the Student Technology Fund, which will expire if not renewed.
-                        <a href="{% url 'doc' 'communications/stf-renewal' %}">read more &raquo;</a>
+                        Thanks to all who voted. Learn more about the Student Technology Fund
+                        <a href="{% url 'doc' 'communications/stf-renewal' %}">here</a>.
```
ja5087, line 107: the `&raquo;` is the arrow symbol

ja5087 (approved):
I would suggest changing the title also (Line 1), but otherwise looks good!

PR #690, opened 2021-04-27
Update requirements

ethanhs:
This caused Django to break because apparently our `CACHES` setting was using an older format, so I fixed that. Also minor changes to make mypy happy.


ja5087 (approved):
#shipit

cg505 (approved):
assuming tested

encadyma (approved):

PR #709, opened 2021-08-07
Clarify Email request procedure

ethanhs:
This is for if you don't have a vhost


ja5087:
What if they only want a mail vhost and not a web one? Or is the new process to use that form always.

ethanhs:
The new domain template requests A, AAAA, and MX records when we email hostmaster, so I figured we might as well ask people to request a vhost.

PR #712, opened 2021-08-24
Printing

ethanhs:
This is the "remote" part of the remote printing workflow. Tomorrow I plan to get the other half finished so we can test things out before the lab opens.

The design of the workflow is that the user uploads a file that is sftp'd to their NFS mount on tsunami. A QR code is generated with information needed to trigger the print in the lab, which the user can bring with them.

Changes to the front page (adding "Print Remotely" to the front page and for when you are signed in)

image

The submission UI: 
image

The QR page:
image


ocfweb/account/print.py
```diff
@@ -0,0 +1,147 @@
...
+            'account/print/index.html', {
+                'title': 'Print remotely',
+                'form': form,
+            },
+        )
+
+
+class PrintForm(Form):
+    username = forms.CharField(
+        label='OCF username',
```
nikhiljha, line 119: Can we grab this from the currently logged in user?

ocfweb/account/templates/account/print/qr.html
```diff
@@ -0,0 +1,21 @@
+{% extends "base.html" %}
+{% load bootstrap %}
+
+{% block content %}
+    <div class="row">
+        <div class="col-md-8 ocf-content-block">
+            <p>
+                Please bring this QR code to the lab to start your print job. You may want to take a photo of it with your phone.
```
nikhiljha, line 8: Maybe suggest a screenshot (or a very good, even picture)? A typical phone camera picture of a screen containing a QR code might be difficult to scan.

nikhiljha (approved):
lgtm

ocfweb/account/templates/account/print/qr.html
saurabhnarain: We could also automatically emails this QR code to the associated email with the account

nikhiljha: Ooh, that's smart too!

ocfweb/account/print.py
```diff
@@ -0,0 +1,147 @@
...
+            password = form.cleaned_data['password']
+
+            file = request.FILES['file']
+
+            double_or_single = form.cleaned_data['double_or_single']
+
+            ssh = SSHClient()
+
+            host_keys = ssh.get_host_keys()
+            entry = HostKeyEntry.from_line(
```
ja5087, line 44: Would prefer we refactor this to share logic with https://github.com/ocf/ocfweb/blob/8b35c17d91a7d722882bf2672c35099e22e573e6/ocfweb/account/commands.py but not a PR blocker

ja5087 (approved):
looks fairly good to me

ocfweb/account/print.py
ethanhs: Yeah I will do this in a followup.

ocfweb/account/print.py
ethanhs: No, I don't believe so. We have their calnet UID and that isn't one to one with an OCF account

ocfweb/account/templates/account/print/qr.html
ethanhs: That's a great idea! It seems I need to update ocflib to do that, so I'll do that tonight if I get to it.

ethanhs: https://github.com/ocf/ocflib/pull/230

ethanhs: Okay I've made this change. It now emails the QR code to you!

ethanhs:
Jenkins will be marked as failed because coveralls is down for maintenance...

ethanhs (commented):
fydai approves

PR #722, opened 2021-09-18
Remove un-used Metabase iframe for session-stats

harrisonkuo:
I didn't realize that the session stats was just hidden underneath a huge iframe.

### DIFF
--- a/ocfweb/stats/templates/stats/session_stats.html
+++ b/ocfweb/stats/templates/stats/session_stats.html
@@ -9,7 +9,6 @@
 
     <div class="ocf-content-block">
     {% stats_navbar %}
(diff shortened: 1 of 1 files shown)


harrisonkuo:
Coveralls is down.

ethanhs:
I think the plan is to re-enable metabase, so I'm not sure if this is the right idea.

harrisonkuo:
> I think the plan is to re-enable metabase, so I'm not sure if this is the right idea.

We can re-add the line when Metabase is up, but right now session-stats is not usable in a sense that there is a huge empty iframe that requires you to scroll several screens down to find the current stats.

ethanhs (approved):

ethanhs:
> 
> 
> > I think the plan is to re-enable metabase, so I'm not sure if this is the right idea.
> 
> We can re-add the line when Metabase is up, but right now session-stats is not usable in a sense that there is a huge empty iframe that requires you to scroll several screens down to find the current stats.

Yeah that's fair.

PR #723, opened 2021-09-21
Remove coveralls

ethanhs:
Coveralls has been down for almost a week(!) and it is causing CI to fail. I think we should just remove it.


dongkyunc (approved):
shipit

PR #728, opened 2021-10-02
Added shorturl page at ocf.berkeley.edu/test/shorturl

bryli:
HTML page is thanks to @strellic (https://github.com/ocf/projects/issues/64#issuecomment-787578155). New page will be linked to ocf.io/shorturl.

### DIFF
--- a/ocfweb/test/shorturl.py
+++ b/ocfweb/test/shorturl.py
@@ -1,0 +1,13 @@
+from django.http import HttpRequest
+from django.http import HttpResponse
+from django.shortcuts import render
+
+
+def test_shorturl(request: HttpRequest) -> HttpResponse:
+    return render(
+        request,
+        'test/shorturl.html',
+        {
+            'title': 'OCF ShortURLs',
+        },
+    )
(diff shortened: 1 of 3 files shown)


ethanhs:
Ah hm, it looks like there is already a version live on the new site: https://new.ocf.berkeley.edu/shorturls/

bryli:
IIRC I was told we just wanted something on the old site until we fully move over to the new site. Could also make the puppet short URL config link to the new ocf website for short URLs though

ethanhs:
That seems reasonable, could you send a screenshot of what it looks like if you get the chance?

bryli:
Now that I'm looking at it, it does look a bit too barebones. (https://imgur.com/jtSQMUR) Maybe I can add in the menu bar, put it in the staff docs, and make it look nicer, then resubmit the PR.

ethanhs:
I might recommend having the template inherit from the ocfweb base html template, and yeah I think putting it in staff docs seems reasonable. Thanks for working on this!

bryli:
The page has been updated to look better, with the same style as the rest of the site. http://i.imgur.com/Py0FfMB.png

PR #732, opened 2021-10-14
Added print quota

origamiman72:
Added print quota API so that it can be used to display print information on the [ocfstatic dashboard](https://github.com/ocf/ocfstatic/issues/251#issuecomment-939566143).

- @origamiman72 and @Kalissaac


Kalissaac (approved):
few years later but we got it merged

PR #740, opened 2021-12-04
Add meeting hours endpoints

ben9583:
Adds endpoints to the newly-added meeting hours methods from ocflib. These methods will allows public access to an API that returns a list of weekly meetings, the next meeting, and the current meeting (if one is happening at the time). The corresponding ocflib pr can be reviewed [here.](https://github.com/ocf/ocflib/pull/238)


bentref (approved):
looks good to me, seems in line with how the other endpoints are done. One more level down!

ben9583:
Thank you! Yeah next step is to get rid of all the hard-coded references to meetings and maybe work on that calendar on ocfstatic.

PR #743, opened 2021-12-24
Update meeting_hours.py

singingtelegram:
builds are failing

### DIFF
--- a/ocfweb/api/meeting_hours.py
+++ b/ocfweb/api/meeting_hours.py
@@ -31,7 +31,7 @@ def get_current_meeting(request: HttpRequest) -> JsonResponse:
     if current_meeting is None:
         return JsonResponse(
             {},
-            status=204,
+            status=200,
         )
 
     return JsonResponse(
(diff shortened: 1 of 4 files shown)


ocfweb/api/meeting_hours.py
```diff
@@ -31,7 +31,7 @@ def get_current_meeting(request: HttpRequest) -> JsonResponse:
     if current_meeting is None:
         return JsonResponse(
             {},
-            status=204,
+            status=200,
```
ethanhs, line 34: This doesn't seem quite right, we are sending an empty response.

ethanhs:
Also, doesn't need to happen in this PR but we should try building images with bullseye and stop using the custom Python :)

ocfweb/api/meeting_hours.py
singingtelegram: true but the test suite wants 200 for each endpoint, so `make test` has been failing since there's no current meeting and the endpoint returns 204

ethanhs: That may be, but we should accept 204 in the test suite, not change the code to be incorrect.

singingtelegram: fixed

ethanhs (approved):
Great! Thank you for fixing this.

PR #745, opened 2021-12-25
Update constitution

singingtelegram:
Update the OCF constitution as per BoD on 11/22.

Re-do of #741 due to build issues.


singingtelegram:
Merging since #741 was approved.

PR #746, opened 2021-12-25
Update constitution

singingtelegram:
Update the OCF constitution as per BoD on 11/22.

Issue was [puppet#1212](https://github.com/ocf/puppet/issues/1212)


singingtelegram (commented):
Approved (same commit as #741)

PR #754, opened 2022-02-11
Add docs for CLI printer removal

Boomaa23:
Fix docs according to https://github.com/ocf/utils/issues/101.

### DIFF
--- a/ocfweb/docs/docs/staff/procedures/printing.md
+++ b/ocfweb/docs/docs/staff/procedures/printing.md
@@ -13,8 +13,8 @@ Alternatively, this can be done using the CUPS command `lpadmin` with proper aut
 
     lpadmin -p printername -r classname
 
-The `printername` is the physical name of the printer (for example, 
-`logjam-double` or `logjam-single`) while the `classname` is the name 
+The `printername` is the physical name of the printer (for example,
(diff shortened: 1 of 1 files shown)


dkess:
The reason why my issue was filed in ocf/utils is because I was suggesting writing a script to do it. When taking a printer out of service, you always want to take it out of service for double and single at the same time. These docs don't make that clear. It's better to have a script where you just select a printer and it automatically runs the command twice.

Boomaa23:
I see. Okay in that case I will work on it some more, add a script, and modify these docs.

dkess:
Please explain that the script has to be run on whiteout.

Also, I get a command not found error when I try it on whiteout now.

Boomaa23:
I will add that it has to be run on whiteout. I'm fairly confident "command not found" is because someone with root has to run `update-symlinks`. An additional problem is that `required` is an invalid argument for a subparser with Python 3.5, so that will have to be fixed first.

Boomaa23:
See https://github.com/ocf/utils/pull/175 for script fix. I don't have root so I'll try to ask someone to run `update-symlinks` for me.

Boomaa23:
`update-symlinks` done, see https://github.com/ocf/utils/pull/176. Should be good to go!

dkess (approved):

PR #755, opened 2022-02-17
Fix meeting-hours caching

ben9583:
The meeting hours API has some endpoints that are time sensitive (`/api/meetings/current`, `/api/meetings/next`) that were previously cached for a long time, meaning the API would return out-of-date results for both of these. This pr fixes that by using a 5-second periodic call, which is also the method the computer availability API uses.

The endpoint `/api/meetings/list` is not affected by this change.


Kalissaac (approved):
thanks!

Kalissaac:
oh wait the tests failed

PR #769, opened 2022-07-27
Fix hostmaster link on vhost form

Boomaa23:
I did a small oops and initialized the `href` in the wrong place, so it looks like the below. This PR moves the initialization of the label to the correct location. See #767.


Kalissaac (approved):
lgtm thanks

PR #844, opened 2024-09-26
Add NixOS Desktop Upgrade Tracker and Staff URL

kinn-edendev:
Added 2 files and modified 1:

desktop_upgrade.html and desktop_upgrade.py are effectively copies of previous Debian Buster upgrade pages, renewed to feature desktop names instead of server names and the current status at the time of this PR.

urls.py was modified to include these new pages so that they can be accessed from /docs/staff.

The fork passed all the tests outlined in the README.md.


kinn-edendev:
That's great, it passed all the tests on Supernova and now its failing checks in files that weren't even modified.

PR #891, opened 2026-05-02
Devenv

jaysa68:
should work on personal laptop, ocf desktop, and koi (staff login server)


24apricots (approved):
maybe we could have the build output in the nix store so that multiple users working on it can use the same cached build output instead of only having it in the venv

PR #899, opened 2026-05-13
print form website

danxliu:
Print from website under account
Use 6 digit otp from joes vm to make sure people are in the ocf


ocfweb/stats/printing.py
```diff
@@ -12,15 +12,15 @@
...
 from ocflib.printing.quota import get_connection
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
 from ocfweb.caching import periodic
 from ocfweb.component.graph import plot_to_image_bytes
+from ocfweb.printing import get_printers
 
-ALL_PRINTERS = ('papercut', 'pagefault', 'logjam', 'logjam-old', 'deforestation')
-ACTIVE_PRINTERS = ('papercut', 'pagefault', 'logjam')
+ALL_PRINTERS = ('papercut', 'pagefault', 'logjam', 'logjam-old', 'deforestation', 'OCF-BW', 'OCF-Color')
```
BNH440, line 22: please don't use the aggregate names here, these variables are referenced elsewhere in ocfweb

ocfweb/stats/printing.py
```diff
@@ -29,7 +29,7 @@ def stats_printing(request: HttpRequest) -> HttpResponse:
         'stats/printing.html',
         {
             'title': 'Printing Statistics',
-            'current_printers': PRINTERS,
+            'current_printers': get_printers().keys(),
```
BNH440, line 32: why did you update the variables above and then not use them

chamburr (approved):

devenv.lock
BNH440, line 1: why did the devenv get updated as a side effect to this pr

conf/ocfweb.conf
```diff
@@ -20,3 +20,7 @@ db = ocfmail
 user = ocfstats
 password = password
 db = ocfstats
+
+[printing]
+# Joe's VM
```
BNH440, line 25: using joe's staff vm is not acceptable for deployment, also there's probably better ways to do this

ocfweb/stats/printing.py
BNH440: see my stats branch for the proper way to do this, although it's far from finished https://github.com/ocf/ocfweb/tree/printer-stats-update

BNH440 (changes requested):
overall I think this is unnecessary, it looks to me like the print queue gets pretty backed up during peak times, I think this would just make things worse

sawansri (approved):

PR #901, opened 2026-05-24
Redirect button to random vhost websites

Storce:
Uses ocflib.get_vhosts() and chose a host randomly, and redirect user to it.


vqbc (commented):
pls replace ", visit" with " — visit"

PR #906, opened 2026-09-05
Fix virtual host request domain length limit

### LINKED ISSUE #905: requested vhost domain limited to 32 chars
the vhost domain request field, where the user inputs their subdomain.studentorg.berkeley.edu, is capped at 32 characters which seems to be arbitrary. This should either be expanded to 64 characters to match the other fields with arbitrary length limits on the form, or expanded to 87 characters, to account for a max dns record length of 63 chars plus the ".studentorg.berkeley.edu" portion

this length limit has become an actual issue for student groups with longer names, see rt#16029

https://github.com/ocf/ocfweb/blob/master/ocfweb/account/vhost.py#L221

Gambit-Checkmate:
The virtual-host request form rejects domains longer than 32 characters, preventing student groups with longer names from requesting hosting.

Set the field limit to a 63-character label plus `.studentorg.berkeley.edu` (87 characters total). Django applies the limit to both server-side field validation and the HTML input's `maxlength`.

Closes #905.

Validation: evaluated the actual field declaration with Django 5.2.5 in isolation: 33-, 64-, and 87-character domains pass, 88 characters fail, and the widget exposes `maxlength="87"`. The original 32-character limit rejects the same 33-character example. Flake8 and `git diff --check` pass. The full OCF website test suite was not run locally; this validation isolates the field from OCF service dependencies.

### DIFF
--- a/ocfweb/account/vhost.py
+++ b/ocfweb/account/vhost.py
@@ -218,7 +218,7 @@ class VirtualHostForm(Form):
     requested_subdomain = forms.CharField(
         label='Requested domain:',
         min_length=1,
-        max_length=32,
+        max_length=63 + len('.studentorg.berkeley.edu'),
         widget=forms.TextInput(attrs={'placeholder': 'mysite.studentorg.berkeley.edu'}),
     )
 


BNH440:
please disclose AI usage in your pr, and never have pr body text be AI written. all text meant to be read by a human (comments, pr body, etc) should be written by a human as well

PR #908, opened 2026-09-12
Clear out obsolete pages, minor style fixes

vqbc:


vqbc:
the commit history is weird im sorry just squash it

ocfweb/announcements/templates/announcements/2016-04-01-renaming.html
kpengboy, line 1: Not too important, but I wonder if there's a way to have hidden announcements. Some like this one are still fun organizational lore...

vqbc: They're all in the mkdocs archive now https://bestdocs.ocf.io/news/2016/03/31/unveiling-a-new-name-for-the-open-computing-facility/

BNH440:
I fixed the history, but yes we should squash anyways

kafu-densha (approved):

